PR #7, opened 2018-02-08
Update docs about default shortcuts

grimpy:


varjolintu:
Btw, this has been already committed to the develop branch of the non-official repo (which is currently more up-to-date). But this could be still approved.

PR #10, opened 2018-02-13
Update jQuery from 3.2.1 to 3.3.1

varjolintu:
Updates jQuery to the latest version 3.3.1. Also some deprecated functions are replaced.


keepassxc-browser/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
 	"manifest_version": 2,
 	"name": "KeePassXC-Browser",
-	"version": "0.5.0",
+	"version": "0.5.2",
```
phoerious: From here on, we shouldn't increase the version anymore for every minor change.

varjolintu: Yes we shouldn't. But it's strange because didn't that already changed in https://github.com/keepassxreboot/keepassxc-browser/pull/8 ?

phoerious: Yeah, weird.

PR #19, opened 2018-02-22
Preliminary test cases

varjolintu:
These are preliminary tests cases for the extension.

To use them, do `npm install` and then you can use the following commands for testing:
- `npm run unit-tests`
- `npm run integration-tests`
- `npm run all-tests`

### DIFF
--- a/test/integration.js
+++ b/test/integration.js
@@ -1,0 +1,151 @@
+var chrome = require('selenium-webdriver/chrome'),
+    firefox = require('selenium-webdriver/firefox'),
+    webdriver = require('selenium-webdriver'),
+    test = require('selenium-webdriver/testing'),
+    assert = require('selenium-webdriver/testing/assert'),
+    until = require('selenium-webdriver/lib/until'),
+    fileUrl = require('file-url'),
+    fs = require('fs')
+
+var wd;
+
+test.before(function(done) {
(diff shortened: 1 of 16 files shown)


test/integration.js
```diff
@@ -0,0 +1,204 @@
...
+		});
+	});
+});*/
+
+
+// Test external URL's and check that input fields are found and the password generator icon is visible
+// This site could also be internal and included in the test folder.
+test.describe('Testing external sites', function() {
+	test.it('Test input fields', function() {
+		wd.get("https://www.github.com/login");
```
weslly: It would be interesting to use the [test form at keepass.info](https://keepass.info/help/kb/testform.html) instead since it is less likely to change than the github login page.

varjolintu: Good idea! Also, of course it's possible to create a localhost based test pages or use the ones from w3school.com.

varjolintu: Still, we need to have test for pages that create the login form from a div that is hidden or created dynamically.

droidmonkey: We can host our own too. I personally prefer to bundle any test pages in with the test suite. Don't rely on "live" sites for your testing.

varjolintu:
This has been rebased and is ready to go. If anyone has any suggestions for more local test sites etc. now it's the time.

varjolintu:
Closed as outdated.

Issue #26, opened 2018-02-28
Request: Compatibility with Firefox ESR (52.6.0)


Ketrel:
## Expected Behavior
Addon should be marked compatible with Firefox ESR (Currently: 52.6.0)

## Current Behavior
Addon reports incompatible with ESR and requires 55+

## Possible Solution
Mark compatible with 52.6.0 ESR

## Steps to Reproduce (for bugs)
n/a

## Debug info
I used Nightly Tester Tools to force addon compatibility to allow the install.  The Addon appears functional in this state, so barring any major back end changes, it should already be compatible in everything but the name.

Ketrel:
Hmm, could you give examples of what I should see if those worked, and which settings would need to be on for it to function?

(Also worst case if it's not possible, could it still be marked compatible, with the caveat that some features may not work fully?)

varjolintu:
The first one is simple: just check if the right mouse click shows the KeePassXC context menu.
With the second one you must use [this](https://www.httpwatch.com/httpgallery/authentication/) page for testing the HTTP Auth (or some similar page) and make a temporary entry to your database to test it.

Ketrel:
Ok, I checked both, and they do appear to be issues.
Since this version of the ESR is feature frozen, would the second option (to allow the install, but note limitations on functionality) be feasible so one doesn't have to use nightly tester tools to install?

(A good deal of people are sticking with the ESR due to legacy extensions, such as Tab Mix Plus, and FlashGot which haven't/can't be updated for Webextensions yet)

varjolintu:
I think a notification about the limitations would be enough. Next ESR which is based is already going to be released at May, but that's not soon enough.

PR #38, opened 2018-03-03
Add support to Firefox ESR

varjolintu:
Solves https://github.com/keepassxreboot/keepassxc-browser/issues/26.

This has a few limitations because of an older WebExtension API:
- Showing context menus on password fields (menus.ContextType)
- HTTP Auth support (webRequest.onAuthRequired)

User's should be notified for these if the wish to use Firefox ESR.

### DIFF
--- a/keepassxc-browser/manifest.json
+++ b/keepassxc-browser/manifest.json
@@ -104,7 +104,7 @@
 	"applications": {
        "gecko": {
            "id": "keepassxc-browser@keepassxc.org",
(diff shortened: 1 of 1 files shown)


phoerious:
Can this be merged or will you continue working on it

varjolintu:
This only needs to be mentioned in the FAQ that these few features are not supported with Firefox ESR.

phoerious:
OK.

Issue #34, opened 2018-03-01
keepassxc-browser doesn't autofill custom placeholders


Generator:
**This issues is not related with "Fill TOTP" from context menu**

keepassxc-browser doesn't auto-fill custom placeholders when specified with "[Choose own credential fields](https://github.com/smorks/keepasshttp-connector/blob/master/documentation/KeePassHttp-Connector.md#43-settings-specified-credential-fields)"

## Expected Behavior

When a page have their specified credential fields, should [auto-fill all custom fields](https://github.com/smorks/keepasshttp-connector/blob/master/documentation/KeePassHttp-Connector.md#37-fill-in-additional-values-via-string-fields) (username/password/KPH placeholders)

## Current Behavior

Only username and password is filled.

## Possible Solution

Custom placeholders working without issues with [KeePassHttp-Connector](https://github.com/smorks/keepasshttp-connector)

## Steps to Reproduce (for bugs)

1. Create a new entry to https://python-totp.herokuapp.com/
2. Register and setup TOTP
3. On new entry go to "Advanced" and create a new attribute `KPH: 1 - TOTP`, with value `{TOTP}`
4. On login page from keepassxc-browser icon "Choose your own credential fields", select username field (skip password) and string field
5. Reload page (if doesn't fill username try to redetect credential fields)

## General Info
KeePassXC - Version 2.3.0 Revision: 4c0ed74
keepassxc-browser - Version: 1.0.0

Operating system: Archlinux
Browser: Firefox 58.0.2
Proxy used: NO

olifre:
Same here, Gentoo, Firefox 58.0.1, same versions of KeePassXC and keepassxc-browser.

TheZ3ro:
Maybe related to #28 ?

olifre:
To cite @Generator 

> This issues is not related with "Fill TOTP" from context menu

So I don't believe it's related. 

I also can't get any other custom fields to work. For example, for a page which takes an additional string field, I can select this using `Choose your own credential fields:`, and select that string field - but it will never be filled automatically. 
I have created an associated attribute called `KPH: 1 - Pin` with corresponding value. The very same attribute worked fine with https://addons.mozilla.org/de/firefox/addon/keepasshttp-connector/ .

Generator:
[...]

PR #41, opened 2018-03-04
Fixes the custom placeholders

varjolintu:
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/34.

StringFields were used in a old list format, not as standard JavaScript arrays.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -1504,15 +1504,16 @@ cip.setValue = function(field, value) {
     }
 };
 
-cip.fillInStringFields = function(fields, StringFields, filledInFields) {
+cip.fillInStringFields = function(fields, stringFields, filledInFields) {
     let $filledIn = false;
 
     filledInFields.list = [];
-    if (fields && StringFields && fields.length > 0 && StringFields.length > 0) {
+    if (fields && stringFields && fields.length > 0 && stringFields.length > 0) {
         for (let i = 0; i < fields.length; i++) {
             const $sf = _fs(fields[i]);
-            if ($sf && StringFields[i]) {
-                cip.setValue($sf, StringFields[i].Value);
+            const stringFieldValue = Object.values(stringFields[i]);
+            if ($sf && stringFieldValue[0]) {
+                cip.setValue($sf, stringFieldValue[0]);
                 filledInFields.list.push(fields[i]);
                 $filledIn = true;
             }


SmoochiesBoss:
@phoerious Do I need to update my database file somehow for the KPH references to work again? Currently they're not working for me with KeePassXC-Browser 1.4.0 and KeePassXC 2.4.1.

varjolintu:
@SmoochiesBoss In what way are they defined now? For TOTP you don't need a KPH attribute. For String Fields, it's enough to have `KPH: ` at the start of the attribute.

SmoochiesBoss:
They are listed under "Advanced" as "Additional attributes". Each starts
with "KPH: " followed by a number and then the actual data is in the box to
the right of that. Defining the fields on a web page using
KeePassXC-Browser works fine (ie, it allows me to define the additional
fields), but it doesn't actually fill in those fields. I just upgraded from
2.3.3 which worked fine.

E

varjolintu:
You don't need the number. `KPH: ` at the start is enough. The fill order is the same as the attribute order.

SmoochiesBoss:
Right. I didn’t change the file but just upgrading from 2.3.3 to 2.4.1 killed the feature for me.

E

varjolintu:
I can reproduce it, but I'm quite sure it's because of the large update of the extension we just made.

varjolintu:
And I found the reason. Thanks for finding the bug!

SmoochiesBoss:
My pleasure!

E

SmoochiesBoss:
Is it a bug or is there something I can do on my end to make it work again?

E

varjolintu:
Just wait for the update.

olifre:
@varjolintu Was #498 the fix you mentioned? 
This did not fix the issue for me, I still observe it with 1.4.3. Is there another, not closed issue / PR tracking status?

varjolintu:
@olifre Yes. At least for me it works correctly, also with auto-fill. It fills the username/password and String Fields.

olifre:
@varjolintu Ok, for me it fails - but it also fails already at the point when I want to select the custom fields. 
Initially, all three fields on the page are shown as "selectable", but as soon as I select the username field, the string field on the page is not selectable anymore. I can only select the password field. 
It happens to me with several login pages (sadly, none of them can be shared). 

An example is shown in these obfuscated screenshots:
Step-1
Step-2
Maybe an issue with the German localization?

varjolintu:
@olifre And using "Mehr anzeigen" doesn't help in step 2?

olifre:
@varjolintu Oh - interesting result! Using "Mehr anzeigen" in step 2 even removes the password field from being selectable, i.e nothing is selectable anymore.

varjolintu:
@olifre Then this must be related to https://github.com/keepassxreboot/keepassxc-browser/issues/508. We are probably changing the "Show more" selection to include any possible input field instead of using the ones from the list of acceptable ones.

olifre:
@varjolintu Ok, understood! I presumed this issue did not apply here, since the password-field on the page in question is of type `password` and the field below (which is only selectable in step 1) is of type `text`, so both should already work. But in an overhaul of `Show more` this issue will probably be fixed for good.

varjolintu:
Strange. That should work normally. I'm mostly using the page https://meine.deutsche-bank.de/trxm/db/ for testing this.

olifre:
@varjolintu For that page, I also have an interesting effect... 
After selecting any field for the username, only the "PIN" field stays selectable. Pressing `Show more` then hides the PIN field and shows the remaining fields. Very strange.

olifre:
@varjolintu I tried with an empty profile (only KeepassXC Browser plugin installed) and observe the same (just to make sure, no other plugins appear to interfere).

varjolintu:
I can reproduce it. Need to inspect this further later.

olifre:
Sorry for bumping - did you already have a chance to look at it? 
For me, that's a pretty blocking regression, since for several pages with extra fiels, the extension fails. 
Let me know if I can help (but I'm rather a C++ guy...) or if we should track this as a separate issue (since this PR is closed, I fear the issue may get lost).

varjolintu:
I haven't fixed this yet.

varjolintu:
Here's the fix: #530.

olifre:
@varjolintu Perfect, many thanks, I'll test it in 1.4.4 then once it comes out and report back :smile:.

PR #54, opened 2018-03-07
Add a new option showLoginNotifications

varjolintu:
Adds a new option `Show notification also when new credentials are detected`. Earlier the only possibility to disable this (which can be an annoying notification) was to disable all notifications. This ensures that error notifications (see https://github.com/keepassxreboot/keepassxc-browser/pull/53) are still shown correctly even when login notifications are hidden.

### DIFF
--- a/keepassxc-browser/options/options.html
+++ b/keepassxc-browser/options/options.html
@@ -130,7 +130,12 @@
             <label class="checkbox">
               <input type="checkbox" name="showNotifications" value="true" /> Show notifications.
             </label>
-            <span class="help-block">Allow KeePassXC-Browser to show notifications when user interaction is needed from the extension icon.</span>
+            <span class="help-block">Show notifications for errors and when user interaction is required.</span>
+          </div>
+          <div class="checkbox">
+            <label class="checkbox">
+              <input type="checkbox" name="showLoginNotifications" value="true" /> Show a notification when new credentials can be saved to the database.
+            </label>
           </div>
         </p>
         <hr />
--- a/keepassxc-browser/background/browserAction.js
+++ b/keepassxc-browser/background/browserAction.js
@@ -239,7 +239,7 @@ browserAction.setRememberPopup = function(tabId, username, password, url, userna
 
         browserAction.show(null, {'id': id});
 
-        if (page.settings.showNotifications) {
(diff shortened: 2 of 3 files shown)


keepassxc-browser/options/options.html
```diff
@@ -130,7 +130,12 @@ <h2>General Settings</h2>
             <label class="checkbox">
               <input type="checkbox" name="showNotifications" value="true" /> Show notifications.
             </label>
-            <span class="help-block">Allow KeePassXC-Browser to show notifications when user interaction is needed from the extension icon.</span>
+            <span class="help-block">Allow KeePassXC-Browser to show notifications about alerts and when user interaction is needed from the extension icon.</span>
```
droidmonkey: This is terribly worded. I have no idea what this setting is for after reading this.

keepassxc-browser/options/options.html
```diff
@@ -130,7 +130,12 @@ <h2>General Settings</h2>
...
+          </div>
+          <div class="checkbox">
+            <label class="checkbox">
+              <input type="checkbox" name="showLoginNotifications" value="true" /> Show notification also when new credentials are detected.
```
droidmonkey: Same here, the wording does not convey the intent.

keepassxc-browser/options/options.html
varjolintu: Maybe something like: "Show notifications on error situations or when user interaction is needed."?

keepassxc-browser/options/options.html
varjolintu: Maybe this could be better: "Show notifications when unsaved credentials are detected."?

droidmonkey: How about "Show a notification when new credentials can be saved to the database"

keepassxc-browser/options/options.html
droidmonkey: "Show notifications for errors and when user interaction is required"

keepassxc-browser/options/options.html
varjolintu: Done.

droidmonkey: Thanks

Issue #83, opened 2018-03-18
Always choose correct database, even when different one is selected


rugk:
## Expected Behavior

The new browser extension should be able to connect to the "right" database always.

If one wants to use different databases at the same time, a switch (or even better: combined UI, so you can use passwords from both databases at the same time) should be integrated into the browser extension.

## Current Behavior

When switching tabs, it only uses (/tries to use) passwords from the currently active database.

## Possible Solution

Short term: Always try to use the database, for which a connection exists. If multiple connections exist for a database, you can keep the old behavior and use the "current" database.

Long term:

New UI addition in browser extension:
* Use database1
* Use database2
* …
* Use all databases

When the last option is selected, it should separate the entries from the different databases.

It should not prevent me from having databases open, which I do not want to connect to KeePassXC-Browser.

## Steps to Reproduce (for bugs)

1. Open a database.
2. Connect to it with browser ext.
3. Open another database.
-> Cannot get entries from first database.

## Debug info

KeePassXC - Version 2.3.1
KeePassXC-Browser Version: 1.0.1
Operating system: Linux
Browser: Firefox
Proxy used: yes

varjolintu:
What about the option in the advanced tab: `Search in all opened databases for matching credentials`?

rugk:
Ah, thanks. Looks what I want to do, but unfortunately it does not work.
I just did the following:
1. Have two databases open. One paired and already connected. The other is not used for thew browser.
2. Enable that option.
3. Switch to the non-browser database in KeePassXC.
3. Get back into Firefox and check the connection.

The connection is still lost/requires me to reconnect (to the new database).

varjolintu:
Of course you need to have the both databases connected. Searches cannot be made to a non-connected database.

rugk:
And I do not want that. It should just ignore not-connected databases.

varjolintu:
Just confirmed it. It really searches even the not-connected databases. I'll do the fix.

PR #108, opened 2018-04-03
Fixes search all databases

varjolintu:
With the current implementation (also valid for passifox/chromeipass/KeePassHTTP-Connector) searching from all databases returns entries also from non-connected databases. This change passes the connected database id's and key's to `get-logins`. A modified handling mechanism is needed to KeePassXC side that ignores non-connected databases based on this new array.

Solves https://github.com/keepassxreboot/keepassxc-browser/issues/83.

This PR can be merged after the KeePassXC side. Doesn't affect the functionality with the older releases.


varjolintu:
KeePassXC side fix: https://github.com/keepassxreboot/keepassxc/pull/1810.

Issue #72, opened 2018-03-13
Fill fields using shortcut does not work on Mac


GMTA:
## Expected Behavior
I should be able to use the Alt + Shift + U or Alt + Shift + P shortcuts to automatically fill in the username and password or just the password respectively on a macOS system.

## Current Behavior
The shortcuts, they do nothing. Instead they insert characters like "¨∏". Using the context menu to fill in the fields _does_ work.

## Possible Solution
* Change the shortcuts to something that works
* Show the correct shortcuts in the Settings dialog

## Steps to Reproduce (for bugs)
1. Go to a username and password form where normal username and password filling through the context menu works as intended.
2. Use the Alt + Shift + U/P shortcut

## Debug info
No errors/warnings in console.

KeePassXC - 2.3.1
keepassxc-browser - 1.0.1
Operating system: Mac
Browser: Firefox
Proxy used: No

varjolintu:
Actually the shortcuts for macOS use MacCtrl instead of Alt. Try them out.

GMTA:
So indeed Ctrl + Shift + U fills out my username. Ctrl + Shift + P does not work, however.

Two issues still remain:
* The text in the Settings dialog is wrong on Mac
* The Ctrl + Shift + P shortcut does not work

varjolintu:
Does it matter if you do the filling on the password or usename field?

You are right, the shortcut descriptions are wrong with macOS.

GMTA:
Ctrl + Shift + P does not work on either the username or password field.

varjolintu:
Strange. Not for me too. I wonder if Firefox has changed some shortcuts that overrides that one. I'll get back to it.

PR #111, opened 2018-04-04
Modify shortcuts

varjolintu:
Alt+Shift+P doesn't work with Firefox. Changed the shortcut to Alt+Shift+I. Also show correct shortcuts for macOS in the settings page.

Solves https://github.com/keepassxreboot/keepassxc-browser/issues/72.

### DIFF
--- a/keepassxc-browser/options/options.html
+++ b/keepassxc-browser/options/options.html
@@ -29,9 +29,9 @@
         <h2>General Settings</h2>
         <hr />
         <p>
-          If you just want to insert username + password into the fields where your focus is, press <code>Alt + Shift + U</code>.
+          If you just want to insert username + password into the fields where your focus is, press <code><span id="mac-user-shortcut">Ctrl + Shift + U</span><span id="default-user-shortcut">Alt + Shift + U</span></code>.
           <br />
-          If you only want to insert the password, just press <code>Alt + Shift + P</code>.
+          If you only want to insert the password, just press <code><span id="mac-pass-shortcut">Ctrl + Shift + I</span><span id="default-pass-shortcut">Alt + Shift + I</span></code>.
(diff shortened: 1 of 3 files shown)


keepassxc-browser/options/options.html
```diff
@@ -29,9 +29,9 @@ <h3 class="muted"><img src="/icons/keepassxc_48x48.png" alt="logo" /> KeePassXC-
         <h2>General Settings</h2>
         <hr />
         <p>
-          If you just want to insert username + password into the fields where your focus is, press <code>Alt + Shift + U</code>.
+          If you just want to insert username + password into the fields where your focus is, press <code><span id="mac-user-shortcut">Ctrl + Shift + U</span><span id="default-user-shortcut">Alt + Shift + U</span></code>.
```
droidmonkey: Doesn't Mac use "Cmd" instead of Ctrl?

keepassxc-browser/options/options.html
```diff
@@ -29,9 +29,9 @@ <h3 class="muted"><img src="/icons/keepassxc_48x48.png" alt="logo" /> KeePassXC-
...
           <br />
-          If you only want to insert the password, just press <code>Alt + Shift + P</code>.
+          If you only want to insert the password, just press <code><span id="mac-pass-shortcut">Ctrl + Shift + I</span><span id="default-pass-shortcut">Alt + Shift + I</span></code>.
```
droidmonkey: Same as above

keepassxc-browser/options/options.html
varjolintu: Using Cmd with macOS can trigger browser's own shortcuts. In this case Ctrl is actually MacCtrl.

Issue #70, opened 2018-03-12
User/password input field detection doesn't work for dropdown menus


cryptomilk:
User/password input field detection doesn't work for drop down menus when credentials for the site are stored in keepassxc.

A good example is:

https://build.opensuse.org/

## Debug info
KeePassXC - 2.3.1
keepassxc-browser - 1.0.1
Operating system: Linux
Browser: Firefox 58.0.2
Proxy used: NO

varjolintu:
Does redetecting credential fields help?

cryptomilk:
Yes, it does.

varjolintu:
If you save those specified credential fields does it solve the issue?

cryptomilk:
I need to 'redetect credential fields' each time I want to log in.

varjolintu:
Cannot reproduce this with Chromium. The dropdown login menu works just fine without any tweaking. However I could reproduce this with Firefox.

varjolintu:
I know a solution to this but I need to make some more testing. I don't like the idea of looping `setInterval`.

cryptomilk:
Thanks for looking into this!

PR #117, opened 2018-04-09
Improved dynamic input field detection

varjolintu:
Replaces the old `setInterval` based input field detection polling with `MutationObserver`. Now any new input fields shown are detected immediately.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/70.

Also solves input field detection for many other pages.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -1140,7 +1140,34 @@ cipFields.useDefinedCredentialFields = function() {
     return false;
 };
 
+MutationObserver = window.MutationObserver || window.WebKitMutationObserver;
 
+// Detects DOM changes in the document
+let observer = new MutationObserver(function(mutations, observer) {
+    if (document.visibilityState === 'hidden') {
+        return;
+    }
+
+    for (const mut of mutations) {
+        if (mut.type === 'childList') {
+            const fields = cipFields.getAllFields();
+
+            // If only password field is shown it's enough to have one field visible for initCredentialFields
+            if (fields.length > (_detectedFields == 1 ? 0 : 1)) {
+                cip.initCredentialFields(true);
+            }
+        }
+    }
+});
+
+// define what element should be observed by the observer
+// and what types of mutations trigger the callback
+observer.observe(document, {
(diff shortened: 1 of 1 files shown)


keepassxc-browser/keepassxc-browser.js
```diff
@@ -1126,7 +1126,34 @@ cipFields.useDefinedCredentialFields = function() {
...
 
+// define what element should be observed by the observer
+// and what types of mutations trigger the callback
+observer.observe(document, {
+    subtree: true,
+    attributes: true,
+    childList: true,
+    characterData: true,
+    attributeOldValue: true,
+    characterDataOldValue: true
```
droidmonkey: This list contains unnecessary items. I don't believe listening for characterdata, attributeoldvalue, and characteroldvalue would ever be involved in the appearance of new form fields. Also, is it possible to only listen for mutations on existing (and new) form blocks? It seems very wasteful to listen to the entire document.

varjolintu: True, I removed a few of those. The [documentation](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#MutationObserverInit) says:
```
Note: At the very least, childList, attributes, or characterData must be set to true. Otherwise, "An invalid or illegal string was specified" error is thrown.
```

varjolintu: Also, listening the entire document is necessary because otherwise new div's added to it wouldn't be detected at all.

keepassxc-browser/keepassxc-browser.js
```diff
@@ -1126,7 +1126,32 @@ cipFields.useDefinedCredentialFields = function() {
     return false;
 };
 
+MutationObserver = window.MutationObserver || window.WebKitMutationObserver;
+
+// Detects DOM changes in the document
+let observer = new MutationObserver(function(mutations, observer) {
+    if (document.visibilityState !== 'hidden') {
```
phoerious: I would prefer you restructure this whole callback function with return guards. In JS it's really tempting to create deeply nested structures, which are hard to understand.

Issue #120, opened 2018-04-13
Auto fill HTTP Auth: When multiple logins - No indication of HTTP auth and cannot enter manually


ProCycleDev:
##Expected Behavior

When you visit a site with a HTTP Auth dialog but there is more than one possible entry in the database a notification appears telling you it is waiting on HTTP authentication. There is a button in the addon icon bubble to summon the manual entry box.

## Current Behavior

When you visit a site with a HTTP Auth dialog but there is more than one possible entry in the database then the page just sits there loading forever. There is a notification that tells you "Multiple credentials detected" but it does not say you encountered an HTTP Auth dialog. Additionally if none of the login choices in the database are for the HTTP Auth then there is no way to manually enter a login.

## Possible Solution

An notification should appear telling you it is waiting on HTTP authentication. When you choose an entry from the addon icon bubble there should be an option to show the HTTP Auth Dialog.

## Steps to Reproduce (for bugs)

1. Add two entries to your database with the same url that asks for HTTP authentication. Ensure neither entry has the correct credientials
2. Attempt to visit the website.
3. Notification appears telling you there are multiple credentials detected
4. You cannot login because there is no way to show the manual entry dialog. If you didn't expect HTTP authentication then you sit there waiting for the page to load and it never does.

## Debug info

KeePassXC - 2.3.1
keepassxc-browser - 1.0.1
Operating system: Windows 7
Browser: Firefox
Proxy used: NO

varjolintu:
I made some improvements to this and added a dismiss button to the popup dialog, plus modified the notification text. Basicly the dismiss button tries to add empty credentials to HTTP auth. Basically you could do the same by just selecting wrong credentials and then the manual popup arrives, but maybe a separate button is more clear :)

gerritgriebel:
JFYI: I had exactly the described faulty behavior: Login form of a test site is protected by http basic auth. So I have two Logins, one for Login form and another one for basic auth. I can not login, it hangs after selecting the right login in basic auth attempt and I get above mentioned notification. My workaround was to disable HTTP Auth in Chrome Extension and it store HTTP-Password in Chrome Passwords.  I'm using todays latest versions, as I switched today to KeePassXC:

KeePassXC - 2.5.4
KeePassXC-Browser - 1.6.2
Operating system: MacIntel
Browser: Chrome/Chromium 81.0.4044.122

varjolintu:
[...]

PR #121, opened 2018-04-14
HTTP Auth fixes

varjolintu:
Adds a Dismiss button to HTTP Authentication dialog that allows a manual entry to the auth dialog.
Also the login dialog CSS style is more simple and clear.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/120.


keepassxc-browser/popups/popup_httpauth.html
```diff
@@ -29,7 +29,10 @@
         <p>
           Select the login information you would like to get logged in with:
         </p>
-        <ul id="login-list"></ul>
+        <div id="login-list" class="list-group"></div>
+        <p>
+          <button id="btn-dismiss" class="btn btn-sm btn-danger"><span class="glyphicon glyphicon-remove"></span> Dismiss and show the manual authentication dialog</button>
```
phoerious: What is a 'manual authentication dialog'?

keepassxc-browser/popups/popup_httpauth.js
```diff
@@ -1,35 +1,55 @@
...
+                resolve(global.page.tabs[tabs[0].id].loginList);
+            });
         });
     });
+};
+
+$(function() {
+    getLoginData().then((data) => {
+        let ll = document.getElementById('login-list');
+        for (let i = 0; i < data.logins.length; i++) {
```
phoerious: ++i

keepassxc-browser/popups/popup_httpauth.html
varjolintu: Maybe it should be "default" instead.

phoerious:
Approved, but needs conflict resolution.

PR #124, opened 2018-04-16
Use Zepto.js instead of jQuery where possible

mehrdadn:
Hi! This is a `keepassxc-browser` version of pfn/passifox#670.  
Please see that PR for details. :) Thank you!


varjolintu:
What are the benefits of Zepto.js when compared to jQuery? I would actually prefer using vanilla JS instead of any 3rd party libraries where possible.

mehrdadn:
It loads *much* faster than jQuery. I would prefer vanilla JS too, but that's a lot more work!
The performance hit is non-negligible when there are lots of subframes (passifox/issues#550).

mehrdadn:
^Oops... fixed link to issue pfn/passifox#550 in my reply just now. (Will delete this.)

varjolintu:
I'd be interested about this if it would replace jQuery totally in the content script(s) (`keepassxc-browser.js`) but this doesn't seem to do so. Maybe this could be improved some more.

mehrdadn:
The problem is there is also jQuery-UI. How would you get rid of that? Write a new GUI library?

varjolintu:
That's the biggest problem yes. Actually there is a possibility to use [dataset](https://www.w3.org/TR/html5/forms.html#the-datalist-element) instead of autocomplete (I tried it but didn't get it to work 100% identically). But for the password generator you would have to use some magic.

mehrdadn:
You're more than welcome to do the surgery if you'd like (you should have the ability to edit my branch). If you wait for me to do it, there's a good chance this project will be dead by the time I get around to it. (Not really joking—notice that issue pfn/passifox#550 was posted in Sep. 2016, and by the time I turned it into the pfn/passifox#670 PR, it was Jan. 2018 and the project was pretty dead.) :\

varjolintu:
I will definitely do some surgery because I would like to get rid of jQuery in content scripts. While Zepto.js is a good replacement, we would need to get rid of jQuery-UI too. I appreciate the help :) It's a good start.

varjolintu:
Here's a branch that might interest you: https://github.com/keepassxreboot/keepassxc-browser/tree/no_jquery.

Autocomplete is replaced by HTML5 dataset (works a bit differently, for example it needs a second click to username field). The only things not working with this one are choosing own credential fields and password generator dialog.

mehrdadn:
Wow, that's awesome! It seemed to load 2.5× faster than my solution when I just tested it!
The two downsides you mentioned don't affect me too much personally since I try to use the keyboard shortcut and don't really use the password generator dialog, but yeah, I can imagine they'd be showstoppers for some people.
Thanks a lot for doing this! I'll start using it now. :)

varjolintu:
No problem :) I probably replace the password generator and other UI things with my own implementation. Trying to avoid any 3rd party UI library. Just keeping it minimal, even if the amount of code increases.

mehrdadn:
Yeah I completely agree! That's definitely what I would do too :)

mehrdadn:
By the way, another change I would very much urge here (which I also tried to do in my commit) is to use `MutationObserver` instead of a timer to react to DOM changes. It's the right way to do it and it'll be more snappy. And since the DOM can change at any time, it'd be able to observe that indefinitely without burning CPU unnecessarily.

mehrdadn:
Update: It turns out this branch is really flaky. On a ton of sites (not all though), it often (but not always) just refuses to fill in some or all of the fields. When I revert to my own version with Zepto.js, it seems to just work fine.

As an example, when I tried https://login.yahoo.com, it refused to fill in anything earlier today (even though it clearly asked me to unlock the database). I tried it again just now, and it filled in the username but didn't fill in anything for the password. I tried my version, and it worked fine. I reverted back to this branch and it (again) refused to fill in both the username and the password.

I haven't had a chance to track down the bug unfortunately but I thought I'd let you know it doesn't seem usable at the moment.

varjolintu:
Well, this is still work-in-progress because I know there are still many bugs left. I haven't even tried two page login sites yet. Thanks for letting me know. Also the Yahoo login page creates a password generator icon even when there's no password field visible. So time to do some more fixes.

varjolintu:
@mehrdadn Btw, the autocomplete is not going to work with password fields without jQuery because the `list` attribute is not supported with `type="password"`. That probably explains the error situation. Filling it from the context menu or keyboard shortcut however works fine.

varjolintu:
I'm closing this PR as I've made such a good progress with jQuery removal. You are welcome to collaborate.

PR #126, opened 2018-04-18
Automatic reload

varjolintu:
Tries reload automatically when KeePassXC is not started.

### DIFF
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -838,20 +838,7 @@ keepass.onNativeMessage = function(response) {
     if (response.action === kpActions.DATABASE_LOCKED || response.action === kpActions.DATABASE_UNLOCKED) {
         keepass.testAssociation((associationResponse) => {
             keepass.isConfigured().then((configured) => {
-                let data = page.tabs[page.currentTabId].stack[page.tabs[page.currentTabId].stack.length - 1];
-                data.iconType = configured ? 'normal' : 'cross';
-                browserAction.show(null, {'id': page.currentTabId});
-
-                // Send message to content script
-                browser.tabs.query({ active: true, currentWindow: true }).then((tabs) => {
-                    if (tabs.length) {
-                        browser.tabs.sendMessage(tabs[0].id, {
-                            action: 'check_database_hash',
-                            hash: {old: kpxcEvent.previousDatabaseHash, new: keepass.databaseHash}
-                        });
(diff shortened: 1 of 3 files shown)


Mogost:
Good feature!

keepassxc-browser/background/keepass.js
```diff
@@ -996,3 +996,38 @@ keepass.decrypt = function(input, nonce) {
...
+        }
+    }, 1000);
+};
+
+keepass.reconnect = function(callback, tab) {
+    return new Promise((resolve, reject) => {
+        keepass.generateNewKeyPair();
+        keepass.changePublicKeys(tab).then((pkRes) => {
+            keepass.getDatabaseHash((gdRes) => {
+                if (gdRes) {
```
phoerious: Negate and return. This gets deep enough already.

keepassxc-browser/background/keepass.js
```diff
@@ -1005,3 +1005,41 @@ keepass.decrypt = function(input, nonce) {
...
+};
+
+keepass.reconnect = function(callback, tab) {
+    return new Promise((resolve, reject) => {
+        keepass.generateNewKeyPair();
+        keepass.changePublicKeys(tab).then((pkRes) => {
+            // Database hash should be received if the reconnection succeeded
+            keepass.getDatabaseHash((gdRes) => {
+                if (!gdRes) {
+                    reject(false);
```
phoerious: Do we need a return after this?

phoerious: Without a return, keepass.testAssociation will still be called after rejection and may produce ugly side effects.

varjolintu: Debugged it and that indeed happens.. added the return.

phoerious: You cannot change the state of a promise once it's been rejected or resolved, but the code after the call to reject() or resolve() is still being executed.

Ealhad:
resolved #264

varjolintu:
This can be re-merged after 2.3.4 is released. However, this means KeePassXC-Browser should set 2.3.4 as the minimum version.

varjolintu:
Reminder for myself:
Disable reconnect if `navigator.platform` is `Win32` or `Windows` and KeePassXC version is lower than 2.3.4. With this there's no need to increase the minimum version. A mention to the documentation is needed. And maybe an additional error message to the extension.

PR #128, opened 2018-04-19
No jQuery

### LINKED ISSUE #50: XML pretty-print gets broken by the extension (Firefox)
## Expected Behavior
Pretty print should work like it does on xml documents

## Current Behavior
image
This issue is mitigated partially by https://github.com/keepassxreboot/keepassxc-browser/blob/develop/keepassxc-browser/manifest.json#L45-L48 so it was "fixed" on documents with .xml and .xsd endings but is still present when the xml document doesn't contain that extension.
Some more information (e.g. that jQuery is the issue) is at https://github.com/smorks/keepasshttp-connector/issues/32

1. Open an xml document without the .xml or .xsd extension with the extension enabled

### LINKED ISSUE #264: FR: Autoreload
Hi

It would be very nice if the addon does reloads regularly, and once it find an open databse it connects to it automatically. At the moment when it looses connection to the database I have to first reload from the interface and I need to click connect database.

Can these steps be automated?

thanks

### LINKED ISSUE #295: Why does this extension load jquery and jquery ui on EVERY page?
I'm looking for an alternative to something like LastPass, thought keepassxc looked pretty lightweight and seems to work well, but upon inspecting the codebase for this extension it's loading jQuery and jQuery UI on EVERY page... the question is, why? I can understand it being a requirement for the options page, possibly, but certaintly not needed for the main frontend functionality of filling form elements etc.

image

In theory, NOTHING except bare minimal functionality should be loaded until;

1. A direct action from the user telling it to fill passwords, or perform some action from the chrome extension bar
2. A password / credential type fields have been detected on the page

### LINKED ISSUE #328: Extension blocking date-picker (calendar) on thalys.com
## Expected Behavior
When going to www.thalys.com, on the homepage it should display a date-picker (or: calendar) when clicking on the date field (My dates), as follows:
[Date picker](https://user-images.githubusercontent.com/12184325/47921764-10b12f00-debe-11e8-99e1-1283bba22129.gif)

## Current Behavior
Instead of opening the calendar, nothing happens upon clicking the date field. I figured it was KeePassXC because I created a fresh Firefox profile and added & removed extensions one by one (many thanks to the people over at Webcompat to [help debug the issue](https://github.com/webcompat/web-bugs/issues/20641l)).

## Debug info
KeePassXC - 2.3.4
KeePassXC-Browser - 1.3.0
Operating system: Linux (Ubuntu 18.04)
Browser: Firefox 63.0

### LINKED ISSUE #372: KeePassXC-Browser breaks PrimeNG dropdown component
I use PrimeNG components in a web application I'm developing and I noticed that when I had the KeePassXC-Browser extension enabled their dropdown component rendered incorrectly.

My initial tests were on a work-issued laptop running the current, up-to-date, Chrome on Windows 7 - but I leave that computer in the office, so I can't reproduce screenshots & version numbers from that one just now. Luckily (or un-luckily, I suppose) the bug is observable in Firefox & Chromium on my personal Debian laptop. The following screenshots and versions are from that computer.

I initally diagnosed it as an extension bug, rather than my own broken css, by reloading the page in an incognito window with no extensions. I then narrowed it down by turning off all other extensions and turning them on and off, one-by-one, to see which extension caused the corruption. Their appears to be a conflict between PrimeNG components and KeePassXC-Browser.

## Expected Behavior

The PrimeNG dropdown component should work just like it does when the extension isn't installed or enabled.

See the full-width dropdown "Select a city" under the "Single" heading:

keepassxc--003

See the full-width options under the "Select a city" dropdown:

keepassxc--004

## Current Behavior

When the extension is installed and enabled, the PrimeNG downdown component is corrupted.

See the truncated, broken dropdown under the "Single" heading:

keepassxc--001

See the truncated options under the truncated, broken dropdown:

keepassxc--002

## Steps to Reproduce
1. Disable the KeePassXC-Browser extension.
2. Visit [the PrimeNG dropdown component demo page](https://www.primefaces.org/primeng/#/dropdown).
3. Observe the component render correctly.
4. Enable the KeePassXC-Browser extension.
5. Refresh the page.
6. Observe the broken component.

## Debug info
KeePassXC - 1.3.0
KeePassXC-Browser - 2.3.4
Operating system: Debian GNU/Linux buster/sid
Browser: Firefox Developer Edition 65.0b8 (64-bit) &
Browser: Chromium Version 70.0.3538.67 (Developer Build) built on Debian buster/sid, running on Debian buster/sid (64-bit)

An example of the bug in Chromium on Linux can be seen below too:

keepassxc--005

### LINKED ISSUE #420: Cannot set property 'cssText' of undefined
## Expected Behavior

The extension never throws an error.

## Current Behavior

Sometimes, jQuery-UI throws an error. It's actually [this line](https://github.com/jquery/jquery-ui/blob/1.12.1/ui/effect.js#L180) that throws the error:

> Uncaught TypeError: Cannot set property 'cssText' of undefined
>    at jquery-ui.min.js:13
>    at jquery-ui.min.js:13
>    at jquery-ui.min.js:6
>    at jquery-ui.min.js:6

This happened on my website (not public, sorry) when there is HTML inserted into the DOM (via `element.innerHTML = '...';`) after the page was loaded.

## Possible Solution

Get rid of jQuery-UI. (Why is this used anyway?)

## Debug info

KeePassXC - 2.3.1
KeePassXC-Browser - 1.3.2
Operating system: Linux
Browser: Chromium

### LINKED ISSUE #422: Connect/reconnect when user uses fill option in right-click menu
If a user is a on a website login page but has not opened KeepassXC yet they will need to open their DB, click connect/reconnect in keepassxc-browser and if the credentials aren't showing up then they have to click redetect logins (or in my experience more than usual you have to reload the page if the login fields don't show any username/password or credentials dropdown automatically after reconnecting). 

If a user explicitly clicks any fill option (like Fill Username/Password) and the DB is not recognized as connected then keepassxc-browser should try to connect/reconnect to the DB and perform the fill action. This makes it easier to use and saves not doing a few of the steps mentioned above that can be done automatically when using the fill options.

varjolintu:
The goal is to remove jQuery totally from the content scripts. Loading jQuery for every page (and iframe) is consuming and slows down browsing. Vanilla JavaScript is preferred instead.

Autocomplete has the following improvements:
- First item is always selected by default
- It can be used without a mouse: <kbd>Backspace</kbd> or <kbd>Delete</kbd> opens the list when input field has no value. Arrows and <kbd>Enter</kbd> selects the credentials. Using <kbd>Escape</kbd> or <kbd>Tab</kbd> (focus disappears) closes the autocomplete.

Password generator shows a "Try again" button when KeePassXC is closed.
A small generic UI library has been written for dynamic creation of elements.

On my machine the basic task for retrieving credentials for a page is much faster. Monitoring the performance shows that the current develop branch with jQuery takes about:
- 3 ms for loading
- 67 ms for scripting
- 25 ms for other stuff

With this PR:
- 2 ms for loading
- 22 ms for scripting
- 14 ms for other

The current status is:
- [x] Replace all functions from keepassxc-browser.js with vanilla JS
- [x] New autocomplete
- [x] New password generator
- [x] New implementation for choosing own credential fields
- [x] Translations
- [x] Fix autocomplete
- [x] Rebase
- [x] Implement automatic reconnect
- [x] Test everything

Fixes #50.
Fixes #295.
Fixes #328.
Fixes #372.
Fixes #420.
Fixes #264.
Fixes #422.


varjolintu:
The new password generator works now! There might be a few bugs left. Trying to find and solve them soon.

varjolintu:
HTML5 dataset cannot be used in password input fields, and with username fields user needs to click twice before the autocomplete menu appears. Maybe we should implement our own autocomplete based on the jQuery-UI code?

droidmonkey:
Im ok with that. It's there no light weight alternative to jQuery?

varjolintu:
@droidmonkey There is. See #124. But still, vanilla JavaScript is faster. And when talking about content scripts, the faster the better.

varjolintu:
New autocomplete implementation has been added. Still needs some fixing because it doesn't work with all dynamic sites. I added some new functionality too: `esc` causes the autocomplete list to close and `backspace` shows the list when input field has no value. So you can use it only with keyboard too.

keepassxc-browser/keepassxc-browser.js
```diff
@@ -825,51 +191,59 @@ cipFields.getAriaHidden = function(field) {
...
+    while (field.parentElement) {
+        parents.push(field = field.parentElement)
+    }
+
+    for (const p of parents) {
+        if (p.style.overflow === 'hidden') {
             return true;
         }
     }
     return false;
```
thetric: `cipFields.getAriaHidden()` and `cipFields.getOverflowHidden()` can be refactored into:

```js
/**
 * Check aria-hidden attribute by looping the parent elements of input field
 * @param {HTMLElement} field
 * @returns {string}
 */
cipFields.getAriaHidden = function(field) {
  return traverseParents(field, val => val, val => val, () => "false");
};

/**
 * @param {HTMLElement} field
 * @returns {boolean}
 */
cipFields.getOverflowHidden = function(field) {
  return traverseParents(field, f => f.style.overflow === "hidden");
};

/**
 * Returns the first parent element satifying the {@code predicate} mapped by {@code resultFn} or else {@code defaultVal}.
 * @param {HTMLElement} element the start element (excluded, starting with the parents)
 * @param {function} predicate matcher for the element to find, type (HTMLElement) => boolean
 * @param {function} resultFn callback function of type (HTMLElement) => {*} called for the first matching element
 * @param {fun} defaultValFn fallback return value supplier, if no element matching the predicate can be found
 */
function traverseParents(
  element,
  predicate,
  resultFn = () => true,
  defaultValFn = () => false
) {
  for (let f = field.parentElement; f !== null; f = f.parentElement) {
    if (predicate(f)) {
      return resultFn(f);
    }
  }
  return defaultValFn();

```

keepassxc-browser/pwgen.js
```diff
@@ -0,0 +1,383 @@
...
+    const className = (isFirefox() ? 'key-moz' : 'key');
+    const size = (field.offsetHeight > 28) ? 24 : 16;
+    let offset = Math.floor((field.offsetHeight - size) / 3);
+    offset = (offset < 0) ? 0 : offset;
+
+    const icon = kpxcUI.createElement('div', 'kpxc kpxc-genpw-icon ' + className,
+        {'title': 'Generate password', 'size': size, 'offset': offset, 'kpxc-genpw-field-id': field.getAttribute('kpxc-id')});
+    icon.style.zIndex = '9999';
+    icon.style.width = String(size) + 'px';
+    icon.style.height = String(size) + 'px';
```
thetric: a simple string concatenation is totally safe

```js
icon.style.width = size + "px";
```

thetric:
Perhaps the jquery js and css files should be deleted, too?

keepassxc-browser/keepassxc-browser.js
varjolintu: Good idea. Makes it more readable too.

keepassxc-browser/pwgen.js
varjolintu: Just wanted to make sure it's clear what's happening here. Also, see [this](https://github.com/airbnb/javascript#type-casting--coercion).

varjolintu:
@thetric They will be deleted. This is still WIP so lots of things are missing.

keepassxc-browser/ui.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+// jQuery style wrapper for getElementById()
+var $ = function(id)
```
rugk: Why only ID? Actually you could use `querySelector` which would be more jQuery-like…

rugk:
Generally you may switch to [plainJS](https://plainjs.com/) for all files in general. Modern browsers do not require jQuery anymore, especially in browser extensions, where you can control which browsers it is using.

keepassxc-browser/ui.js
rugk: Also the syntax is different: For query you of course need to do `$('#id')`. Your current implementation requires the ID without a hash.

But generally, do you really need to do it anyway? I think using the plain JS functions is easy enough and you also seem to do so everywhere in this PR at least.

varjolintu: Plain JS is the way to go. Writing a small wrapper just makes the code more readable.

phoerious: Just make $ an alias for document.querySelector.

keepassxc-browser/popups/popup.html
```diff
@@ -95,5 +89,12 @@
         </div>
       </div>
     </div>
+
+    <script type="text/javascript" src="../browser-polyfill.min.js"></script>
+    <script type="text/javascript" src="../global.js"></script>
+    <script type="text/javascript" src="../jquery-3.3.1.min.js"></script>
+    <script type="text/javascript" src="../options/bootstrap.min.js" /></script>
+    <script type="text/javascript" src="popup_functions.js" /></script>
+    <script type="text/javascript" src="popup.js" /></script>
```
rugk: For that, you can also easily use the [defer attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) instead of moving stuff around. 😄

keepassxc-browser/content/define.js
```diff
@@ -191,7 +191,7 @@ kpxcDefine.markAllPasswordFields = function(chooser) {
         const field = e.currentTarget;
         kpxcDefine.selection.password = field.getAttribute('kpxc-id');
         field.classList.add('kpxcDefine-fixed-password-field');
-        field.innerHTML = 'Password';
+        field.textContent = 'Password';
```
rugk: Yeah, always get rid of `innerHTML`. AMO's security check will like you… :smile: :+1:

varjolintu: Indeed. I learned this the hard way..

varjolintu:
Latest PR's included in the branch. Autocomplete still needs some work, and I want to get rid of w3c.css as it might break some sites.

varjolintu:
Custom credential fields and password generator remade without W3.CSS. Next I'll try to improve the autocomplete.

varjolintu:
The autocomplete has been fixed. Some related comments in the PR text.

mehrdadn:
Hi! Just thought I'd let you know, I tried out commit ec47413 and it fails to fill in fields too often to be usable. It also seems to be very fragile and often desiring to re-connect to the database (KeePass) with a new key despite being already connected before.

I've always had these kinds of problems whenever I've tried to check back for updates here over the course of the past months, so they're not new. I would recommend trying out the code extensively on many websites. (I'd give specific examples but due to privacy reasons I find it a bit difficult...)

varjolintu:
@mehrdadn Sorry to hear that. I haven't had such problems, and I have a big list of different websites I am testing every time before I push in a commit. 

Currently I'm fixing the small bugs here and there. If you encounter a bug, please debug the content scripts.

mehrdadn:
@varjolintu I just invited you to a private repo where I'll put down a list of websites I'm having trouble with. (I'd debug them myself but I didn't immediately see the problem and don't have the bandwidth to dig deeper at the moment unfortunately.)

varjolintu:
@mehrdadn Just checked those sites. If the page uses a single username field in the first page, please be sure to include the URL to Site Preferences (tab in the extension settings) and enable Username-only detection.

varjolintu:
@z3ntu This can be already tested :) Just trying to do the rebase next so the commit tree would be a little bit cleaner.

mehrdadn:
@varjolintu I'm confused, so are you saying that is your intended workflow? That every user who comes across a username-only page goes into Site Preferences and enters the URL of the page in there? That's impractical, and I don't understand why this should be necessarily. I've certainly never had to do this to get correct behavior before.

varjolintu:
@mehrdadn That's how it's always worked. The reason for the extra settings is that there is no way to determine that a input field with type `text` is actually a username field in a page. This would cause e.g. search fields to trigger credential retrieval requests to KeePassXC. Meaning that if you have a GitHub account, every GitHub page with a search field will be asked for credentials if you haven't accepted them automatically.

There is no proper workaround for this behaviour.

droidmonkey:
I think what he means is that it could be slightly more streamlined (ie, a dedicated button on the drop down menu that adds the site and takes you to the configuration page in one shot).

varjolintu:
@droidmonkey That would be nice indeed.

mehrdadn:
@varjolintu But this was simply not true on, for example, version 1.1.4 (commit 1217769e). On that commit (and I'm testing this right now), websites 1, 2, 4, and 5 worked correctly (website 3 seemed to have been fixed by another local patch I had). However, on commit 1099529, none of them work correctly.

Note that, by "worked correctly", here is what I mean: I do *not* have auto-fill-in enabled. Instead, when I see a username field, I press the keyboard shortcut for filling in the username and password fields. This is the behavior that used to fill in the fields correctly but now doesn't.

P.S. Note that website 3 does have a password field; I believe the issue there is different.

z3ntu:
Weirdly I can't get the extension working (i.e. I get `Timeout or not connected to KeePassXC.` even though I didn't "connect" it with this extension) when I load the git repo as an unpacked extension in Chromium (both develop and no_jquery branches). Is this something that happens when you load it unpacked or is this something else?
image

z3ntu:
Found a problem: on https://banking.raiffeisen.at when you press "Benutzername" (username) you it detects the password field and the username field but when you try to autofill it, I get the popup "Error: Cannot find fields to fill in." even though I can fill it fine through the extension icon.

keunes:
[supposedly also fixes #328]

varjolintu:
@z3ntu Thanks for reporting that. It seems that some characters in the ID are replaced and the matching won't work properly. As a workaround you can choose custom credential fields for that site.

@mehrdadn Thanks for clearning that yout. I thought you was using the autocomplete menu, and the extra Site Preferences stuff is mainly for that. I'll check the PR if it still doesn't work properly.

whisdol:
I tried out 6cfa4ef today and noticed one bug:
The Firefox autocomplete suggestions are now displayed on top of the KeePassXC credential list (left). In the current/JQuery version, the Firefox suggestions are disabled (right).
image
(This is with auto-fill off.)

varjolintu:
@whisdol Thanks for reporting the issues. https://github.com/keepassxreboot/keepassxc-browser/pull/128/commits/177c5322b3f893dc3929a1e74390ae70e29df1a5 fixes the first one. Second one I cannot reproduce. Are you locking it when logins are visible, or just the default popup?

whisdol:
177c532 does indeed fix the autocomplete, thanks!
For me, the popup stays in the connected-mode consistently after clicking the lock button, both when logins are displayed (because the username field is focused) and when no logins are displayed. The extension icon switches to the locked-mode, though:
image

A further note:
After clicking into (/out of) the username field, the logins are only displayed (/hidden) after a noticeable delay (<1s, but noticeable). In the stable/jquery version, there is no delay.

varjolintu:
@whisdol Thanks for reporting this. I'll make the necessary changes. Maybe the reason for the delay is that the list is created/deleted instead of created/hidden like in the jQuery version. That behaviour can be changed of course.

z3ntu:
I had some troubles getting the login information filled in on twitch.tv with the login prompt. Firstly it took forever (surely 10 or 20 seconds) to ask if I want to give access to the login credentials. Then they appeared when clicking on the login field and I could select them but the password didn't get filled in. Then I tried doing a right click on the password field and select "Fill password only" which didn't do anything. Only when clearing the username field again, the autofill worked...

varjolintu:
@z3ntu I have problems with `twitch.tv` even with the latest stable release. When debugging, the extension doesn't even show in the list in content scripts. So something really strange happens here.

keepassxc-browser/_locales/en/messages.json
```diff
@@ -104,7 +104,7 @@
         "description": "Checkbox text below the password generator input field."
     },
     "passwordGeneratorBits": {
-        "message": "??? Bits",
+        "message": " Bits",
```
rugk: Hmm, should not you use proper placeholders for this? See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization#Retrieving_message_strings_from_JavaScript

keepassxc-browser/content/ui.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+// jQuery style wrapper for getElementById()
```
rugk:
```suggestion
// jQuery style wrapper for querySelector()
```

it's not only about IDs

keepassxc-browser/css/keepassxc-browser.css
```diff
@@ -0,0 +1,14 @@
+.cip-genpw-icon {
+    position: absolute;
+    cursor: pointer;
+}
+
+.cip-genpw-icon.key {
+    background: url('chrome-extension://__MSG_@@extension_id__/icons/key.png') right no-repeat;
```
rugk:
```suggestion
    background: url('./icons/key.png') right no-repeat;
```

Why not use a relative path? Should not this work?
Also this obviously leads to the fact that you need to duplicate the work here…

keepassxc-browser/css/keepassxc-browser.css
```diff
@@ -0,0 +1,14 @@
...
+    background-size: contain;
+}
+
+.cip-genpw-icon.key-moz {
+    background: url('moz-extension://__MSG_@@extension_id__/icons/key.png') right no-repeat;
```
rugk: Or better like that, so it's still an absolute path, but without that ugly "protocol":

```suggestion
    background: url('/icons/key.png') right no-repeat;
```

keepassxc-browser/global.js
```diff
@@ -4,9 +4,9 @@ const IGNORE_NOTHING = 'ignoreNothing';
...
 
-var schemeSegment = '(\\*|http|https|ws|wss|file|ftp)';
-var hostSegment = '(\\*|(?:\\*\\.)?(?:[^/*]+))?';
-var pathSegment = '(.*)';
+const schemeSegment = '(\\*|http|https|ws|wss|file|ftp)';
+const hostSegment = '(\\*|(?:\\*\\.)?(?:[^/*]+))?';
+const pathSegment = '(.*)';
 
 var isFirefox = function() {
     if (!(/Chrome/.test(navigator.userAgent) && /Google/.test(navigator.vendor))) {
```
rugk: Wait `isFirefox` returns true, when it is Chrome???  :astonished:

keepassxc-browser/global.js
```diff
@@ -4,9 +4,9 @@ const IGNORE_NOTHING = 'ignoreNothing';
...
 const IGNORE_FULL = 'ignoreFull';
 
-var schemeSegment = '(\\*|http|https|ws|wss|file|ftp)';
-var hostSegment = '(\\*|(?:\\*\\.)?(?:[^/*]+))?';
-var pathSegment = '(.*)';
+const schemeSegment = '(\\*|http|https|ws|wss|file|ftp)';
+const hostSegment = '(\\*|(?:\\*\\.)?(?:[^/*]+))?';
+const pathSegment = '(.*)';
 
 var isFirefox = function() {
```
rugk: Oh, hey, this is not really needed. We are not in ugly websites – we have proper APIs for that. :laughing: 

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBrowserInfo

(Okay, just see browser support is not yet that good, but hey… I assume other browsers implement it too. You may just fallback or so.)

keepassxc-browser/options/options.html
```diff
@@ -5,23 +5,24 @@
...
       <link rel="stylesheet" href="bootstrap.min.css" />
       <link rel="stylesheet" href="options.css" />
-      <script type="text/javascript" src="../browser-polyfill.min.js"></script>
-      <script type="text/javascript" src="../global.js"></script>
-      <script type="text/javascript" src="../jquery-3.3.1.min.js"></script>
-      <script type="text/javascript" src="bootstrap.min.js"></script>
-      <script type="text/javascript" src="options.js"></script>
+      <script src="../browser-polyfill.min.js"></script>
+      <script src="../global.js"></script>
+      <script src="../jquery-3.3.1.min.js"></script>
```
rugk: Hu? Was not this PR about removing jQuery?

```suggestion
```

keepassxc-browser/popups/popup.html
```diff
@@ -1,49 +1,51 @@
...
     <link rel="stylesheet" href="../options/bootstrap.min.css" />
-    <script type="text/javascript" src="../browser-polyfill.min.js"></script>
-    <script type="text/javascript" src="../global.js"></script>
-    <script type="text/javascript" src="../jquery-3.3.1.min.js"></script>
-    <script type="text/javascript" src="../options/bootstrap.min.js" /></script>
-    <script type="text/javascript" src="popup_functions.js" /></script>
-    <script type="text/javascript" src="popup.js" /></script>
+    <script src="../browser-polyfill.min.js"></script>
+    <script src="../global.js"></script>
+    <script src="../jquery-3.3.1.min.js"></script>
```
rugk: Again…

```suggestion
```

keepassxc-browser/popups/popup_login.html
```diff
@@ -1,27 +1,29 @@
...
     <link rel="stylesheet" href="../options/bootstrap.min.css" />
-    <script type="text/javascript" src="../browser-polyfill.min.js"></script>
-    <script type="text/javascript" src="../global.js"></script>
-    <script type="text/javascript" src="../jquery-3.3.1.min.js"></script>
-    <script type="text/javascript" src="../options/bootstrap.min.js" /></script>
-    <script type="text/javascript" src="popup_functions.js"></script>
-    <script type="text/javascript" src="popup_login.js"></script>
+    <script src="../browser-polyfill.min.js"></script>
+    <script src="../global.js"></script>
+    <script src="../jquery-3.3.1.min.js"></script>
```
rugk: Too many files…
```suggestion
```

keepassxc-browser/popups/popup_remember.html
```diff
@@ -1,16 +1,18 @@
...
-    <script type="text/javascript" src="../browser-polyfill.min.js"></script>
-    <script type="text/javascript" src="../global.js"></script>
-    <script type="text/javascript" src="../jquery-3.3.1.min.js"></script>
-    <script type="text/javascript" src="../options/bootstrap.min.js" /></script>
-    <script type="text/javascript" src="popup_functions.js"></script>
-    <script type="text/javascript" src="popup_remember.js"></script>
-    <style type="text/css">
+    <script src="../browser-polyfill.min.js"></script>
+    <script src="../global.js"></script>
+    <script src="../jquery-3.3.1.min.js"></script>
```
rugk: :thinking: 
```suggestion
```

keepassxc-browser/_locales/en/messages.json
varjolintu: The `messages.json` file is not up-to-date. Placeholders are already used. See https://github.com/keepassxreboot/keepassxc-browser/pull/128/files#diff-8ec61d063c3922fa1e1e4d3c60a50be3R129.

keepassxc-browser/global.js
varjolintu: But if you look the parenthesis, this works just like it should. It returns true if it's not Chrome. The logic is upside down.

keepassxc-browser/options/options.html
varjolintu: I decided to leave jQuery for options and popup pages until there's a version of Bootstrap released that uses vanilla JavaScript. (Seems it could be released with version 5).

keepassxc-browser/css/keepassxc-browser.css
varjolintu: That just doesn't work. Even adding the style manually to the JS code the icon doesn't work properly without this kind of path. You can try and verify it yourself.

keepassxc-browser/_locales/en/messages.json
rugk: Actually the line you've linked seems to prepend `???` to `passwordGeneratorBits`, so it actually _does not_ use placeholders… :smile:

keepassxc-browser/css/keepassxc-browser.css
rugk: Ah, it is in a content script.

keepassxc-browser/_locales/en/messages.json
varjolintu: Yes, you are right about that. I need to add the placeholders to the JS side too. With HTML pages that already works.

varjolintu: Fixed :)

keepassxc-browser/_locales/en/messages.json
```diff
@@ -104,7 +104,7 @@
         "description": "Checkbox text below the password generator input field."
     },
     "passwordGeneratorBits": {
-        "message": " Bits",
+        "message": "$1 Bits",
```
rugk: Yeah, exactly that's what I meant. :smiley: 

Just a small note, [MDN recommends](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization#Direct_placeholder_usage):
> This may seem quicker and less complex, but the other way (using "placeholders") is seen as best practice. This is because having the placeholder name (e.g. "url") and example helps you to remember what the placeholder is for — a week after you write your code, you'll probably forget what $1–$8 refer to, but you'll be more likely to know what your placeholder names refer to.

varjolintu:
@rugk I can investigate using the ES6 modules but it's unclear if this comes with any actual benefits. Now the minimum version number supported is 52 so jumping to 60 is a big hop.

And btw, #19 is already using some of this stuff.

rugk:
Yeah, it is just a possibility for refactoring/future work. May not be done, now, however.
And BTW https://github.com/keepassxreboot/keepassxc-browser/pull/19 uses CommonModules of NodeJS, not ES6 (standardized) modules.

whisdol:
I found another little difference in behavior to the jquery-version in the login flow.

Steps to reproduce:
* Open github.com/login
* Click into username field so the auto-fill suggestion is shown
* *Click* on the autofill suggestion to fill the credentials

Expected behavior:
* Autofill credentials are filled in
* The username field is still active
* Pressing <kbd>Enter</kbd> performs the login

Observed behavior:
* Autofill credentials are filled in
* No input field is selected/active
* Pressing <kbd>Enter</kbd> does nothing

varjolintu:
@whisdol Fixed.

guillaume-uH57J9:
Removing jQuery would be nice, keepass-browser is not always very reactive so there's hope it could improve a bit. 
Do you wish me to add a review on the current PR, or do you intend to refactor this?

varjolintu:
@guillaume-uH57J9 I will merge the latest changes from the new PR's to this and rebase everything. I have no intend to refactor this further, only some small changes. Reviews are very welcome.

mehrdadn:
@varjolintu what is the status from my earlier issues? Have you had a chance to look into them?

varjolintu:
@mehrdadn You mean the pages? I replied for those but I cannot find the answer anywhere. Maybe I should edit the README of your repo directly. I know you weren't happy with the [current workflow](https://github.com/keepassxreboot/keepassxc-browser/pull/128#issuecomment-435845108).

mehrdadn:
@varjolintu huh, I'm confused... the last reply I saw from you was [this one above](#issuecomment-436145900) which was in response to [this reply](#issuecomment-435975423) in which I had explained it wasn't just an issue with that particular workflow... I thought you said you were going to look into it since it wasn't working properly. Feel free to edit the README if you'd like!

varjolintu:
@mehrdadn Sorry. For some reason few messages were hidden. I'm gonna re-check the pages and the behaviour when I have the time.

droidmonkey:
@guillaume-uH57J9 excellent comments, thank you

keepassxc-browser/content/autocomplete.js
```diff
@@ -0,0 +1,156 @@
+'use strict'
+
+var kpxcAutocomplete = {};
+kpxcAutocomplete.elements = [];
+
+kpxcAutocomplete.create = function(input, showListInstantly = false) {
+    let _index = -1;
+
+    input.addEventListener('click', function(e) {
```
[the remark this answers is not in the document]
varjolintu: I have, but decided to keep the current behaviour, for now. When selecting the input using `Tab` key, the autocomplete list can still be opened using `Delete` or `Backspace`. I added these shortcuts to make the user experience more friendly for people who like to use the keyboard instead of mouse.

keepassxc-browser/content/autocomplete.js
```diff
@@ -0,0 +1,156 @@
...
+    function showList(input) {
+        closeList();
+        const div = kpxcUI.createElement('div', 'kpxcAutocomplete-items', {'id': 'kpxcAutocomplete-list'});
+
+        // Element position
+        const rect = input.getBoundingClientRect();
+        div.style.top = String((rect.top + document.body.scrollTop) + input.offsetHeight) + 'px';
+        div.style.left = String((rect.left + document.body.scrollLeft)) + 'px';
+        div.style.minWidth = String(input.offsetWidth) + 'px';
+        div.style.zIndex = '2147483646';
```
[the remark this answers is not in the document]
varjolintu: It does. Basically it's INT_MAX - 1. This value is derived from the passifox/chromeipass base. It's just a "maximum" value for CSS z-index, which is an integer. And the interesting thing is that there's no standard for integer maximum value. So basically this is a compromise that seem to work with all browsers.

keepassxc-browser/content/autocomplete.js
```diff
@@ -0,0 +1,156 @@
...
+        } else if (e.key === 'ArrowUp') {
+             --_index;
+            activateItem(item);
+        } else if (e.key === 'Enter') {
+            if (input.value === '') {
+                e.preventDefault();
+            }
+
+            if (_index >= 0 && item && item[_index] !== undefined) {
+                item[_index].click();
```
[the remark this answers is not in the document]
varjolintu: Good point!

keepassxc-browser/content/ui.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+// jQuery style wrapper for querySelector()
+var $ = function(elem)
```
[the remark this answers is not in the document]
varjolintu: The scope is valid only between content scripts. It doesn't collide with jQuery or the page's JavaScript itself. In the current build `$` calls directly jQuery because jQuery is included in the content scripts.

varjolintu:
@guillaume-uH57J9 Thank you for these! I really appreciate it.

varjolintu:
@mehrdadn I updated your repo's README. Everything works for me.

mehrdadn:
@varjolintu Are you sure? :( I just tested it right now and only 2 of those (3rd and 4th one) _sometimes_ work for me if I'm lucky. Otherwise they don't work at all. I'm testing with the key combo, although whenever I tried the menu that didn't work either.

varjolintu:
@mehrdadn I'm sure. I tried many ways to get it to fail. Have you tried to test this with `web-ext` or something? That will use a fresh browser profile and you have to connect it to KeePassXC every time you launch the browser from the tool. I tested Firefox with that, and Chromium with a normal installed version.

mehrdadn:
Nope I've never used `web-ext`, I don't use Firefox. I only tried it on Chrome. I could maybe try on another profile but somehow I doubt it should be an issue with the profile...

varjolintu:
@mehrdadn Just make sure you are using the latest sources.

mehrdadn:
@varjolintu I tried on a different profile, but with similar settings. You're indeed right that it works much better there, though it's not foolproof -- sometimes it takes e.g. 5-10 seconds before it keypresses trigger it to fill in, but it rarely fails outright. On my current profile though for some reason it almost never works. (I'm mostly testing at the 3rd website.) I have no clue what's going on really, but for some reason it feels like there's a race condition of some sort, where a different ordering between two events affects the behavior. I'll take a look at it later if I get the chance, but it might be a while. Sorry if this isn't too helpful otherwise, and thanks for looking into it so far.

varjolintu:
@mehrdadn I sugges you put breakpoints to the following lines:
https://github.com/keepassxreboot/keepassxc-browser/blob/no_jquery/keepassxc-browser/content/keepassxc-browser.js#L1041
https://github.com/keepassxreboot/keepassxc-browser/blob/no_jquery/keepassxc-browser/content/keepassxc-browser.js#L1082

Or add a `console.log()` message in those functions where you can debug the parameters or scoped variables. Maybe that clears things a bit and you can see is the extension even trying to fill the values, or where it errors.

mehrdadn:
@varjolintu I finally had a chance to take a brief look into this. While I'm not 100% sure, it seems to me that, at least on the 3rd website in my list, it's likely a race condition. What happens is a `null` `url` (and also `submitUrl`) is passed to `retrieve_credentials`. (I do see an error in the console log indicating `url` is `null`, and I've verified it while debugging. The only thing I'm not 100% sure about is whether this is actually the sole cause of the problem or not.)

Moreover, it seems the content script makes far too many attempts to retrieve credentials. You would think that single fill-in request should only result in one message sent to the extension's background page, but I easily see at least 3 messages sent, if not more—sometimes `url` is empty for some of them, sometimes it's filled for the first one, sometimes it's filled for a later one. This suggests to me that the code is in fact broken in several places, and has probably had patches applied on top of more patches to re-try the same thing over and over, until hopefully something works. A proper extension should only send one message per fill-in request and get it right the first time without race conditions. So I think, if you cannot reproduce my problem, you should look into fixing the multiple-request issue. Sending 1 request properly is the right thing to do anyway, and chances are it would fix this problem too.

That said, the bug seems to be that `cip.url` is not filled in when the `retrieve_credentials` is sent, so it might be worth looking into why such a race condition might occur. (I can try to fix it too, but I'm not sure when I might get the chance. It could be tomorrow or it could be many weeks from now.) Thanks!

varjolintu:
@mehrdadn Thank you! These findings are absolutely correct. I'll try to clean up the code.

mehrdadn:
:) I suspect this is going to be a very intense refactor, since I seem to recall there were race conditions even in the old chromeIPass code (they just occurred less frequently). I'm not sure what the best way to do this is, but it may make sense to do this in a different branch, and/or to merge all your other branches and then work on this afterward; otherwise, you may end up having to discard or re-do work on other branches just to accommodate the refactor...

varjolintu:
@mehrdadn Yes. That's what I would also do. Old chromeIPass worked totally synchronously so the chance for race conditions was a little smaller.

mehrdadn:
No the problem isn't fixed when I do that... but I'm also very reluctant to test more hacks on top of hacks if I'm being honest!

For the frames, note that the login form may not necessarily be in the main frame. I've seen cases where it isn't in the past. But yeah, you'd only want to send one request per box being filled in, whatever frame it's coming from.

rugk:
Regarding race conditions: Such things are likely best fixed in JS by using proper [Promises](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Promise), so you can build chains and dependencies ("only do X when Y is done" etc.). This may be a major refactoring, however, so I agree that this may rather not be done in this branch/PR.

varjolintu:
@rugk That's what I'm working on at this point. Instead of chains I decided to use `async` and `await` with Promises. It's a big refactor, yes, but the core I have already built detects input fields really fast. Actually so fast that the site kayak.com shows the username and password fields' (using `getBoundingClientRect()`) as full of zeroes, and those are not added to the array of accepted fields. This is probably caused by animations in the web design.

These kind of problems could be solved with polling the input field detection (like the old chromeIPass did) but that would slow down the content script a bit. `MutationObserver` is nice and dynamic but using it creates the issues like described above. Maybe the polling coud be implemented with `MutationObserver` so it could trigger when needed, but would not run in the background all the time.

(Thinking aloud)

mehrdadn:
I think you want MutationObserver. Using it doesn't necessarily imply a race condition. Promises vs. something else is a minor issue.

varjolintu:
I think this PR is closing to a point that it's no longer WIP. Additional reviews are very welcome.

mehrdadn:
@varjolintu Just checking, you intend to fix the race conditions in a different PR rather than this one, right? Because currently this seems to work most of the time (but still not all of the time) on website 3 in my list, but fails to work almost all the time on the other websites.

keepassxc-browser/popups/popup_httpauth.html
```diff
@@ -1,38 +1,48 @@
...
   <body>
     <div class="container">
       <div id="settings" class="settings">
+<<<<<<< HEAD
         <button id="btn-options" class="btn btn-sm btn-success"><span class="glyphicon glyphicon-cog"></span> <span data-i18n="popupSettingsText"/></button>
         <button id="btn-choose-credential-fields" class="btn btn-sm btn-warning"><span class="glyphicon glyphicon-list-alt"></span> <span data-i18n="popupChooseCredentialsText"/></button>
+=======
+        <button id="btn-options" class="btn btn-sm btn-success"><span class="glyphicon glyphicon-cog"></span><span data-i18n="popupSettingsText"></span></button>
+        <button id="btn-choose-credential-fields" class="btn btn-sm btn-warning"><span class="glyphicon glyphicon-list-alt"></span><span data-i18n="popupChooseCredentialsText"></span></button>
+>>>>>>> a815c7f... No jQuery
```
z3ntu: merge conflict markers

keepassxc-browser/popups/popup_httpauth.html
```diff
@@ -1,38 +1,48 @@
...
 
       <div class="credentials">
         <p data-i18n="popupAuthText"></p>
         <div id="login-list" class="list-group"></div>
         <p>
+<<<<<<< HEAD
           <button id="btn-dismiss" class="btn btn-sm btn-danger"><span class="glyphicon glyphicon-remove"></span> <span data-i18n="popupButtonDismissHttpAuth"/></button>
+=======
+          <button id="btn-dismiss" class="btn btn-sm btn-danger"><span class="glyphicon glyphicon-remove"></span><span data-i18n="popupButtonDismissHttpAuth"></span></button>
+>>>>>>> a815c7f... No jQuery
```
z3ntu: more merge conflict markers

z3ntu:
@varjolintu You should also copy the text from https://github.com/keepassxreboot/keepassxc-browser/pull/128/files#r238533726 to the PR description and maybe link to https://github.com/twbs/bootstrap/pull/23586 :)

varjolintu:
@mehrdadn Both of those sites still need the Username-only detection to be enabled from Site Preferences. Have you done that?

mehrdadn:
@varjolintu: Ah you're right, I think that's what the problem with Google was. (Microsoft doesn't -- likely it has a hidden password field for this purpose.) It does work with that option checked. But I also think you should avoid requiring setting such an option. It's a massive pain to have to do this for every site. (And I don't see why it should be necessary either to be honest.)

Also, for some reason every site on my list seems to be working now too.... go figure.

mehrdadn:
(To be clear, I didn't change the settings for the sites on my list. I'm not sure why they're working now.)

rugk:
> It's a massive pain to have to do this for every site. (And I don't see why it should be necessary either to be honest.)

I guess it can have false-positives easily…
However, for a good out-of-the-box experience, you may pre-configure some major services possibly…

mehrdadn:
You can get quite a good accuracy if you search for text like "login" or "signin" etc. (which I believe the codebase did at some point if not currently).

That said, I think one thing that is constantly missed here is that there are multiple flows in the software that can work. "Fill everything upon load" is most definitely _not_ the only way to use the software (and I'd argue it's actively insecure and harmful, but that's another issue), yet it seems every discussion here revolves around the assumption that that's how the software is used. As I've mentioned before, my workflow us to press a keyboard shortcut to fill in boxes -- and only then do I expect boxes to be filled in. At that point, pretty much anything that isn't a password field can be assumed to be a username field. All this friction that gets added in between to cater to other workflows just makes the life of someone like me 10 times harder.

droidmonkey:
It's a one time deal to define a website that has username only field so I doubt it is 10x harder.

mehrdadn:
...

varjolintu:
@rugk Pre-configuring some of the major sites is on the TODO list. That would solve a lot of possible issues.

> But I also think you should avoid requiring setting such an option. It's a massive pain to have to do this for every site. (And I don't see why it should be necessary either to be honest.)

@mehrdadn I've probably explained this before, but there's no standard way to define an input field as an username field. There would be a lot of false positives for example from search fields (those are not standardized either, so many sites don't have an ID or anything that could identify those fields as search), and that would cause retrieve credentials validation (some prefer using without remembering the access) popup dialogs on every tab change or page reload.

The thing I agree with is that if you decide to fill the credentials from the keyboard or the context menu, then those fields should be filled, even if the input field is not regognized as a username field. Or if the page has just single username field. So that thing is something we want definitely to improve. We are not trying to force users to a certain workflow, but to make multiple workflows possible.

mehrdadn:
@varjolintu: Yes I realize there's nothing bulletproof, but what I'm trying to say is that I don't I think a few heuristics (like what I wrote: searching for `login` or something in the `class` or `id` fields, etc.) would drastically decrease the false negative rate without increasing the false positive rate significantly.

mehrdadn:
But in any case, the only workflow I use is the trigger-based one, so that's the one that'd affect me directly.

z3ntu:
image
I don't remember if this message is the same in the latest stable, but this message is not fitting for when keepassxc is just not opened.

varjolintu:
@z3ntu The error messages need a little but of tuning, yes..

rugk:
> I don't remember if this message is the same in the latest stable, but this message is not fitting for when keepassxc is just not opened.

I _do_ remeber it and often see it in stable, so it's likely not about this PR. So, AFAIK, this is off-topic/a different issue.

varjolintu:
Those error messages are going to be modified to the next version. That error message is shown when browser is started without KeePassXC in the background. So, that will be a little more informative in the future :)

keepassxc-browser/_locales/en/messages.json
```diff
@@ -132,7 +136,7 @@
         "description": "Password icon title text."
     },
     "passwordGeneratorError": {
-        "message": "Cannot receive generated password.<br />Is KeePassXC running?",
+        "message": "Cannot receive generated password.\nIs KeePassXC running?",
```
whisdol: This line break is not actually displayed. A way to display it is styling the div with `white-space: pre-wrap`, but I'm not sure if there are more elegant solutions.

whisdol: This message is also shown if KeePassXC is running, but not connected. Should the message be changed to "Is KeePassXC running and connected?"

whisdol:
I found other differences in the behavior of the autocomplete-dialog that I would consider bugs (tested in Firefox on macOS and Windows):

## <kbd>Enter</kbd> does not fill username field
Steps to reproduce:

1. Open github.com/login
2. Click into username field so the auto-fill suggestion is shown and selected (marked green)
3. Press <kbd>Enter</kbd>

Expected behavior:

- Username is filled in
- Password is filled in

Observed behavior:

- Username is blank/not filled in
- Password is filled in


This error only occurs when only one credential is suggested for a particular site. When multiple credentials are suggested, the username is filled in.

## <kbd>Keydown</kbd> on an active input field does not open autocomplete list
Steps to reproduce:

1. Open github.com/login, the username field should be active
2. Press <kbd>Keydown</kbd>

Expected behavior:
- Autocomplete list is shown

Observed behavior:

- Nothing happens

## When autocomplete-credentials are selected, <kbd>Tab</kbd> does not fill them in
Steps to reproduce:

1. Open github.com/login
2. Click into username field so the auto-fill suggestion is shown and selected (marked green)
3. Press <kbd>Tab</kbd>

Expected behavior:

- Selected credentials are filled in

Observed behavior:

- Credentials are not filled in

keepassxc-browser/_locales/en/messages.json
varjolintu: Line changes in translations are tricky. It would be the best to avoid them. And yes, this message should be changed.

varjolintu:
@whisdol
## <kbd>Enter</kbd> does not fill username field
This was a simple bug which I'll push soon.

## <kbd>Keydown</kbd> on an active input field does not open autocomplete list
I wasn't even aware that Keydown opened the input field in the current version. This branch only used <kbd>Backspace</kbd> and <kbd>Delete</kbd> for that (when the input field is empty), but of course `Keydown` can be added.

## When autocomplete-credentials are selected, <kbd>Tab</kbd> does not fill them in
Same as above. Now only <kbd>Enter</kbd> fills the credentials.

whisdol:
Thanks, pressing <kbd>Enter</kbd> now fills the username field as well!

Regarding <kbd>Keydown</kbd> and <kbd>Tab</kbd>, I would argue that opening the autocomplete list/filling the credentials is expected by the user as it is the standard behavior in this and other use cases. Without KeePassXC-Browser, Firefox also opens the autocomplete list on an active input field on <kbd>Keydown</kbd>. When a suggestion is selected, <kbd>Tab</kbd> fills the field with the suggestion and moves to the next field. Auto-Complete suggestions in code editors (I explicitly checked in VS Code, Sublime, so others may vary) are also inserted with <kbd>Tab</kbd>.

(For the sake of completeness of the "standards"-argument: The Firefox autocomplete suggestion is not selected by default as it is in this branch. In that way, this branch is closer to the behavior of code editors, which I like.)

varjolintu:
@whisdol Thanks for the clarification. If code editors behave like that it surely is a valid point.

keepassxc-browser/content/autocomplete.js
```diff
@@ -106,25 +106,39 @@ kpxcAutocomplete.create = function(input, showListInstantly = false) {
...
+                ++_index;
+                activateItem(item);
+            }
         } else if (e.key === 'ArrowUp') {
-             --_index;
+            --_index;
             activateItem(item);
-        } else if (e.key === 'Enter') {
+        } else if (e.key === 'Enter' || e.key === 'Tab') {
             if (input.value === '') {
```
whisdol: Now this prevents <kbd>Tab</kbd>ing to the password field (or anywhere) when the input field is empty. I think we should only prevent the default action from being executed for <kbd>Enter</kbd>, which we could check here.

varjolintu: Ah, true! Thanks for noticing.

keepassxc-browser/_locales/en/messages.json
varjolintu: Resolved with https://github.com/keepassxreboot/keepassxc-browser/pull/128/commits/ec3eb66dbf5859363bd8efddbbc288641b70f92e.

keepassxc-browser/_locales/en/messages.json
```diff
@@ -200,27 +212,31 @@
         "description": "Shown when no username is set in the credentials."
     },
     "credentialsNoLoginsFound": {
-        "message": "Error:\nNo logins found.",
+        "message": "Error No logins found.",
```
z3ntu: Did you intend to remove the colon?

varjolintu: Yes. But `:` should have stayed. And maybe I'll put these to separate lines later, with just general "Error" message as the first one, and the actual error to the second line.

varjolintu: No. And maybe I'll put these to separate lines later, with just general "Error" message as the first one, and the actual error to the second line. I'll fix this soon.

rugk: >  with just general "Error" message as the first one, and the actual error to the second line

This may be hard to localize, in case some languages put the "Error" somewhere else or so…

varjolintu: True.. Maybe it's better to leave just as it is. I'll just add the colon back. (It's frustrating to have these tiny one character errors)

varjolintu:
Replaced the ugly polling mechanism (`setInterval()`) in password generator with asynchronous IntersectionObserver with https://github.com/keepassxreboot/keepassxc-browser/pull/128/commits/d9928d1c324361acf444b0af1390477bd68c2693.

whisdol:
Another tiny UX bug: When clicking on the autocomplete suggestion, the username field does not stay active.
Steps to reproduce:

1. Open github.com/login
2. Click into username field so the auto-fill suggestion is shown and selected (marked green)
3. Click on the autocomplete suggestion
4. Press <kbd>Enter</kbd>

Expected behavior:

After 3., the username field is active. Pressing <kbd>Enter</kbd> performs the form submission, i.e. login.

Observed behavior:

After 3., no input field is active. Pressing <kbd>Enter</kbd> does nothing.

varjolintu:
Finally it seems that we are going to merge this for the next release. This PR also includes #355 which is still not quite ready. I'll see if I can improve it in this PR so we can delete #355. It would be also nice to include all translations in the same release with this one.

So if anyone finds any errors or bugs, let me know :)

varjolintu:
Automatic reconnect is implemented. If the option is disabled and user is triggering the popup, it will try reconnect automatically. When the option is enabled, of course everything happens in the background.

droidmonkey:
In Settings -> Site Preferences, typing a URL in the input box and pressing +Add doesn't work.

z3ntu:
Just wanting to say that it feels really solid now. The auto-reconnect is contributing a lot to that I think, as the instances where the field doesn't get filled in, I have to press the icon, press reconnect, press redetect credential fields and then fill it in, do not exist anymore :+1:

varjolintu:
Found one bug: for example with Protonmail, the autocomplete menu doesn't activate after logout. Plus the autocomplete menu didn't work properly with the mailbox password field.

Gonna fix them soon.

varjolintu:
All is ready for merge.

z3ntu:
I didn't test if this is because of this branch or it's a problem in general but the "Also fill in next password field" option doesn't work on https://bugs.lttng.org/account/register

(after writing this I found a few issues I'll try to fix)

z3ntu:
So this diff fixes that
```diff
diff --git a/keepassxc-browser/content/pwgen.js b/keepassxc-browser/content/pwgen.js
index cac1ace..5bc3607 100644
--- a/keepassxc-browser/content/pwgen.js
+++ b/keepassxc-browser/content/pwgen.js
@@ -78,7 +78,7 @@ kpxcPassword.initField = function(field, inputs, pos) {
         }
     }
 
-    field.setAttribute('kpxc-genpw-next-field-exists', found);
+    field.setAttribute('kpxc-pwgen-next-field-exists', found);
 };
 
 kpxcPassword.createIcon = function(field) {
@@ -229,7 +229,10 @@ kpxcPassword.showDialog = function(field, icon) {
         const checkbox = $('.kpxc-pwgen-checkbox');
         if (checkbox) {
             checkbox.setAttribute('checked', fieldExists);
-            checkbox.setAttribute('disabled', !fieldExists);
+            if(fieldExists)
+                checkbox.removeAttribute('disabled')
+            else
+                checkbox.setAttribute('disabled', '');
         }
     }
 };
@@ -283,7 +286,7 @@ kpxcPassword.fill = function(e) {
 
         field.value = password.value;
         if ($('.kpxc-pwgen-checkbox').checked) {
-            if (field.getAttribute('kpxc-pwgen-checkbox')) {
+            if (field.getAttribute('kpxc-pwgen-next-field-id')) {
                 const nextFieldId = field.getAttribute('kpxc-pwgen-next-field-id');
                 const nextField = $('input[data-kpxc-id=\'' + nextFieldId + '\']');
                 if (nextField) {

```
First hunk fixes a typo that made the right attribute not being set
Second hunk fixes the "disabled" attribute on the "fill next password", as the fact that the disabled attribute is present (regardless of the values) disables the checkbox
And the third hunk fixes the if to actually fill in the next password field

varjolintu:
@z3ntu Thanks! I totally missed that.

I would also like to thank everyone who tested this PR, found bugs and committed some fixes. I couldn't been able to do this alone. Really appreciate it.

z3ntu:
One more thing: the `<label class="kpxc-pwgen-checkbox-label" for="kpxc-pwgen-checkbox">Also fill in the next password-field</label>` in the password generator isn't really "connected" with the checkbox itself because the checkbox doesn't have an `id` attribute but just has the `kpxc-pwgen-checkbox` as class and as such, you can't click the label to toggle the checkbox

varjolintu:
That is also fixed.

z3ntu:
Thanks for the hard work @varjolintu !

guillaume-uH57J9:
As it got published I was able to test the new version with these change. It fixed the keepass-browser's main annoyance : failure to connect/fill forms on the first try. 
Thanks for the effort on this, the extension now works as well as earlier extension based on Firefox' password manager API.

mehrdadn:
@varjolintu Thanks for the effort! I just started testing it out.

For some bizarre reason, with my particular combination of settings, I seem to actually get an automatic form submission (?!?!) when I fill in the password. And even worse, it's a broken submission, so the browser just reloads the page entirely when I try to fill something in credentials, and if I try to refresh, warns me that it would resubmit the form. (For example, this happens on Bank of America.)

I'm using 3fef9eb; my settings are:

- [ ] Always ask where to save new credentials
- [ ] Activate password generator.
- [ ] Automatically retrieve credentials.
- [ ] Automatically reconnect to KeePassXC.
- [x] **Automatically fill in single-credential entries.**
- [x] Activate autocomplete for username fields.
- [x] Show notifications.
- [ ] Show a notification when new credentials can be saved to the database.
- [x] Save domain only.

I think the issue is with **Automatically fill in single-credential entries**, because without it, it seems to work much better. Though even without that, it's still flaky and I still have issues autodetecting and filling in certain sites.

droidmonkey:
This PR is merged. Please open a new issue with your details.

Issue #94, opened 2018-03-23
Username+password field detection not working on personalcapital.com


nbartos:
The login page is here:  https://home.personalcapital.com/page/login/goHome.  This is a multi-step form with username first then password second.  When using the chromeIPass extension, I can use the right click "fill user+pass" for the username field, and right click "fill pass only" for the password field.

## Expected Behavior
There should be the ability to fill form data in one form or another, as there is with chromeIPass.

## Current Behavior
I have tried doing "choose own credentials for this page" and also right clicking with "fill user+pass" and "fill pass only", but I cannot data to contently fill in the username or password fields.

## Possible Solution
Allow the right click "fill user+pass" and "fill pass only" options to work as they do in chromeIPass.

## Steps to Reproduce (for bugs)
1.  Create credential entry in keepassxc for home.personalcapital.com.
2.  Go to https://home.personalcapital.com/page/login/goHome
3.  Right click on username field and go to keepassxc -> fill user+pass
4.  Observe that the credentials are not filled in the form

KeePassXC - 2.3.1
keepassxc-browser - 1.0.1
Operating system: Linux
Browser: Chrome
Proxy used: YES

varjolintu:
Cannot reproduce this. Filling the user from the context menu works without any problems. Also works from the popup.

varjolintu:
I ran some new issues with this page. Trying to make some necessary fixes.

nbartos:
I have found that the problem is intermittent.  Sometimes when I do a reload of the page, it works.  Often I have the problem on paypal.com when completing a transaction from another site.  I also hit it on many other sites, it's really getting annoying.  Whenever I use the old chromeIPass, the right click "fill user+pass" and "fill pass only" work very reliably.

varjolintu:
I think I may have broken something if chromeIPass still works. However, I found a reason for it. Usually credentials are identified with a username and password combination. And there's no situation handled (properly) when only username field is visible. Earlier I added a support for just identifying the password field.

And the good news is that I already have a fix under development, and for now it seems to work every time. I still need to verify it with multiple sites though.

nbartos:
That makes sense, it's possible that I've only been hitting this on sites that split out the username and password into separate pages (as that's getting a lot more common these days).  If you have a test build of the extension I can install, I'd be happy to test.

varjolintu:
[...]

Issue #36, opened 2018-03-01
KeePassXC Browser not working on sites, where login is split into multiple sites


t-aus-m:
KeePassXC Browser (aswell as KeePasshttp before) is not working on sites, where login is split into multiple sites. Most known example for this would be Google.

## Expected Behavior
The browser plugin detects credential fields correctly

## Current Behavior
On sites like Google, where the login is split into zwo seperate sites, the plugin does not work properly. You have to type your email adress manually. Selecting the password works though.

## Possible Solution
Eighter hardcode a behavior for the few sites that use multiple sites for login into the plugin, or maybe add a way to make custom credential fields work with those sites.

## General Info
KeePassXC fork - 3.2.0
keepassxc-browser - 1.0.0

Operating system: Win 10, Ubuntu 17.10 _(does not matter)_
Browser: 59.0b13 (64-Bit) _(does not matter)_
Proxy used: NO

varjolintu:
You are correct. Google page really has some difficulties even when I try to choose my own credential fields. If I use `https://accounts.google.com/signin/v2/sl/pwd` as a URL for the entry the popup detects the credentials but nothing happens when choosing them. Autocomplete doesn't work. So there's a room from improvement. At some point this worked though.

pfoo:
Isn't this related to the missing multiple-url-entries feature of keepassxc ?

varjolintu:
Actually no. I can retrieve the same credentials from both login and password page. Still I have troubles with the input fields. I'm not sure if Google has added some nasty JavaScript to disable this kind of behaviour from content scripts etc. Separate login/password pages like http://addons.mozilla.org/ works normally.

seekM:
This issue still exists for me: The username field is not detected for me when opening https://accounts.google.com. However, when I enter my username manually and press enter I am forwarded to a new webpage. The passwort field on this webpage (still https://accounts.google.com) is detected. I am using KeePassXC-Browser 1.5.1 / Firefox 69.0.2 on Windows / KeePassXC 2.4.3.

varjolintu:
@seekM Have you enabled the Username-only support for the site through Site Preferences? Correct URL for that page is `https://accounts.google.com/*`.

seekM:
I did not! Thank you, it works now.

varjolintu:
@seekM Glad it works! The popup should also inform you if it's possible to add the current site automatically to Site Preferences and enable the Username-only support.

seekM:
[...]

PR #132, opened 2018-04-30
Improve detection of username fields

varjolintu:
This PR improves detection of username fields on pages there's only username field visible. Also, autocomplete will be enabled for these inputs.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/94
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/36

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -489,7 +489,7 @@ cipForm.init = function(form, credentialFields) {
     // TODO: could be called multiple times --> update credentialFields
 
     // not already initialized && password-field is not null
-    if (!form.data('cipForm-initialized') && credentialFields.password) {
+    if (!form.data('cipForm-initialized') && (credentialFields.password || credentialFields.username)) {
         form.data('cipForm-initialized', true);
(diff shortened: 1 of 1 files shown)


nbartos:
In general, this is a big improvement for filling in username fields.  However, when compared to the released keepassxc-browser 1.0.1, I cannot get it to fill in the password by any means on accounts.google.com:

I have tried clicking on the password input form, where there is no drop down.

I have tried clicking the 'redetect credential fields', which doesn't seem to do anything.

I have tried right clicking on the password field and selecting 'fill pass only'.  This makes a popup show up that says 'more than one login was found'.  But after this, going to the keepassxc-browser icon or performing any other interaction as mentioned above still does not work.

I have tried all of these things in many different orders, but still no dice.

varjolintu:
@nbartos Thank you for noticing that. I'll try to improve the behavior also with the password input fields.

seekM:
The username field is not detected for me when opening https://accounts.google.com. However, when I enter my username manually and press enter I am forwarded to a new webpage. The passwort field on this webpage (still https://accounts.google.com) is detected. I am using KeePassXC-Browser 1.5.1 / Firefox 69.0.2 on Windows.

PR #133, opened 2018-05-01
Add browser-polyfill source map

allanbreyes:
Fixes the console warning:
```
Source map error: TypeError: NetworkError when attempting to fetch resource.
Resource URL: moz-extension://.../browser-polyfill.min.js
Source Map URL: browser-polyfill.min.js.map
```

I'm only observing this on localhost. Alternatively, we can just remove the source map URL, but I figure it's better to add it then not.

Fetched from: https://unpkg.com/webextension-polyfill@0.1.0/dist/browser-polyfill.min.js.map

MD5: b3a5040f57aaaf760ee964796519e393


varjolintu:
Source maps are used only for debugging purposes. Adding that to the release is not going to happen.

PR #134, opened 2018-05-01
Remove source map URL

allanbreyes:
Re-do of #133. Removes the [`sourceMappingUrl`](https://developer.mozilla.org/en-US/docs/Tools/Debugger/How_to/Use_a_source_map) comment in the minified `browser-polyfill.min.js`. This is a QoL improvement that fixes a console warning on local development (i.e. localhost). Per @varjolintu, the source map won't make it into a release. However, the polyfill seems to be pretty short, so perhaps a source map isn't needed for the KeepassXC Browser devs.


allanbreyes:
Note: for anyone else who's wondering how to suppress these warning messages in local development, you can [edit (any) Firefox extension](https://stackoverflow.com/a/20126970/3557448). Note, however,

PR #157, opened 2018-05-13
Stop querying changed text nodes for children elements

BendingBender:
Currently, this plugin produces quite a few exceptions in the browser's console when the observed list of changed nodes contains text nodes:

```
keepassxc-browser.js:1147 Uncaught TypeError: mut.target.querySelectorAll is not a function
    at MutationObserver.<anonymous> (keepassxc-browser.js:1147)
(anonymous) @ keepassxc-browser.js:1147
```

This pull request fixes the issue by filtering out texts nodes from the list of mutated nodes.


phoerious:
Thanks!

PR #162, opened 2018-05-16
Fix filling TOTP with multiple pages

### LINKED ISSUE #161: "Fill TOTP" does nothing on Firefox 61b
Since the latest release of the extension (as of May 2018) I think there's a regression on the Fill OTP on Firefox, the code is not generated. At this time only noticed when logging on Github.
It works on Chromium and it worked on Firefox.

## Expected Behavior
The OTP generation should generate a token like it did before

## Current Behavior
During the OTP auth pahse, the OTP generation fails and the keepassxc icon shows a (non-selectable) confirmation of the credentials (see attached image)

- Sign in on gihub
- keepassxc correctly autofills user+pass
- next step ORP, keepassxc does not produce the TOTP. Shows a  a (non-selectable) confirmation of the credentials.

## Debug info
KeePassXC-Browser Version: 1.1.3
KeePassXC Version: 2.3.3 
Operating system: Ubuntu 18.04, x86_64
Browser: Firefox only, Chromium is ok
Proxy used: YES

varjolintu:
If login screen has a separate page for TOTP filling the value will be preserved outside of content script.

Fixes #161.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -22,12 +22,16 @@ browser.runtime.onMessage.addListener(function(req, sender, callback) {
                 if (cip.u) {
                     cip.setValueWithChange(cip.u, cip.credentials[req.id].login);
                     combination = cipFields.getCombination('username', cip.u);
-                    _loginId = req.id;
+                    browser.runtime.sendMessage({
+                        action: 'page_set_login_id', args: [req.id]
+                    });
                     cip.u.focus();
                 }
                 if (cip.p) {
                     cip.setValueWithChange(cip.p, cip.credentials[req.id].password);
-                    _loginId = req.id;
+                    browser.runtime.sendMessage({
+                        action: 'page_set_login_id', args: [req.id]
+                    });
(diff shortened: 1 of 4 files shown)


keepassxc-browser/keepassxc-browser.js
```diff
@@ -1515,23 +1518,26 @@ cip.fillInFromActiveElementTOTPOnly = function(suppressWarnings) {
...
-                const val = s["KPH: {TOTP}"];
-                if (val) {
-                    cip.setValue($sf, val);
+
+    browser.runtime.sendMessage({
+        action: 'page_get_login_id'
+    }).then((pos) => {
+        if (pos >= 0 && cip.credentials[pos]) {
+            // Check the value from stringFields (to be removed)
+            const $sf = _fs(fieldId);
```
phoerious: Why the $?  _sf() is also not very readable.

varjolintu: It's a common practice to use `$` when a variable is something jQuery uses directly. I've already removed these in the WIP No jQuery PR. I'll change this to a more readable form.

PR #166, opened 2018-05-20
MutationObserver optimizations

varjolintu:
MutationObserver optimizations introduce a new `cipObserverHelper` object that optimizes the detection of changed/added/removed elements.

Attribute changes that are not modifying `display` or `z-index`are ignored. New elements are stored to `_observerIds` array that prevents credential field initialization to fields that are already detected.

Maximum input fields on one page is limited to 100. After that KeePassXC-Browser doesn't do any detection.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/177.
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/171.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -6,6 +6,7 @@ _called.retrieveCredentials = false;
 _called.clearLogins = false;
 _called.manualFillRequested = 'none';
 let _loginId = -1;
+const _maximumInputs = 100;
 
 // Count of detected form fields on the page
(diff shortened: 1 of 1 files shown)


varjolintu:
Found a few bugs while testing. WIP.

varjolintu:
Added a separate handling for `attributes` and `childList`. Attributes can be modified hundreds of times in a page after a single mouse click that opens a dialog. Now attributes that don't modify CSS `display` are ignored.

Also, saving identified mutations are saved with `classList` or `id` instead of just `id` which is not always set.

Issue #186, opened 2018-06-10
KeepassXC browser fills search fields since 1.1.4


olifre:
## Current Behavior
KeepassXC fills the following input field with my username, since 1.1.4:
```
<input name="search" id="search" value="enable_puppetlabs_pc1_repo" placeholder="Filter ..." class="autocomplete-input form-control ui-autocomplete-input" data-url="/hosts/auto_complete_search" autocomplete="off" data-cip-id="search" type="text">
```
It seems the code only looks for `data-cip-id`. However, that is set to `search` here, so it's for sure not a username field. 

## Possible Solution
Add logic to check the actual value of `data-cip-id`. 

## Steps to Reproduce (for bugs)
This happens in the Foreman user interface (i.e. if installing the software from https://www.theforeman.org/ ), but likely also at many other places. 

## Debug info
KeePassXC - 2.3.3
keepassxc-browser - 1.1.4
Operating system: Linux
Browser: Firefox
Proxy used: YES

varjolintu:
Clearly we need to ignore input fields if its id or name contains "search", when the type is still "text". I guess there's no other way to identify these. In the meantime you can choose the credential fields manually for that site.

theonlydoo:
same here, super annoying

varjolintu:
This is happening because of the change that allowed only username field to be detected for KeePassXC-Browser. It allowed many pages to work that remained undetected in previous versions.

A fix has been done, and it's waiting for merge.

theonlydoo:
thanks!

droidmonkey:
This still happens in Gmail and Inbox. We really need to move to an opt-in setting for single field fills, or explicit site registration. Heuristics DO NOT WORK in this case.

PR #189, opened 2018-06-11
Ignore search fields

varjolintu:
Checks input fields for any `search` values from all attributes. This is needed because only a few pages actually use the standard `type="search"` for input fields.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/186.
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/190.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -819,7 +819,22 @@ cipFields.prepareId = function(id) {
     return id.replace(/[:#.,\[\]\(\)' "]/g, function(m) { return '\\'+m; });
 };
 
-// Check aria-hidden attribute by looping the parent elements of input field
+/**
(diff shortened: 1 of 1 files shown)


keepassxc-browser/keepassxc-browser.js
```diff
@@ -853,6 +853,22 @@ cipFields.getOverflowHidden = function(field) {
     return false;
 };
 
+cipFields.isSearchField = function(target) {
+    const closestForm = target.closest('form');
+    const attributes = target.attributes;
+
+    for (const attr of attributes) {
+        const formAction = closestForm ? closestForm.getAttribute('action') : null;
+        const role = closestForm ? closestForm.getAttribute('role') : null;
```
phoerious: The element having this attribute is not required to be a form

Issue #193, opened 2018-06-12
Random text fields get auto-filled with username


jlindgren90:
## Expected Behavior
Only username fields should get auto-filled with username.
In a previous version (1.1.3?) it seemed that the plugin would only look for a username field if there was also a password field on the same page.  This wasn't perfect, but it's better than what v1.1.5 is doing.

## Current Behavior
Random text fields get auto-filled with username.

Some examples:
The "Estimated time" field on a Redmine bug edit got auto-filled: https://i.imgur.com/cldlg4i.png
The title field of this bug has gotten auto-filled with my GitHub username THREE TIMES while I've been typing this report (overwriting the title I entered!): https://i.imgur.com/k1BVu5m.png

v1.1.4 was even worse (for example, auto-filling the main search field on google.com) but it seems that the general problem wasn't fixed, only specific fields (like search fields) blacklisted.

## Possible Solution
Revert to the behavior of v1.1.3

## Steps to Reproduce (for bugs)
1. Click the "New Issue" button in this GitHub project with GitHub credentials saved and auto-fill enabled

## Debug info
KeePassXC - 2.3.3 (irrelevant)
keepassxc-browser - 1.1.5
Operating system: Linux
Browser: Firefox
Proxy used: I don't think so? How would I tell?

tohn:
I also have this issue. Whenever this happens, I have to kill the browser tab so I can use the browser again ...

Issue #192, opened 2018-06-11
v1.1.4 does not work (UI anymore) at all


rugk:
## Expected Behavior

Work as usual…

## Current Behavior

The UI state is complexly broken.
The **only way** inserting usernames/password works is when by right-clicking on the input field and choosing "add username/password" or similar.

It does not:
* show the key next to the password field
* show any passwords in the popup (always only lets me click "recheck credential fields")
* show any dropdown when clicking into username/pw fields

This happens on all sites.

I hope it is related, but it also does not show me the actual KeePassXC-browser version:

![grafik](https://user-images.githubusercontent.com/11966684/41251627-c49c8f16-6dba-11e8-9078-581b1f945474.png)

Some logs here:
<details>

``` 
TypeError: cip.settings.ignoredSites is not iterable keepassxc-browser.js:1350:1
20:55:51.561 Content Security Policy: 'x-frame-options' wird wegen 'frame-ancestors'-Direktive ignoriert. (unbekannt)
20:55:51.652 Content Security Policy: Die Einstellungen der Seite haben das Laden einer Ressource auf self blockiert ("script-src"). Source: (function (NAVIGATOR, OBJECT) {

    OBJ....  keepassxc-browser:1
20:55:51.652 Content Security Policy: Die Einstellungen der Seite haben das Laden einer Ressource auf self blockiert ("script-src"). Source: (function (ERROR) {

    const V8_STACK_....  keepassxc-browser:1
20:55:52.069 Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist.  webrequest.js:118
	onBeforeRequest moz-extension://48e8a53b-891a-424d-881d-0c47a6d247cc/js/webrequest.js:118:3
20:55:52.114 TypeError: cip.settings.ignoredSites is not iterable keepassxc-browser.js:1350:1
20:55:53.554 Error: WebExtension context not found!  ExtensionParent.jsm:861:13
20:55:53.591 Content Security Policy: Die Einstellungen der Seite haben das Laden einer Ressource auf self blockiert ("script-src"). Source: (function (NAVIGATOR, OBJECT) {

    OBJ....  releases:1
20:55:53.591 Content Security Policy: Die Einstellungen der Seite haben das Laden einer Ressource auf self blockiert ("script-src"). Source: (function (ERROR) {

    const V8_STACK_....  releases:1
[...]

Issue #187, opened 2018-06-10
Credential fields not auto-filling


tfvlrue:
## Expected Behavior
I have the "Automatically fill-in single credentials entry." setting enabled.
In 1.1.2, credential fields for all websites would automatically fill in. This is the expected behavior.
In 1.1.3, occasionally fields would fail to be detected. I worked around this by clicking the "Redetect Credential Fields" button, and it would fill in the fields.

## Current Behavior
As of 1.1.4, many websites do not auto-fill, and clicking the "Redetect" button fails to do anything.
Fields simply are not auto-filled and there doesn't seem to be a way to get them to fill. The username picker is also not appearing when clicking in the username box.
Additionally, sometimes the page becomes unresponsive and Firefox shows a yellow bar explaining the keepassxc-browser add-on is slowing down the page. (Clicking "Stop" makes the page interactive again.)

## Possible Solution
This is a regression that has started in 1.1.3 and got significantly worse in 1.1.4.

## Steps to Reproduce (for bugs)
Here are a few examples of sites which worked in 1.1.3 but no longer work in 1.1.4:

1. https://www.ally.com/
2. https://nb.fidelity.com/public/nb/default/home
3. https://verified.capitalone.com/sic-ui/#/esignin

Let me know if there is any more info I can provide.

Edit: One more piece of info, not sure if useful. As soon as I reverted to version 1.1.3, all of the credential fields in the tabs I had open immediately got filled in, without refreshing the page or anything.

## Debug info
KeePassXC - 2.39
keepassxc-browser - 1.1.4
Operating system: Win
Browser: Firefox 60.0.2 (64-bit)
Proxy used: NO

varjolintu:
Indeed the autofill is doing something strange. I'm on it.

varjolintu:
This is related to the new MutationObserver changes. In the meantime a workaround is to disable auto-fill.

cryptomilk:
I do not use 'auto-fill' of credentials, I have just ticked auto-retrieve. However when clicking on the username field normally you can select the credentials but this does not work. It looks like that the credential field detection doesn't work at all.

varjolintu:
@cryptomilk That bug is a different issue and has been fixed. A release is soon.

tfvlrue:
[...]

PR #194, opened 2018-06-12
Disable single inputs

varjolintu:
Disables single input fields for now. An opt-in feature will be implemented later.
This also fixes one bug from `isVisible()` that allowed input fields to be detected which contained a CSS class with hidden visibility. Now the check is being done against `getComputedStyle()`.

Fixes: https://github.com/keepassxreboot/keepassxc-browser/issues/193
Fixes: https://github.com/keepassxreboot/keepassxc-browser/issues/192
Fixes: https://github.com/keepassxreboot/keepassxc-browser/issues/187

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -6,6 +6,7 @@ _called.retrieveCredentials = false;
 _called.clearLogins = false;
 _called.manualFillRequested = 'none';
 let _loginId = -1;
+let _singleInputEnabledForPage = false;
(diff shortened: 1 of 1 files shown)


rugk:
:cry: I use single-password fields quite often… Could not you just use the behaviour as it was before the update?

But I don't want to annoy you here. It's likely not an easy thing to detect these…

phoerious:
It will be replaced by a white-list for websites using single-field username fields. For now, we disabled the feature to prevent unnecessary popups.

PR #208, opened 2018-06-17
Replace Ignored sites with Site preferences

varjolintu:
Ignored sites settings tab is replaces with Site preferences tab. It allows to modify site specific settings. `Ignore form submits` is enabled by default. This PR also adds a feature for whitelisting pages with only a single username field visible, e.g. `login.live.com` or `accounts.google.com`.

Here's a preview [screenshot](https://i.imgur.com/I5F30BO.png).

### DIFF
--- a/keepassxc-browser/background/browserAction.js
+++ b/keepassxc-browser/background/browserAction.js
@@ -205,10 +205,10 @@ browserAction.setRememberPopup = function(tabId, username, password, url, userna
     browser.storage.local.get({'settings': {}}).then(function(item) {
         const settings = item.settings;
 
-        // Don't show anything if the site is in the ignore list
-        if (settings.ignoredSites !== undefined) {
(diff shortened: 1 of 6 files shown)


droidmonkey:
What does "ignore form submits" mean?

varjolintu:
I was thinking about drop down selection too, at least for the ignore setting. "Ignore form submits" should then just be "Ignore site" with options "Ignore credentials detection" and "Full ignore".

varjolintu:
Checkboxes updated to drop down selection. Screenshot updated.

droidmonkey:
That looks much better, I'll suggest wording changes later, but the layout is great

varjolintu:
Added a feature to the manual add: a slash is added to the end of URL's if it is missing. This ensures the compatibility with Match Patterns.

phoerious:
Needs conflict resolution.

Issue #209, opened 2018-06-18
Error in chrome console


noone-silent:
Error in chrome console:
```
Uncaught TypeError: target.className.includes is not a function
    at Object.cipObserverHelper.handleObserverAdd (keepassxc-browser.js:1235)
    at MutationObserver.<anonymous> (keepassxc-browser.js:1304)
```

## Expected Behavior
No impact detected

## Current Behavior
No impact detected

## Possible Solution
Wrap target.className in String on line 1235: 
```
if (target.className && (String(target.className).includes('kpxc') || String(target.className).includes('ui-helper'))) {
```

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Debug info

KeePassXC - 2.3.3
keepassxc-browser - 1.1.7
Operating system: Linux
Browser: Chrome
Proxy used: Which proxy? HTTP? Then no

varjolintu:
Which site produces this error? I wonder if that's element at all.

noone-silent:
As an example, I get this error on this page:

https://github.com/keepassxreboot/keepassxc-browser/graphs/code-frequency

I think it has something to do with canvas or svgs.

danielkaneider:
@phoerious This bug still happens with the 1.2.0 release. Please test it for example on this site: http://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/annotations/basic/ 
IMHO instead of checking for some svg elements, we could check `typeof target.className === 'string'``  This should work independently whether the target element is of type svg-text, or svg clippath, or something else

varjolintu:
Check against that seems to work. I'll make the fix.

PR #212, opened 2018-06-20
Ignore SVG elements

varjolintu:
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/209.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -1231,8 +1231,22 @@ cipObserverHelper.getId = function(target) {
     return target.classList.length === 0 ? target.id : target.classList;
 };
 
-cipObserverHelper.handleObserverAdd = function(target) {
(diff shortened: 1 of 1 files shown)


keepassxc-browser/keepassxc-browser.js
```diff
@@ -1231,8 +1231,22 @@ cipObserverHelper.getId = function(target) {
     return target.classList.length === 0 ? target.id : target.classList;
 };
 
-cipObserverHelper.handleObserverAdd = function(target) {
+cipObserverHelper.ignoredElement = function(target) {
+     // Ignore SVG elements
+    if (target.className && (target.nodeName !== 'svg' || target.parentNode.nodeName !== 'svg')) {
```
steven-hadfield: The `target.className` seems unnecessary for this conditional

varjolintu: The whole conditional needed to be fixed. Thanks for noticing this.

PR #242, opened 2018-07-30
Hide the connected key(s) partially in the settings tab

varjolintu:
Hides the connected key(s) partially in the Connected databases settings tab.
The first eight characters are shown. This enables the comparison with the key saved in KeePassXC without revealing it totally.

### DIFF
--- a/keepassxc-browser/options/options.js
+++ b/keepassxc-browser/options/options.js
@@ -156,6 +156,10 @@ options.showKeePassXCVersions = function(response) {
     $('#tab-general-settings button.checkUpdateKeePassXC:first').attr('disabled', false);
 };
 
+options.getPartiallyHiddenKey = function(key) {
+    return !key ? 'Error' : (key.substr(0, 8) + '*'.repeat(10));
+};
+
 options.initConnectedDatabases = function() {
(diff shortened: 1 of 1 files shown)


keepassxc-browser/options/options.js
```diff
@@ -156,6 +156,16 @@ options.showKeePassXCVersions = function(response) {
     $('#tab-general-settings button.checkUpdateKeePassXC:first').attr('disabled', false);
 };
 
+options.getPartiallyHiddenKey = function(key) {
+    if (!key) {
+        return 'Error';
+    }
+
+    let hiddenKey = key.substr(0, 8);
+    hiddenKey += '*'.repeat(key.length - 8);
```
droidmonkey: Recommend just showing a fixed number of asterisks (10?)

varjolintu: I agree. No need to count the length.

phoerious: I'd prefer five. Ten seems excessive.

varjolintu: What matters is that user understands this is clearly a partially hidden key, not some black magic.

phoerious: I don't think that matters much. Users also understand when their credit card numbers are partially hidden and you will rarely see more than four asterisks for that.

PR #245, opened 2018-07-31
Replace tabs on 4 spaces in manifest

Diadlo:


varjolintu:
I would actually suggest to convert all tabs to spaces.

Diadlo:
@varjolintu Done

PR #246, opened 2018-07-31
Fix ignored elements function

varjolintu:
Ignores any elements that don't have a className as a string. This includes SVG elements.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/209.
Fixes https://github.com/keepassxreboot/keepassxc/issues/2207.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -1257,16 +1257,14 @@ cipObserverHelper.getId = function(target) {
 };
 
 cipObserverHelper.ignoredElement = function(target) {
-    // Ignore SVG elements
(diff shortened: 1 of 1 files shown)


victornoel:
@varjolintu @phoerious  any news on that? it is really annoying when developing web apps using svg, the console is full of errors :)

droidmonkey:
@varjolintu can you explain further why this blocks all SVG elements? Seems like a very wide net.

varjolintu:
@droidmonkey Sure. The whole idea of the function is to ignore all objects that cannot have input fields. While SVGElement is derived from Element, it doesn't have a className property. This is the simplest way to check if the element can actually have any content we can use for further checking.

Issue #211, opened 2018-06-20
Save Credentials stopped working


cogliostro:
## Expected Behavior
Enter username and password into credential fields. Right click field -> KeepassXC Browser -> Save credentials. Extension icon starts blinking red. Shows options add or update when clicked.

## Current Behavior
Enter username and password into credential fields. Right click fields -> KeepassXC Browser -> Save credentials. Nothing happens.

## Debug info

KeePassXC - 2.3.3
keepassxc-browser - 1.1.7
Operating system: Linux (Fedora 27)
Browser: Chrome/Firefox/Vivaldi/Chromium
Proxy used: YES

varjolintu:
Cannot reproduce this. Is your database open when you do this?

oidualc:
I can confirm this behaviour on Firefox Nightly 62.0a1 for a while now, both on Windows and Linux, the database is open.
Instead it's working fine for me on Firefox Stable 60.0.2 and all versions of Chrome, so not sure what the causes of this bug might be.

varjolintu:
Here Firefox and Chromium stable works too. It might be an issue in the Firefox Nightly. Often there are internal API changes that might break something.

cogliostro:
I have this issue in  vivaldi, chromium and firefox stable. Not sure how to debug this. Any ideas?

varjolintu:
@cogliostro This works with Chrome: go to the page you have the problem, right click mouse and select Inspect, and select debugging tab. keepassxc-browser.js should open. In cipForm.onSubmit() function is the one that should trigger the popup in the last line.

You should set a breakpoint to https://github.com/keepassxreboot/keepassxc-browser/blob/develop/keepassxc-browser/keepassxc-browser.js#L548.

Please note that AJAX form submits cannot be properly detected, so if it doesn't work for a certain site, that might be the case.

motherload:
Can confirm this behaviour on firefox-nightly 62.0a1.20180624 , keepassxc plugin version 1.1.7, arch linux, the database is open.

I can see the blinking icon, but it doesn't do anything, no matter how many times I click on it. I never see the option to save credentials in the database.

varjolintu:
@motherload No need to write to two different issues. The workaround did not work for you? Can you debug the extension?

rdd2:
I've got the same behaviour. Context menu 'Save credentials' does nothing and there is also no prompt on updated credentials being sent.

Ubuntu 18.04 with KDE
[...]

Issue #253, opened 2018-08-05
Icon blinks red but no popup to save or update credentials on Firefox


samueltardieu:
## Current Behavior

On Firefox62.0b14/ArchLinux (but I noticed that on earlier versions too), creating/changing a password or explicitly asking to save credentials will make the KeepassXC-browser icon blink red, as it should, but no popup will be presented to save the credentials. Only the regular "Settings / choose credentials / close db" choices are available.

I've tried disabling all my extensions, as well as adding a dummy site to the ignore list as was suggested in an earlier issue for password save problems to no avail. Clicking the icon once, twice or more during the blinking delay does not change the situation.

## Expected Behavior

The popup allowing to save or update credentials should appear when clicking the blinking red icon.

## Steps to Reproduce (for bugs)

1. Install a recent Firefox (such as 62.0b14) on Linux (or maybe other systems)
2. Install keepassxc-browser
3. Connect to a keepassxc database
4. Go to https://secure.nikonsso.com/create-nikon-account (easy to create account, correctly auto-detected fields, no captcha) and fill up the fields manually or using the password generator icon
5. Either choose "KeepassXC-Browser / Save credentials" from the page contextual menu, or just submit the form.
6. KeepassXC-Browser icon will blink red, clicking it will not make the popup for saving the credentials appear.

## Debug info
KeePassXC-Browser Version: 1.2.0
KeePassXC Version: 2.3.3 
Operating system: Linux (ArchLinux/x86_64)
Browser: Firefox (62.0b14)
Proxy used: no

varjolintu:
A possible duplicate issue: https://github.com/keepassxreboot/keepassxc-browser/issues/211

samueltardieu:
@varjolintu I don't think so, in #211 nothing happens (not even the blinking red icon) as far as I can tell. Here the password change / save request is recognized, the problem is the missing popup.

samueltardieu:
Indeed, looks like it.

varjolintu:
Please follow the issue in the thread linked earlier.

PR #256, opened 2018-08-06
Fixed updating credentials from the context menu

varjolintu:
When using context menu to save changed password the popup dialog is closed if no username is available.

This fix enables to use the username of the selected entry in the popup, and also disables the New button when no username is used for credential saving.

If user tries to update an old password (using the context menu from the old password's input field), a notification will be displayed and the popup is closed.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/211.

### DIFF
--- a/keepassxc-browser/popups/popup_remember.js
+++ b/keepassxc-browser/popups/popup_remember.js
@@ -5,15 +5,20 @@ var _tab;
 function _initialize(tab) {
     _tab = tab;
 
-    // no credentials set or credentials already cleared
-    if (!_tab.credentials.username) {
+    // No credentials set or credentials already cleared
+    if (!_tab.credentials.username && !_tab.credentials.password) {
         _close();
         return;
     }
 
-    // no existing credentials to update --> disable update-button
+    // No existing credentials to update --> disable Update button
     if (_tab.credentials.list.length === 0) {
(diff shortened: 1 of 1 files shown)


samueltardieu:
Note: it doesn't fix #253 for creating passwords on Firefox 62.0-b14 (blinking red, but no popup to create password, on Nikon site described in #253).

varjolintu:
@samueltardieu Everything works with stable build of Firefox, which we support. Beta and Nightly builds can have errors like this, and we would be exhausted to hunt down every tiny change they make.

samueltardieu:
Ok, that is your right of course not to support beta versions. However, I would have thought that it was helpful to you to report problems that will happen in future versions in advance.

varjolintu:
@samueltardieu That is helpful indeed, so thanks. I have a list of issues which happen with Beta and Nightly builds. I'll always check them when new stable version is released.

varjolintu:
TODO: If context menu was used from the old password input field, the old password is saved instead.

varjolintu:
Finally fixed the issue above. A separate `retrieve_credentials` was needed for this. If the user has not allowed the remember option, a new allowance dialog will be shown before the actual credential update dialog.

Issue #234, opened 2018-07-22
Don't show »update keepassxc« if check updates = never


jo-so:
## Expected Behavior
When I select *never* for the option *Check for updates of KeePassXC*, I don't expect I'll ever get a warning about an outdated version. If I'm using KeePassXC from a distribution like Debian stable, I don't want to go and download the newest version from the website, but I have to use the newest version of the browser extension from the Firefox library.

PR #285, opened 2018-08-31
Hide update message when updates are never checked

varjolintu:
Hides the update message of KeePassXC if the user has selected option `Never` from the settings.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/234.

### DIFF
--- a/keepassxc-browser/background/event.js
+++ b/keepassxc-browser/background/event.js
@@ -206,7 +206,7 @@ kpxcEvent.onCheckUpdateKeePassXC = function(callback, tab) {
 };
 
 kpxcEvent.onUpdateAvailableKeePassXC = function(callback, tab) {
-    callback(keepass.keePassXCUpdateAvailable());
+    callback(page.settings.checkUpdateKeePassXC > 0 ? keepass.keePassXCUpdateAvailable() : false);
 };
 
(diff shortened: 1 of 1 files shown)


keepassxc-browser/background/event.js
```diff
@@ -206,7 +206,7 @@ kpxcEvent.onCheckUpdateKeePassXC = function(callback, tab) {
 };
 
 kpxcEvent.onUpdateAvailableKeePassXC = function(callback, tab) {
-    callback(keepass.keePassXCUpdateAvailable());
+    callback(page.settings.checkUpdateKeePassXC > 0 ? keepass.keePassXCUpdateAvailable() : false);
```
droidmonkey: I think it would be better to just never enter this function to begin with.

varjolintu: That's not a function. It's the variable where the setting is stored. `0` means it never checks for the update. The real function where the check happens is `keepass.keePassXCUpdateAvailable()`.

droidmonkey: No what I meant was this whole block is an anonymous function, just don't call it (ie, ```kpxcEvent.onUpdateAvailableKeePassXC()```) if the update check is disabled.

droidmonkey: Nevermind, I see how this is used now, I was confused.

PR #304, opened 2018-09-29
Shortcut improvements

stefansundin:
- Get current shortcut keys.
- Add button that opens shortcut page, it can be opened by using `chrome.tabs.create()`.

Hi everyone. Thanks for a great tool and extension. I decided to make some small improvements to the explanations for the shortcuts and also a button that opens the page in Chrome where you can configure the shortcuts.

Although `chrome://extensions/configureCommands` now redirects to `chrome://extensions/shortcuts`, I left it as the old URL to be more backwards compatible.

I tested this in Firefox, and getting the shortcut keys work as expected. It is now possible to customize the shortcut keys in recent Firefox versions, but it is not as easy as in Chrome, since the extension must provide an input field and then trigger the update.

Hope you like it!

### DIFF
--- a/keepassxc-browser/options/options.js
+++ b/keepassxc-browser/options/options.js
@@ -118,6 +118,18 @@ options.initGeneralSettings = function() {
     $('#blinkMinTimeout').val(options.settings['blinkMinTimeout']);
     $('#allowedRedirect').val(options.settings['allowedRedirect']);
 
+    browser.commands.getAll().then(function(commands) {
+        commands.forEach(function(command) {
+            var shortcut = document.getElementById(`${command.name}-shortcut`);
+            if (!shortcut) return;
+            shortcut.textContent = command.shortcut || 'not configured';
+        });
+    });
+
+    $('#configureCommands').click(function(){
+        browser.tabs.create({ url: 'chrome://extensions/configureCommands' });
+    });
+
     $('#blinkTimeoutButton').click(function(){
         const blinkTimeout = $.trim($('#blinkTimeout').val());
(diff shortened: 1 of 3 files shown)


keepassxc-browser/options/options.js
```diff
@@ -118,6 +118,18 @@ options.initGeneralSettings = function() {
     $('#blinkMinTimeout').val(options.settings['blinkMinTimeout']);
     $('#allowedRedirect').val(options.settings['allowedRedirect']);
 
+    chrome.commands.getAll(function(commands) {
```
varjolintu: Please do not use `chrome` namespace. Browser-polyfill ensures you can use `browser` everywhere.

keepassxc-browser/options/options.js
```diff
@@ -118,6 +118,18 @@ options.initGeneralSettings = function() {
...
+        commands.forEach(function(command) {
+            var shortcut = document.getElementById(`${command.name}-shortcut`);
+            if (!shortcut) return;
+            shortcut.textContent = command.shortcut || 'not configured';
+        });
+    });
+
+    $('#configureCommands').click(function(){
+        chrome.tabs.create({ url: 'chrome://extensions/configureCommands' });
```
varjolintu: Same here.

varjolintu:
Thanks for the PR. I'll test this with macOS too.

varjolintu:
Just tested this. Works great! Maybe it could be possible to make those necessary input fields for Firefox >= 60 in the future so it could have the same functionality. Thanks.

Could you rebase the branch?

stefansundin:
Done. I think you can get the same result by using the "Squash and merge" option in the dropdown menu.

https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits

Issue #305, opened 2018-09-29
Different behavior after database unlock on macOS and Windows


whisdol:
## Expected Behavior
KeePassXC-Browser should show the matching credentials on a page after the database is unlocked. This is working as expected with Firefox on Windows.

## Current Behavior
After the database is unlocked, users need to switch tabs back and forth or click "Redetect credential fields" to retrieve the credentials for the current site.

## Steps to Reproduce
1. Open Firefox with KeePassXC-Browser
2. Open KeePassXC
3. Make sure the extension shows "Database not opened"
4. Open a site with credential fields with a matching entry in the KeePassXC database, e.g. https://github.com/login
5. Click into the username field (no credentials will be suggested as the database is locked)
6. Switch to KeePassXC and unlock the database
7. Return to Firefox

On Windows, the matching credentials will now be listed below the username field. On macOS, the credentials are not shown.

## Debug info
KeePassXC - 2.3.4
KeePassXC-Browser - 1.2.0
Operating system: Mac
Browser: Firefox

varjolintu:
Can you reproduce this with the latest 1.3.0 release?

whisdol:
Yes, 1.3.0 also requires a reload/tab switch after unlocking the database on macOS.
I tried my hand at debugging this and noticed that:
1. On the first database unlock when `cip.detectDatabaseChange` gets called, response.new is not equal to response.old (as expected). The credentials are only received/displayed under the login field if I debug the javascript file with some breakpoints in the `cip.detectDatabaseChange` and `cip.initCredentialFields` functions. If I do not set any breakpoints, the credentials are not received/displayed.
2. If I then lock the database, keep the browser open, (reload the page,) and unlock the database again, the credentials are never received. In the `cip.detectDatabaseChange`-function, response.new now equals response.old, so `cip.initCredentialFields` is not called again.
3. (Unrelated to this issue) If credentials are loaded and I lock the database via the lock button in the extension, the credentials are not cleared.

Hope that helps a bit. Let me know if there is anything I should dig into more.

droidmonkey:
This sounds like a race condition or a discarding of a promise.

varjolintu:
@whisdol Thanks. That helps a lot! I'm already investigating the issue.

varjolintu:
Fixed. The database hashes were not properly updated to content scripts when using the popup icon.

PR #309, opened 2018-10-11
Fix unlocking the database

varjolintu:
When locking the database from the popup lock icon the database hashed are not updated properly to the contect script.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/305.
Fixes https://github.com/keepassxreboot/keepassxc/issues/2480.

Also modifies the login popup to a locked state after clicking the lock icon.

### DIFF
--- a/keepassxc-browser/background/event.js
+++ b/keepassxc-browser/background/event.js
@@ -87,7 +87,7 @@ kpxcEvent.onShowNotification = function(callback, tab, message) {
 
 kpxcEvent.showStatus = function(configured, tab, callback) {
     let keyId = null;
-    if (configured) {
+    if (configured && keepass.databaseHash !== '') {
         keyId = keepass.keyRing[keepass.databaseHash].id;
     }
 
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -17,8 +17,8 @@ keepass.nativePort = null;
 keepass.keySize = 24;
 keepass.latestVersionUrl = 'https://api.github.com/repos/keepassxreboot/keepassxc/releases/latest';
 keepass.cacheTimeout = 30 * 1000; // milliseconds
-keepass.databaseHash = 'no-hash'; //no-hash = KeePassXC is too old and does not return a hash value
-keepass.previousDatabaseHash = 'no-hash';
+keepass.databaseHash = '';
+keepass.previousDatabaseHash = '';
 keepass.keyId = 'keepassxc-browser-cryptokey-name';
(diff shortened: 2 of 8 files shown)


keepassxc-browser/background/event.js
```diff
@@ -87,7 +87,7 @@ kpxcEvent.onShowNotification = function(callback, tab, message) {
 
 kpxcEvent.showStatus = function(configured, tab, callback) {
     let keyId = null;
-    if (configured) {
+    if (configured && keepass.databaseHash !== 'no-hash') {
```
droidmonkey: I do not like "magic values". Where is this set? Wouldn't a blank databaseHash be sufficient and just as telling?

keepassxc-browser/background/keepass.js
```diff
@@ -1062,3 +1048,23 @@ keepass.updatePopup = function(iconType) {
...
+            keepass.updatePopup(configured ? 'normal' : 'cross');
+
+            // Send message to content script
+            browser.tabs.query({ active: true, currentWindow: true }).then((tabs) => {
+                if (tabs.length) {
+                    browser.tabs.sendMessage(tabs[0].id, {
+                        action: 'check_database_hash',
+                        hash: {old: keepass.previousDatabaseHash, new: keepass.databaseHash}
+                    });
+                    keepass.previousDatabaseHash = keepass.databaseHash;
```
droidmonkey: If we are setting the previous hash to the "new" hash, when is the "new" hash set? This function makes no sense to me.

varjolintu: The new hash is set when getDatabaseHash() has a response. This is called quite often.

keepassxc-browser/background/event.js
varjolintu: Probably. This is originally from chromeIPass. Basically `no-hash` means the database is locked.

droidmonkey: Yeah i kind of figured, we can look at it later if necessary

keepassxc-browser/background/keepass.js
droidmonkey: Wouldn't this be subject to race conditions then?

varjolintu: No. Updating these values cannot happen at the same time. In this case the update function is called only when user is locking or unlocking the database from KeePassXC or from the extension popup. And updating the current database hash only happens when doing some operation (retrieving credentials at page load etc).

keepassxc-browser/background/event.js
varjolintu: I changed the "magic value" to an empty string.

PR #320, opened 2018-10-26
refactor: return a promise instead of using a callback function

thetric:
The callback-based api will soon be removed. `webextension-polyfill` logs a warning in the console:

> Returning a Promise is the preferred way to send a reply from an onMessage/onMessageExternal listener, as the sendResponse will be removed from the specs (See https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage)


varjolintu:
I have noticed this error message too. Thanks for the PR. I'll review this soon.

Issue #308, opened 2018-10-10
Keyboard shortcuts to fill username/password do not work on Jira login form


heapifyman:
## Expected Behavior
Keyboard shortcuts to fill username/password (Alt+Shift+U and Alt+Shift+I) should work on Jira login form.

## Current Behavior
Pressing Alt+Shift+U on Jira login page has no effect. Pressing Alt+Shift+I opens the "Issues" menu dropdown.

## Possible Solution
It would be ideal, if user could specify custom shortcuts in keepassxc-browser for filling username/password. That way one could choose a shortcut that is not likely to collide with any shortcuts in other applications. Also, one could use the same shortcut on both Linux and Mac.

## Steps to Reproduce (for bugs)
1. Go to a Jira login page (e.g. https://www.jfrog.com/jira/login.jsp?os_destination=%2Fbrowse%2FRTFACT)
2. Focus login field
3. Press Alt+Shift+U

## Debug info
KeePassXC - 2.3.4
KeePassXC-Browser - 1.3.0
Operating system: Linux
Browser: Firefox 62.0.3 (64-Bit)

varjolintu:
You can define your own keyboard shortcuts with Chromium-based browsers. It's possible to do with Firefox too, but it's more complicated and didn't make yet to 1.3.0 release.

heapifyman:
is there an ETA for custom keyboard shortcuts support in FF?

varjolintu:
Not at this point. The main problem with is that you have to enter the keyboard shortcut manually to an input field, and the whole feature is not supported in older Firefox releases. I haven't event started planning the feature yet any further.

varjolintu:
PR ready for configuring keyboard shortcuts with Firefox.

PR #326, opened 2018-10-29
Keyboard shortcut configuration page for Firefox

varjolintu:
Adds a missing keyboard shortcut configurator for Firefox. However, this is still WIP because all platforms are not tested yet. Also, the minimum version of Firefox must be 60. See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/commands#Browser_compatibility

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/308.

A screenshot.
screenshot


varjolintu:
Resolved conflicts and rebased.

Issue #339, opened 2018-11-28
Won't fill multiple passwords on multiple pages


nagromc:
## Expected Behavior

The [ProtonMail login page](https://mail.protonmail.com/login) is composed of:
1. Page one: the username and password
2. Page two (optional): the OTP (if a 2FA has been activated)
3. Page three (optional): the mailbox password (if you have this particular [password enabled](https://protonmail.com/support/knowledge-base/switch-two-password-mode/))

On page two, I can use <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>T</kbd> to fill the password.

One page three, I expect the _Mailbox password_ field to be filled.

## Current Behavior

On page three, I cannot find a way to autofill the _Mailbox password_ field (which is different from the account pass word from step 1).

## Possible Solution

I'm not sure how I am supposed to play with the UI, the `storage.js`, or the `KPH:` attributes in my KeePass custom fields.

I am currently using KeePass with the [KeePassNatMsg](https://github.com/smorks/keepassnatmsg) plugin. Maybe I must use KeePassXC.

A documentation may help, or maybe it is a PEBCAK problem.

## Debug info

KeePass - 2.40
[KeePassNatMsg](https://github.com/smorks/keepassnatmsg) plugin - 2.0.3.0
KeePassXC-Browser - 1.3.0
Operating system: Win
Browser: Firefox

varjolintu:
Is the page URL for mailbox password different from the default one? If so, you could make another entry with that URL and the mailbox password.

nagromc:
> Is the page URL for mailbox password different from the default one?

Yes. Page one and two are https://mail.protonmail.com/login. Page three is https://mail.protonmail.com/login/unlock.

I realized my issue description was incomplete. At the moment, I have one unique entry in my KeePass DB:
- User name field: my ProtonMail user name
- Password field: my ProtonMail **account** password
- URL field: https://mail.protonmail.com/login
- A custom String field: my ProtonMail **mailbox** password
    - At the moment, its name is _Clé déchiffrement_ (_Decryption key_ in English), but can be changed if needed. That was the `KPH:` attempt I was talking about in the initial issue.

So you are telling me I need to split this entry in two:
1. Main entry
    - User name field: my ProtonMail user name
    - Password field: my ProtonMail **account** password
    - URL field: https://mail.protonmail.com/login
2. Secondary entry
    - User name: empty?
    - Password  field: my ProtonMail **mailbox** password
[...]

PR #340, opened 2018-12-03
Add URL change detection to MutationObserver

varjolintu:
Adds URL change detection to MutationObserver. Credential fields are reinitialized when the change happens. This ensures that some pages detect another password field page properly (e.g. ProtonMail's Mailbox password).

Also modifies `cipFields.isSearchField` slightly to ignore word `research` from `search`.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/339.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -15,6 +15,9 @@ var _detectedFields = 0;
 // Element id's containing input fields detected by MutationObserver
 var _observerIds = [];
 
+// Document URL
+let _documentURL = document.location.href;
+
 browser.runtime.onMessage.addListener(function(req, sender, callback) {
     if ('action' in req) {
         if (req.action === 'fill_user_pass_with_specific_login') {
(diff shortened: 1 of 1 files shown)


keepassxc-browser/keepassxc-browser.js
```diff
@@ -15,6 +15,9 @@ var _detectedFields = 0;
 // Element id's containing input fields detected by MutationObserver
 var _observerIds = [];
 
+// Document URL
+let _documentURL = document.location.href;
```
RealOrangeOne: Careful using `let`, it's not supported on some older browsers, and will result in a syntax error. https://caniuse.com/#feat=let

droidmonkey: I don't see an issue? There are 0.25% usage of desktop browsers that don't support it. And I believe of that percentage, 80% is IE which doesn't support WebExtensions.

RealOrangeOne: True. My main reservation was in trying to keep it consistent with the rest of the codebase, which seems to use `var`.

varjolintu: I need to change this behaviour in the future. `var` is preferred when the variable must be used globally in the other scripts. `let` keeps the scope in one file.

varjolintu:
Resolved conflicts and rebased.

Issue #342, opened 2018-12-07
Feature request: fine-grained control for Autofill HTTP Basic Auth


jsachs:
Related to #259 

We've hard disabled Autofill HTTP Basic Auth within our organization because of the scary things that can happen when credentials are silently submitted to Basic Auth. However, we have a few users with workflows that are dramatically improved by using HTTP Basic Auth. We'd like them to be able to use this feature on these workflows that we've already vetted. It feels like a good way to support this would be adding an option to allow only specific entries to use HTTP Basic Auth autofill.

We can see this working one of two ways, so this issue might also belong in the keepassxc repo. There could be a browser option that requires you to confirm an entry being used for HTTP autofill the first time it is used. There could also be an implementation of this where indivitual entries have a checkbox or support a metadata entry than indicates to the browser that this specific entry is allowed to be used for autofill.

varjolintu:
It could be done. This is also related: https://github.com/keepassxreboot/keepassxc-browser/issues/273.

varjolintu:
This will be fixes with the referenced PR (and https://github.com/keepassxreboot/keepassxc/pull/2542). When credentials are retrieved and allowance checked from HTTP Basic Auth dialogs, those permissions will be handled separetely. Denying access in Basic Auth will not affect to the permissions on the actual site.

When combined with https://github.com/keepassxreboot/keepassxc/pull/2143 the behaviour is extended even further.

varjolintu:
The behaviour of the related PR will be changed so the confirmation popup for HTTP Basic Auth will trigger every time instead of saving separate permissions. It's more simple approach, and it's possible to deny those entries from other pages.

Issue #273, opened 2018-08-24
HTTP Auth & form on the same site


Neraud:
When a site uses both HTTP Auth and a standard form login, you can't automatically inject both passwords.

## Expected Behavior
Ideally : 
 - the user can flag each entry to be used for HTTP Auth, Logon forms, or both
 - the browser extension sorts between the available entries depending on the type of login required
 - if a single one is applicable, it automatically injects, else it fallbacks to the manual user selection

Also, the "Return only best-matching credentials" setting should still work.

Worst case example : 
http://domain/app1 & http://domain/app2 have different HTTP Auth and Login forms.
The user should be able to have 4 entries : 
 - http://domain/app1 for HTTP Auth
 - http://domain/app1 for Login form
 - http://domain/app2 for HTTP Auth
 - http://domain/app2 for Login form

No manual interaction should be required once these entries are properly set up.

## Current Behavior
Currently, I don't think I can flag an entry for HTTP Auth and another for Login form.
So when I have those 2 entries available for the same site, I am prompted to select the one for HTTP Auth, and then have to select the one for Login form.

## Possible Solution
I don't know how the extension works internally, but I guess adding variables to the "KeePassXC-Browser Settings" could be a solution.

Adding a new couple of optional attributes there could work.
For example : 
`{
	"Allow": ["domain"],
	"Deny": [],
	"Realm": "",
	"AllowForHttpAuth": true,
	"AllowForLoginForm": true
}`

To preserve the behaviour for existing entries, I assume having both set to true by default would work.

Best UX would be to use a dialog on the browser to let the user flag his entries when multiple ones are applicable, but I assume that's a complicated enhancement. 
I don't think many users are impacted by this use case, so I guess having to manually edit the attribute would be sufficient.

## Steps to Reproduce (for bugs)
Unfortunately I can't give you access to the site I'm having issue with.
Any site with both HTTP Auth and Login form would work.

## Debug info

KeePassXC - 2.3.3
keepassxc-browser - 1.2.0
Operating system: Win
Browser: Chrome & Firefox
Proxy used: YES

PR #343, opened 2018-12-08
Retrieve credentials with HTTP Basic Auth option added

varjolintu:
When retrieving credentials from HTTP Basic Auth dialogs, an additional option is added to the request. KeePassXC side handles the permissions for HTTP Basic Auth separetely. If an older KeePassXC version is used, this extra parameter has no affect.

Related KeePassXC PR: https://github.com/keepassxreboot/keepassxc/pull/2542

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/273.
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/342.

This requires KeePassXC 2.4.0 so merging this will have to wait.

### DIFF
--- a/keepassxc-browser/background/httpauth.js
+++ b/keepassxc-browser/background/httpauth.js
@@ -51,7 +51,7 @@ httpAuth.retrieveCredentials = function(tabId, url, submitUrl, forceCallback) {
     return new Promise((resolve, reject) => {
         keepass.retrieveCredentials((logins) => {
             resolve(logins);
-        }, tabId, url, submitUrl, forceCallback);
+        }, tabId, url, submitUrl, forceCallback, false, true);
     });
 };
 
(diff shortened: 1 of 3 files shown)


PhilippVerpoort:
Sorry, I'm a bit confused.

In my version of the Firefox browser plugin (version 1.5.1, dated September 28th), I can each time select which password to use to fill the HTTP Auth. But it's a long list and includes other credentials that should be used on a form on the site, and not for HTTP Auth. Is there no way to define which credentials should be used for HTTP Auth (hence removing all others from that list)?

varjolintu:
@PhilippVerpoort It's not possible at this point. Maybe in the future because we will roll a separate Browser Integration page for the entries.

kmichailg:
+1 on showing only entries with `Use this entry only with HTTP Basic Auth`. I've reached a point where I have tons of logins on a UAT server with HTTP Basic Auth enabled. This would be a very welcome change.

varjolintu:
@kmichailg With KeePassXC 2.7.0 you can set a whole group of credentials to HTTP Basic Auth only if you wish.

PR #354, opened 2018-12-18
Fix reconnecting the extension with older KeePassXC releases

varjolintu:
Added an extra check to use publicKey as identification ID with older KeePassXC releases. Fixes multiple issues with Linux distributions based on Ubuntu which still deliver KeePassXC 2.3.1 as default.

Broken with https://github.com/keepassxreboot/keepassxc-browser/pull/258.


keepassxc-browser/background/keepass.js
```diff
@@ -409,7 +409,9 @@ keepass.associate = function(callback, tab) {
                     keepass.handleError(tab, kpErrors.ASSOCIATION_FAILED);
                 }
                 else {
-                    keepass.setCryptoKey(id, idKey);    // Save the new identification public key as id key for the database
+                    // Use public key as identification key with older KeePassXC releases
+                    const savedKey = keepass.currentKeePassXC.versionParsed > 231 ? idKey : key;
```
phoerious: Just a general remark: how is this supposed to work if we ever have a patch release with two digits?

varjolintu: Not good. It would return `2.4.10` as `2410` (and a possible `2.4.0.1` as `2401`).

phoerious: And 2.41.0 would be the same as 2.4.10, right? This either needs a dedicated compare function or you need to pad with zeros up the point that it becomes unlikely that we will ever have a version with more digits (e.g., 2004000 for 2.4.0).

varjolintu: Or then make an object that holds major, minor, patch and build. For this fix you can actually use that numeric compare because no version will go under it. `keepass.currentKeePassXC` also holds `version` variable which is the textual form. In this fix that could be used too with just checking `"2.3.0"` and `"2.3.1"`. The minimum version accepted by the extension is 2.3.0 anyway.

varjolintu:
Added two more functions to the PR:
`keepass.parseVersion(version)` parses the version string to an object. This allows retrieving and saving the same property to localStorage without overriding the old one:
```
{
    'version': <string>, 
    'parsed': 
        {
            'major': <integer>, 
            'minor': <integer>,
            'patch': <integer>,
            'build': <integer>,
            'snapshot': <boolean>
        }
};
```

`keepass.compareVersion(minimum, current)` takes version strings and compares them using the object above. Returns false if `current` is smaller than `minimum`.

keepassxc-browser/background/keepass.js
```diff
@@ -1072,3 +1070,46 @@ keepass.updateDatabase = function() {
...
+keepass.compareVersion = function(minimum, current) {
+    const min = keepass.parseVersion(minimum).parsed;
+    const cur = keepass.parseVersion(current).parsed;
+
+    if (min === undefined || cur === undefined) {
+        return false;
+    }
+
+    if (min.major !== cur.major) {
+        return min.major < cur.major;
```
phoerious: Why not return an AND series of <= comparisons?

keepassxc-browser/background/keepass.js
```diff
@@ -1072,3 +1070,46 @@ keepass.updateDatabase = function() {
         });
     }, null);
 };
+
+// Parse the version number. Major, minor and patch are mandatory
+keepass.parseVersion = function(version) {
+    const ver = /^(\d+)\.(\d+)\.(\d+)?\.?(\d+)?-?(\w+)?/.exec(version);
```
phoerious: Couldn't you just use split()?

varjolintu: That would need two splits if the version number contains `snapshot`, like `2.4.0-snapshot`.

phoerious: That should still be faster than a regex.

varjolintu: I'll try to improve this a bit.

varjolintu:
I ended up removing the version number parsing. It's not really needed. You can make a simple compare with the version number strings by just padding zeroes to the each number and then making a direct comparison. Now it's nice and clean.

varjolintu:
After this has been merged, https://github.com/keepassxreboot/keepassxc-browser/pull/126 can be implemented again.

PR #355, opened 2018-12-24
[WIP] Automatic reconnect

varjolintu:
Automatic reconnect reimplemented. Added an option which is enabled by default. Enabling or disabling it affects directly to the automatic reconnect mechanism.

It is disabled for Windows if KeePassXC is older than 2.3.4. This is because of the process spawning behaviour in Chromium-based browsers which needed a special handling to the KeePassXC side.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/264.
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/422.

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -451,6 +451,10 @@
         "message": "Automatically retrieve credentials.",
         "description": "Automatically retrieve credentials checkbox text."
     },
+    "optionsCheckboxAutomaticReconnect": {
+        "message": "Automatically reconnect to KeePassXC.",
+        "description": "Auto-reconnect checkbox text."
+    },
     "optionsCheckboxAutoFillSingleEntry": {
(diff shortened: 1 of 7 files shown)


droidmonkey:
After further testing this doesn't seem to work for me or it takes longer than I expected. After clicking the icon the reconnect worked, but pressing any of the keyboard shortcuts or using the right click menu did not elicit a connection.

varjolintu:
I just tested this with both Firefox and Chromium under W10. Automatic reconnect worked always. But when the automatic reconnect is disabled, using the popup or keyboard shortcuts should trigger the reconnect, true. I'll check the behaviour.

xeropresence:
What's the status on this pr? Does someone else need to give it a poke?

varjolintu:
@xeropresence I'm still in the middle of doing the last feature for this, which is not mandatory, but annoying if it's missing.

varjolintu:
This will be solved by #128.

Issue #357, opened 2018-12-28
Credentials available from Locked database after logging in (User) with KP database locked from being logged out


iverik-no:
I was able to fill in credentials with a locked database (actually closed Keepass!)

I use Protonmail in my example, though this wasn't the site on which I initially encountered the issue. I tested a few sites, and found the error to be present at Protonmail too. As it's more likely that someone here with debugging/testing/dev knowledge will have an account at Protonmail, I reasoned it makes more sense to refer mainly to that page.

It doesn't seem to happen every time. I had it occur many times in a row now while prodding at it, but then it stopped.

## Expected Behavior

To have to open the database before being able to fill in UN/PW.

## Current Behavior / What I observed

KeePassXC setting "Lock database when session is locked or lid is closed" is enabled.

What happened:
- I unlock my PC several hours after it locked itself (screensaver --> lock on resume).
- I have Firefox running (from before).
- I click on an already open tab, in which I'm not currently logged in.
- I click a credentials field
- The KeePass dropdown for filling in Username/Password appears (both entries I have for the site in question).
- I click one of them, and it supplies the username and password, thus logging me in.
- All the while KeePassXC wasn't running.

Testing, by unlocking KP database, locking user session, logging in user session, and then checking websites for the logins, I found these three different results happening:

1. I see the credentials dropdown already dropped down, ready for filling credentials. I found this on Protonmail.
2. I click the username field, as for "normal" Keepass login. Then credential options appear in the dropdown.
3. ...or nothing happens, as intended.

## Steps to Reproduce (for bugs)
1. Open database
2. Make Firefox detect it
3. Lock user session
4. Log back in
5. Go to e.g. Protonmail
6. Click the username field for credentials
7. Credentials appear

It doesn't seem to happen every time. I had it occur many times in a row, but then it stopped.

## Debug info

KeePassXC-Browser Version: 1.3.0
KeePassXC Version: 2.3.4 
Operating system: Linux / Ubuntu --> Kuuntu
Browser: Firefox 64

droidmonkey:
This sounds like the credentials are being cached and not cleared when the database is locked.

iverik-no:
Did a few tests just now, and refreshing the page does remove/clear the credentials from the page.
And they do clear "by them selves" after a short time when I start clicking in and out of the credentials fields.

varjolintu:
I'm quite sure this has something to do with autocomplete caching the values. Of course those should be also cleared when the database is locked.

[...]

PR #358, opened 2018-12-29
Clear credentials when the screen is locked

varjolintu:
Fix for clearing the credentials when locking the screen.

Document visibility is hidden when the screen is locked. That kind of situation was not handled in the extension. The database hashes are updated only when the current document/page is visible. Now, when the document is hidden, all credentials are cleared just in case.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/357.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -1401,22 +1401,26 @@ cip.init = function() {
     });
 };
 
+// Clears all from the content and background scripts, including autocomplete
+cip.clearAllFromPage = function() {
+    cipEvents.clearCredentials();
+
+    browser.runtime.sendMessage({
+        action: 'page_clear_logins'
+    });
+
+    // Switch back to default popup
+    browser.runtime.sendMessage({
+        action: 'get_status',
+        args: [ true ]    // Set polling to true, this is an internal function call
+    });
+};
+
 // Switch credentials if database is changed or closed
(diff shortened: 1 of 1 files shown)


droidmonkey:
This did not work for me. I went to a page that had credentials, they were loaded into the drop-down selection. I locked my computer and unlocked. When I returned KPXC was locked and the browser icon showed the locked icon, but the credentials were still loaded.

Why not call the clear credentials function when the icon is changed to the lock symbol?

varjolintu:
@droidmonkey Fixed. Check if it happens anymore.

whisdol:
While testing this (merged into PR #387), I ran into the following scenario (FF65, Windows 10):
1. Open Firefox and a [website with frames](https://www.kvb.koeln/app/index.html)
1. Open KeePassXC
1. Click reconnect
1. Click into the username field, credentials are suggested in the autocomplete-dialog. They are not listed in the extension menu. (Not sure if this 
1. Close KeePassXC
1. Lock screen
1. Unlock screen
1. Click into login field.

In my tries, the credentials were not cleared and still suggested in the autocomplete-dialog.

On a [website without frames](https://www.dkb.de/banking), the credentials were cleared after step 5.

varjolintu:
@whisdol Thanks! I'll check what causes this.

whisdol:
This fix works for me, credentials are now (also) cleared after closing KeePassXC.

Issue #386, opened 2019-01-18
Chase.com login fields fail to detect as of version 1.3.2


tfvlrue:
## Expected Behavior

Login fields should be detected on all websites.

## Current Behavior

The login fields on https://www.chase.com/ are no longer detected as of version 1.3.2

## Possible Solution

Tested in Firefox 64.0.2 and Chrome 71. This looks like a regression in version 1.3.2. I reverted back to 1.3.1 in Firefox and the login fields were detected once I clicked the "Redetect login fields" button.

## Steps to Reproduce (for bugs)

1. Navigate to https://www.chase.com, making sure the browser is full screen
2. The login box will appear in the upper right after a second.
3. Notice the fields are not auto-detected.
4. Click the "Redetect login fields" button. Notice they are still not detected (password generation icon does not appear, and credentials do not auto-fill). In version 1.3.1, it worked.

## Debug info

KeePassXC - 2.41
KeePassXC-Browser - 1.3.2
Operating system: Windows 7
Browser: Chrome/Firefox

Thanks for looking into it. Let me know if there's any other info I can provide.

droidmonkey:
I can confirm this as well

varjolintu:
It's broken by https://github.com/keepassxreboot/keepassxc-browser/pull/371. I made a fix where I moved some of the behaviour to be executed in each frame. Some sites need this.

I also noticed that the site URL varies a lot. Even when you go to the frontpage, the actual URL for the login is always `secure05b.chase.com` or similar. This is why the entry URL should be `https://chase.com` so it will be handled as `https://*.chase.com`.

PR #387, opened 2019-01-18
Revert #371

varjolintu:
#371 fixed some race conditions, but it's not working properly. Another type of solution is needed. The change needs to be reverted.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/386.
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/402.


varjolintu:
Doing some minor fixes to this PR, hold on.

whisdol:
Even with this fix (68bcdd1), the following workflow does not always work for me on https://www.kvb.koeln/app/index.html (scroll down and click "Login", the URL for the database entry is https://kvb.mobilesticket.de):
1. Start Firefox, KeePassXC
2. Open https://www.kvb.koeln/app/index.html, click "Login" so the credential fields are shown
3. Click Reopen Database in the extension
4. Enter Password in KeePassXC, return to Firefox
5. Click "Redetect Credential Fields"

It seems hard to reproduce now, but it definitely failed the first time I tried it. The autocomplete suggestions were only shown after switching tabs. I then tested it 5 more times, and the credentials were loaded as expected after redetecting the credential fields.

However, the 5. step should ideally not be necessary (but maybe that's not in the scope of this PR?).

varjolintu:
@whisdol I'll check those pages if I can reproduce it. It's preferred that everything should work without using the "Redetect credential fields" button.

whisdol:
Ok, I found a more reliable way to reproduce it:
1. Start Firefox
2. Open website (also tried this on https://www.dkb.de/banking/finanzstatus?$event=init, so I don't think it's necessarily restricted to websites with frames)
3. KeePassXC-Browser should say "Timeout or not connected to KeePassXC."
4. Start KeePassXC-Browser, unlock database
5. Return to Firefox
6. Click "Reload" in the extension, it should now be connected
7. Click "Redetect login fields"

With this workflow, credentials are never loaded for me.

varjolintu:
@whisdol I could not reproduce the latter one, but the first one does happen.

varjolintu:
You could try (or merge) https://github.com/keepassxreboot/keepassxc-browser/pull/358 because it's related to the issue you are describing.

varjolintu:
@whisdol Great! I'll check the popup behaviour.

varjolintu:
I ended up to a solution that reverts the #371. It just doesn't work properly and in the end it causes more problems than it solves. A much bigger refactor is needed for this, and it makes sense only if it's made after #128 is merged.

@droidmonkey Please review this again.

Sorry for the inconvenience guys.

PR #69, opened 2018-03-12
[WIP] Linting

varjolintu:
Preliminary support for eslint.
`npm run lint`

PR #391, opened 2019-01-20
Code cleaning

varjolintu:
Cleans the code:
- if-else is always on the same line
- Use Unix end-of-line sequences
- Use spaces when passing Objects or arrays
- Use single-line quotes
- Use `const` when possible


varjolintu:
@droidmonkey I used eslint with almost identical settings from #69. I'll upgrade the same settings to that PR. I wanted to make the actual code cleaning to a separate PR.

thetric:
> * if-else is always on the same line
> * Use Unix end-of-line sequences
> * Use spaces when passing Objects or arrays
> * Use single-line quotes

Some of these formatting rules can be automated using [Prettier](https://prettier.io/). Prettier can format many formats (HTML, JS, CSS, JSON, ...) and integrates with many editors (VS Code, IntelliJ/Webstorm, Vim, ...)

Issue #388, opened 2019-01-18
Cannot save authentication data in Chrome+Vivaldi


Lalarian:
Hello! I enter data into any registration form, and confirm.
The KeepassXC-Browser symbol flashes red, showing that it has recognized new authentication data to store. I click on the KeepassXC-Browser-Icon and it takes like 7 seconds until the little window opens, with KeepassXC-Browser showing the data it just captured.
Then I click on "New" for the extension to add the new authentication data to the connected database and the little window disappears, like it had save the data successfully - but it hasn't. The data I just entered i not in the connected database.

## Current Behavior

The newly added data is not addes to the database, despite all signs saying that it was

## Possible Solution
I don't know. Seems to be a bug, I think in the extension, and not in KeepassXC itself, because it's working fine from Waterfox/Firefox.

## Steps to Reproduce (for bugs)
1. enter registration data on web.de (but any other website is same!)
2. Confirm correctness of your personal data
3. Click the red flashing KeepassXC icon
4. Click 'new'
5. Check the databse for the freshly created account and see: Its not there.

## Debug info

KeePassXC - 2.4.0-snapshot Revision: 726bbb2
KeePassXC-Browser - Version: 1.3.2
Operating system:Linux - Xubuntu 18.04
Browser: Chrome+Vivaldi

varjolintu:
Seems to work from the context menu with no problems. #387 is possible fix for this. Could you try it?

jenzinger:
I have the exact same issue. I am on MacOS with KeePassXC 2.3.4 and XC Browser 1.3.2 in Chrome 71.0.3578.98.

I made 2 curious observations:

1. It worked once, but then never again, not even after reinstalling extension. The same behavior occured when I used KeePassHttp-Connector. It saved once to the KeePassXC database, then never again. This makes me wonder whether this is something more general with the communication with the main application?

2. When I click on save credentials in the extension pop-up, the database file seems to be modified, according to the entry for the file on my hard drive. Also I am syncing the database file and the sync updates the file on my remote drive - however the new password entry is not seen when I open the database. Closing and reopening KeePassXC does not change that.

droidmonkey:
just to make sure, did you look in the right folder?

![image](https://user-images.githubusercontent.com/2809491/51794103-9e3cac00-2199-11e9-9856-63487f0743f0.png)

Lalarian:
[...]

Issue #382, opened 2019-01-16
Allow selection of KeePassXC group when creating new entry


wohali:
## Expected Behavior
When creating a new entry, I should be able to select the group within which I place the entry. For instance:
* Work (Job 1)
* Work (Job 2)
* Personal
* Financial
* Government
* Online Retailers
* Gaming

etc.

These would map to the groups inside of KeePassXC.

The dialogue could even allow for creation of new groups inside of KeePassXC.

## Current Behavior
All browser passwords are placed in a single group, KeePassXC-Browser Passwords. The user must then manually drag these to the group after creation - not convenient.

## Possible Solution
Provide a popup window on creation of entry that allows selection of group.

## Steps to Reproduce (for bugs)
Not a bug.

## Debug info
Not a bug.

varjolintu:
This would be very nice. We need to discuss if this would be more convenient to add to the KeePassXC side instead.

wohali:
I can see adding the bulk of the functionality to KeePassXC, but the point is that you can pick the group at the instant you create the new entry, which occurs in the dropdown in the browser.

Inspiration of course comes from Kee:

![mwsnap 2019-01-22 10_38_16](https://user-images.githubusercontent.com/112292/51546290-deb2c780-1e31-11e9-8797-0ee0a2baf040.png)

varjolintu:
That would be possible too. KeePassXC 2.40 is released soon so I'll try to add this behaviour before that (just a basic support for a group parameter: if it's set, then use a group set by the user, otherwise use the default).

droidmonkey:
We can also aim for 2.4.1 to add the hook. But if you can make it ready before the beta this week that would be awesome.

wohali:
@varjolintu would the group list be exposed to the browser plugin as well, to populate the dropdown? Icons are only a nice to have.

varjolintu:
@wohali That would be needed, yes.

varjolintu:
It's done. See https://github.com/keepassxreboot/keepassxc-browser/pull/396 and https://github.com/keepassxreboot/keepassxc/pull/2637.

PR #396, opened 2019-01-23
Select group when adding new credentials

varjolintu:
Allows selecting a group when adding new credentials to KeePassXC from the popup dialog.

A new command `get-database-groups` is added to the protocol. If older version than KeePassXC 2.4.0 is used, credentials are saved to the default KeePassXC-Browser Passwords group. For saving, both name and group UUID are used.

The groups are read from a dynamic array where each group can contain N children:
```
{
    name: '<group name>',
    uuid: '<group UUID>',
    children: [...]
}
```

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/382.
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/388.
Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/427.

Related KeePassXC PR: https://github.com/keepassxreboot/keepassxc/pull/2637.

groups_keepassxc

groups_popup

Also, a default group can be set, or the group can be asked every time:
Screen Shot 2019-03-09 at 10 32 40


varjolintu:
I updated the PR to include a tree view.

droidmonkey:
Is the tree collapsed by default? (Should be) the arrows don't seem to indicate the correct state.

varjolintu:
@droidmonkey It's not collapsed, or cannot be collapsed. It's just a tree view. What do you suggest instead of the arrows? Or should it definitely be able to collapse?

varjolintu:
I changed the icon to arrow down, indicating open tree. For now.

keepassxc-browser/background/keepass.js
```diff
@@ -696,6 +702,71 @@ keepass.lockDatabase = function(tab) {
...
+                const res = keepass.decrypt(response.message, response.nonce);
+                if (!res) {
+                    keepass.handleError(tab, kpErrors.CANNOT_DECRYPT_MESSAGE);
+                    callback([]);
+                    return;
+                }
+
+                const message = nacl.util.encodeUTF8(res);
+                const parsed = JSON.parse(message);
+                keepass.setcurrentKeePassXCVersion(parsed.version);
```
droidmonkey: this seems out of place

keepassxc-browser/background/keepass.js
```diff
@@ -696,6 +702,71 @@ keepass.lockDatabase = function(tab) {
...
+
+                if (keepass.verifyResponse(parsed, incrementedNonce)) {
+                    groups = parsed.groups;
+                    keepass.updateLastUsed(keepass.databaseHash);
+                    callback(groups);
+                } else {
+                    console.log('getDatabaseGroups rejected');
```
droidmonkey: Would you want to show a proper error at this point? Would the CANNOT_DECRYPT_MESSAGE be appropriate here?

droidmonkey: Something like "Invalid or corrupt message sent from KeePassXC"

varjolintu: Something like that yes. Currently the verifyResponse() returns false from multiple reasons, so the message must be a general one. The one you proposed or something like: "Message sent from KeePassXC cannot be verified".

varjolintu:
I added one more thing to the PR. A default group can be specified, or the user can set an option that asks for the group every time when saving new credentials.

Screen Shot 2019-03-09 at 10 32 40

droidmonkey:
There is still a problem. When defining the group to save to, if it does not exist then the entry is silently dropped. The group should be created to store the new entry if it doesn't already exist. This will likely require changes on KeePassXC side.

varjolintu:
It also seems that triggering notifications from the popup doesn't work. Some changes are needed for that too. Probably we need to call the background script to activate them.

varjolintu:
@droidmonkey That is now done. KeePassXC side support is here: https://github.com/keepassxreboot/keepassxc/pull/2790.

keepassxc-browser/_locales/en/messages.json
```diff
@@ -312,9 +312,13 @@
         "description": "A popup message shown choosing what group user wants to use for new credentials."
     },
     "popupRememberErrorDefaultGroupNotFound": {
-        "message": "Error: Specified default group not found.",
+        "message": "Error: Specified default group not found. Creating all necessary groups.",
```
droidmonkey: This is not an error in this case, just an info. Should be reworded as: "Info: Default group created in connected database"

keepassxc-browser/_locales/en/messages.json
```diff
@@ -312,9 +312,13 @@
...
         "description": "Error message shown when default group set cannot be found."
     },
+    "popupRememberErrorCreatingNewGroup": {
+        "message": "New group(s) cannot be created.",
```
droidmonkey: "Error: .... "

droidmonkey:
I thought of a little vulnerability with create entry/group. There is really nothing stopping abuse of the API to create an entry or group. We prevent retrieving information from the database by showing allow/deny dialog, but we do not prevent writing to the database (from KeePassXC).

varjolintu:
@droidmonkey Yes. We need to restrict this in some way.

droidmonkey:
We can introduce the protections in 2.4.1

Issue #25, opened 2018-02-28
Extension icon is missing HiDPI support


ArchangeGabriel:
The icon is pixellised.

## Expected Behavior
The icon should be crisp on HiDPI display.

## Current Behavior
The loaded icon does not seem to take into account HiDPI.

## Possible Solution
Other extensions do it, so it should be easy.

## Steps to Reproduce (for bugs)
1. Open Firefox in HiDPI mode.
2. See all icons being crisp excepted for keepassxc-browser.

## General Info
KeePassXC - N/A
keepassxc-browser - 1.0

Operating system: Linux
Browser: Firefox (Nightly)
Proxy used: N/A

varjolintu:
FYI: Chromium doesn't support SVG icons in the manifest.json.
This means a different manifest.json should be used for WebExtension and Chromium based browsers. So let's put this to a wishlist.

ArchangeGabriel:
Is this supposed to be fixed in 1.3.3? Because it does not seems so…
![addons](https://user-images.githubusercontent.com/490531/55336853-220e5f80-548d-11e9-94e1-aea7dff4f69c.png)

ArchangeGabriel:
Also, below is on Chromium, and even if it does not support SVG, why does Decentraleyes and uBO look OK and keepassxc-broswer does not?
![chromium](https://user-images.githubusercontent.com/490531/55337000-67329180-548d-11e9-8966-1a488a7301c4.png)

This is a minor nitpicking, but this is one of the only flaw of this extension. ;)

varjolintu:
No it's not fixed yet. It's in the next release. Sorry or closing this.

We need to make some testing and see which one of the icons is suitable for Chromium so it doesn't look blurred.

ArchangeGabriel:
@varjolintu Oh OK I see, the commit only allowed to fix this, but actually does not, correct?

Maybe you should ask Decentraleyes dev, @Synzvato? Or @gorhill.

varjolintu:
I think it would be ok to just include all possible sizes to `manifest.json`. I don't have a HiDPI display so I cannot directly test it though.

varjolintu:
I just noticed that when icons are changed dynamically in the code, the size 19x19 is always used. I'll fix this to the No jQuery branch. Of course the size of those icons must be much bigger than that.

PR #398, opened 2019-01-25
Create separate manifests using a build script

varjolintu:
Create separate manifest files for Firefox and Chromium based browsers. This makes possible to use SVG icons with Firefox version of the extension.

The build script reads the original `manifest.json` file's `version` and outputs .zip files with the version number included. It also modifies the content based on the browser:
- For Firefox all icons used are switched to SVG icons
- for Chromium the extra `applications` object is removed

To run the script, run: `node build.js`.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/25.

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,0 +1,48 @@
+'use strict';
+
+const fs = require('fs');
+const extra = require('fs-extra');
+const zaf = require('zip-a-folder');
+
+const DEST = 'keepassxc-browser';
+const DEFAULT = 'manifest_default.json';
+const BROWSERS = {
+    'Firefox': 'manifest_firefox.json',
+    'Chromium': 'manifest_chromium.json',
+};
+
+function adjustManifest(manifest) {
+    const manifestFile = fs.readFileSync(DEFAULT, 'utf8');
+    const data = JSON.parse(manifestFile);
+    const browser = manifest.substring(manifest.indexOf('_') + 1, manifest.indexOf('.'));
+
+    if (manifest.includes('firefox')) {
+        for (const elem in data['icons']) {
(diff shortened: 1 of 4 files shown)


thetric:
Why does it use Python instead of JavaScript and Node.js? I don't see the advantage of adding another scripting language to the project

varjolintu:
You're right. Tests and eslint are using Node.js already so maybe it would be more general approach to do this in JS.

build.js
```diff
@@ -0,0 +1,45 @@
...
+
+fs.copyFileSync(`${DEST}/manifest.json`, `./${DEFAULT}`);
+
+for (const browser in BROWSERS) {
+    console.log(`KeePassXC-Browser: Creating extension package for ${browser}`);
+    const fileName = adjustManifest(BROWSERS[browser]);
+    fs.copyFileSync(BROWSERS[browser], `${DEST}/manifest.json`);
+    exec(`rm -rf ${fileName}`);
+    exec(`zip -r -FS ${fileName} ${DEST}/* > /dev/null`);
+    exec(`rm -rf ${BROWSERS[browser]}`);
```
thetric: This code is Linux/Mac-specific and doesn't work on Windows. To make it cross platform I suggest to replace `rm  -rf` with [fs-extra](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/remove-sync.md) and `zip` with [archiver](https://www.npmjs.com/package/archiver)

varjolintu: Good point.

phoerious:
Okay, although I merged it now, this needs fixing. The Firefox package does not contain SVGs. It's handy to have an automated build script, though.

PR #407, opened 2019-02-06
Add Russian translation

shvchk:


phoerious:
Our translations are managed on Transifex. Please add your translation there.

Issue #417, opened 2019-02-20
Deploy translations from Transifex to the plugin


Code2Mirabeau:
hello, 

I saw on [transifex.com](https://www.transifex.com/keepassxc/keepassxc-browser/) that the extension was translated into French (100%), however, following its installation (chrome) I see that this is not the case (always in English). 

I manually add the file "fr / messages.json" in the chrome folder, and now I have the extension in French
there is just a problem of offset, related to the length of the text which suddenly switches the red icon
[See Here](http://image.noelshack.com/fichiers/2019/08/3/1550703466-screenshot-1.png)

The second thing, when viewing the json file "messages.json" (from Transifex) I see that no tag "Description" was translated, but only the tag "message" example:

`"errorMessageUnknown": { 
           "message": "Unknown error.", 
           "description": "Unknown error." 
}`

In short, if there is need for a helping hand for the translation into French, I am available for that.

Alain

droidmonkey:
This was because of a failure to actually download and deploy the translations when the plugin was last updated. All contributions are made on Transifex.

Code2Mirabeau:
For information, I changed the text of the translation to avoid graphic shift ....

Old One: [ScreenShot](http://www.noelshack.com/2019-08-3-1550703466-screenshot-1.png)
New One: [ScreenShot](http://image.noelshack.com/fichiers/2019/08/4/1550704134-screenshot-2.png)

PR #457, opened 2019-03-28
Improve the build script

varjolintu:
- Translations are now updated from Transifex before packaging (uses the same tx client as KeePassXC)
- Added the missing SVG icons for Firefox build

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/417.


jaller94:
⚠️ This PR also raises the version number to 1.3.3.

varjolintu:
From 1.3.3 to 1.4.0 yes, because that's the next build.

Issue #165, opened 2018-05-18
Unwanted Data-cip-id in CKEditor


JeroenED:
When the extension is running and CKEditor is editing a page with input fields in HTML-view, several attributes are added to input fields. Additionally, when loading a page that already has these attributes but KeepassXC is unable to find them in the database, the form fields are losing their type and name attributes.

## Expected Behavior
Fields should not be changed when CKEditor (or another WYSIWYG editor) is loaded.

## Current Behavior
* Additional attributes (data-cip-id) are added to input tags containing a unique identifier
* When loading a page containing saved data-cip-id attributes that were not saved in the database name and type attributes are removed.

## Possible Solution
* Exclude ckeditor classes from running the browser-addon

## Steps to Reproduce (for bugs)

1. Make keepassbrowser running
2. Open CKEditor page with input-fields in html-mode
3. (data-cip-id is added)
4. Save the page
5. Reopen page in CKeditor
6. (name and type attributes are gone)

## Debug info

KeePassXC-Browser Version: 1.0.1
KeePassXC Version: 2.3.3 
Operating system: Win
Browser: Firefox
Proxy used: NO

varjolintu:
Maybe the correct approach for this is extending Ignored Sites feature. A checkbox could be added which prevents loading the content script(s) for a certain pages.

JeroenED:
I do believe this could be a more feasable solution. But it would require support for wildcards. In my case the url is `https://test.jeroened.be/admin/pages/edit/2` where `2` is the ID of the page.

Is there a workaround possible for now that I can use? Maybe by using a additional add-on?

varjolintu:
For now the only workaroung would be to modify the extension's `manifest.json` with `exclude_matches`. This of course means downloading the source code and loading extension manually.

An example:
```
"content_scripts": [
    {
      "exclude_matches": ["*://developer.mozilla.org/*"],
      "matches": ["*://*.mozilla.org/*"],
      "js": ["jquery.js", "borderify.js"]
    }
  ]
```

varjolintu:
With https://github.com/keepassxreboot/keepassxc-browser/pull/176 you can add the site to a ignore list and with `Full ignore` option enabled no input fields are modified or handled at all.

varjolintu:
Good news. I added wildcard support to the Ignored Sites PR. It follows the [match patterns](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Match_patterns).

[...]

PR #458, opened 2019-03-30
Fixed ignored sites in firefox

JeroenED:
Because I was freaking out on issue #165 I tried find a solution. Appeared to be the most simple fix ever.

Here is the PR for it :)

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/165.

### DIFF
--- a/keepassxc-browser/keepassxc-browser.js
+++ b/keepassxc-browser/keepassxc-browser.js
@@ -1559,7 +1559,7 @@ cip.initCredentialFields = function(forceCall) {
         cip.initializeSitePreferences();
         if (cip.settings.sitePreferences) {
             for (const site of cip.settings.sitePreferences) {
-                if (site.url === document.location.href || siteMatch(site.url, window.top.location.href)) {
+                if (site.url === window.top.location.href || siteMatch(site.url, window.top.location.href)) {
                     if (site.ignore === IGNORE_FULL) {
                         return;
                     }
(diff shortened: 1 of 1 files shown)


keepassxc-browser/keepassxc-browser.js
```diff
@@ -1559,7 +1559,7 @@ cip.initCredentialFields = function(forceCall) {
         cip.initializeSitePreferences();
         if (cip.settings.sitePreferences) {
             for (const site of cip.settings.sitePreferences) {
-                if (site.url === document.location.href || siteMatch(site.url, document.location.href)) {
+                if (site.url === document.location.href || siteMatch(site.url, window.top.location.href)) {
```
droidmonkey: The only time ```document.location.href``` and ```window.top.location.href``` would be different is in a frames scenario. In this case, is this the desired behavior? You would also want to change the other half of the if statement if it is.

JeroenED: Yes, this is indeed the required behavior. #165 handles about wysiwyg editors who add additional data-cip-id attributes to input fields. They use iframes which firefox sees as a frame with a different document.location (which is actually a empty string) Chrome/chromium has different behavior and was already solved. This fix does not change anything to chromium.

droidmonkey: Great!

varjolintu:
Excellent! Thank you.

JeroenED:
Can you give some clear instructions on how repack the XPI? I know the fix, I can redownload the xpi, I can repack the xpi, but I can't seem to install the xpi. I get a corrupted package error.

(Yes, it's a very annoying issue to me 😉)

varjolintu:
Just zip it. Or you can load the extension from a path using about:debugging (load the manifest.json file through it).

PR #504, opened 2019-04-24
Fix autocomplete tab

### LINKED ISSUE #497: Dropdown autocompletion with tab breaks own auto-type feature

1. Following use case: You have **one entry** for a website saved and keepassxc-browser shows you a dropdown.
  Now, for security reasons or so, you have a different database (or even the same one) with a different entry that is (_deliberately_) not suggested there.
2. So you now want to use the auto-type feature of KeePassXC to type that one in – _instead_ of the one that is suggested to you.
3. You focus the username input.
4. You go to KeePassXC and do <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd>.

## Expected Behavior

Auto-type works and enters different entry.

## Current Behavior

(Using the default auto-type sequence, obviously.)

What auto-type actually does:
* type in username
* <kbd>tab</kbd> (to jump to password field)
* keepassxc-browser actually recognizes <kbd>tab</kbd> as selecting the suggested entry -> enters username and password; in my test also selected whole password entry, like this on Twitter:
  image
* so now auto-type goes on entering the password

**The result:** It uses the username from the dropdown, and password from auto-type. Login thus (~likely~ hopefully :wink:, if you don't re-use the passwords) fails.

## Possible Solution

Remove selection by tab?
Or somehow conditionally disable it, so it at least does not trigger for own auto-type?

## Debug info

Operating system: Linux
Browser: Firefox

varjolintu:
Fixes the autocomplete with <kbd>Tab</kbd> so it does not fill any credentials if the username is not in the list. Previously this prevented filling another credentials via Auto-Type.

Fixes #497.

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -133,30 +133,41 @@ kpxcAutocomplete.create = function(input, showListInstantly = false, autoSubmit 
             return;
         }
 
-        const item = getAllItems();
+        const items = getAllItems();
         if (e.key === 'ArrowDown') {
-            // If the list is now shown, show it
-            if (item.length === 0) {
+            // If the list is not visible, show it
+            if (items.length === 0) {
                 _index = -1;
                 showList(input);
             } else {
                 // Activate next item
                 ++_index;
-                activateItem(item);
+                activateItem(items);
             }
         } else if (e.key === 'ArrowUp') {
             --_index;
-            activateItem(item);
-        } else if (e.key === 'Enter' || e.key === 'Tab') {
+            activateItem(items);
+        } else if (e.key === 'Enter') {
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/autocomplete.js
```diff
@@ -133,30 +133,41 @@ kpxcAutocomplete.create = function(input, showListInstantly = false, autoSubmit
             return;
         }
 
-        const item = getAllItems();
+        const items = getAllItems();
         if (e.key === 'ArrowDown') {
             // If the list is now shown, show it
```
droidmonkey: Minor: should be "not shown"

varjolintu: My tired eyes gives a thank you.

PR #512, opened 2019-04-27
Fix auto-submit from keyboard shortcut

### LINKED ISSUE #511: Auto-submit breaks with multiple logins
I have enabled the setting "Auto-submit login forms".  It works well provided that there is only one matching login.

## Expected Behavior
If there are multiple matching logins, then do not auto-submit until after I choose one.

## Current Behavior
If there are multiple matching logins, then the shows a prompt but then immediately auto-submits with blank fields which causes an error.

1. Save multiple logins for a site.
2. Enable setting "Auto-submit login forms".
3. Visit the site login page and press AS+U

## Debug info

KeePassXC - 2.4.1
KeePassXC-Browser - 1.4.3
Operating system: Win
Browser: Firefox

varjolintu:
When using keyboard shortcut for credential filling and there's multiple credentials, auto-submit triggered before any credentials were selected via autocomplete.

Also modifies autocomplete.js to allow showing the list without creating a duplicate autocomplete for the input.

Fixes #511.


keepassxc-browser/content/autocomplete.js
```diff
@@ -1,10 +1,21 @@
...
+kpxcAutocomplete.autoSubmit = false;
 kpxcAutocomplete.elements = [];
+kpxcAutocomplete.started = false;
+kpxcAutocomplete.index = -1;
+kpxcAutocomplete.input = undefined;
 
 kpxcAutocomplete.create = function(input, showListInstantly = false, autoSubmit = false) {
-    let _index = -1;
+    //let _index = -1;
+    console.log('start');
```
droidmonkey: Remove comment and temporary debug message

keepassxc-browser/content/autocomplete.js
```diff
@@ -1,10 +1,21 @@
...
+
+    kpxcAutocomplete.autoSubmit = autoSubmit;
+    kpxcAutocomplete.input = input;
+    if (!kpxcAutocomplete.started) {
+        kpxcAutocomplete.started = true;
```
droidmonkey: No need to check if it is false if you are going to set it to true no matter what

varjolintu:
All changes done.

PR #520, opened 2019-05-04
Allow filling username only from keyboard shortcut

varjolintu:
Allow filling username using a keyboard shortcut when username field is the only visible one in the page.

For the Google page the DOMException was not handled correctly which caused the content script to fail.

Please note that if the page has scripts that include unhandled exceptions, keyboard shortcuts are only handled after those are thrown. This may require multiple uses of the shortcut, or wait till the page has loaded completely. Also, the input field must have focus.

Examples:
- https://accounts.google.com
- https://id.atlassian.com/login

Fixes solution 2. in https://github.com/keepassxreboot/keepassxc-browser/pull/520.


AdamPS:
Thanks for a quick response @varjolintu - I am looking forward to trying this in a new release.

PR #537, opened 2019-05-09
Support for displaying expired credentials

varjolintu:
Adds support for "expired" field in the protocol. If an entry has that property set, an extra string "[Expired]" is shown at the end of the entry title in the popup and autocomplete menu to inform the user.

KeePassXC side PR: https://github.com/keepassxreboot/keepassxc/pull/3134.

Screenshot:
Screen Shot 2019-05-09 at 8 44 24

### DIFF
--- a/keepassxc-protocol.md
+++ b/keepassxc-protocol.md
@@ -204,7 +204,8 @@ Response message data (success, decrypted):
     {
         "login": "user2",
         "name": "user2",
-        "password": "passwd2"
+        "password": "passwd2",
+        "expired": "true"
     }],
     "nonce": "tZvLrBzkQ9GxXq9PvKJj4iAnfPT0VZ3Q",
     "success": "true",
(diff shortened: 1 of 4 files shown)


keepassxc-protocol.md
```diff
@@ -204,7 +204,8 @@ Response message data (success, decrypted):
     {
         "login": "user2",
         "name": "user2",
-        "password": "passwd2"
+        "password": "passwd2",
+        "expired": "true"
```
droidmonkey: Shouldn't this be "bool" since it can be true or false.

varjolintu: Maybe, but I wanted to maintain the compatibility with other values. The success property also sets this as a string.

droidmonkey: Yah after viewing the rest of the protocol document, makes sense.

PR #549, opened 2019-05-24
Fix reconnect issues

### LINKED ISSUE #523: KeePassXC-Browser will only work if KeePass was launched beforehand
If KeePass 2 wasn't launched prior to launching Browser the extension doesn't work. I have to disable the extension and then enable it again to make it work.

1. Launch Browser
2. Launch KeePass 2
3. KeePassXC-Browser is gray (Reload, Can't decrypt message)
4. Disable/Enable the extension to get it to work

KeePass - 2.41
KeePassXC-Browser - 1.4.3
Operating system: Win 7
Browser: Opera

### LINKED ISSUE #544: Reconnect issues
## Expected Behavior

The extension should reconnect to KeePassXC in all situations.

## Current Behavior

Auto-reconnect only works properly if KeePassXC is started before the browser.

## Possible Solution

This is a tricky one. Fixing the actual behaviour here is easy, but when I inspected the case some more I noticed a few problems:
- Reconnect causes memory leaks (that I could not remove totally, thanks JavaScript).
- There's no way to detect a situation where keepassxc-proxy exits right away (I tried it to send messages to stdout when this happens, the extension didn't get any of those), meaning it cannot connect to KeePassXC. It's uncertain if the proxy exits "too quick" so the `onDisconnect` listener doesn't have time to attach to the event.

The possible fix would be using the solutions above, but for now it's a dead end. One possible (and the best) solution is to modify `keepassxc-proxy` that it can handle the reconnect instead of the extension. These attempts have also been unsuccessful and I need more time to inspect this solution.

1. Start your browser with Auto-Reconnect enabled
2. See the mag.. nothing happens

## Debug info

KeePassXC - 2.4.2
KeePassXC-Browser - 1.4.4 (dev)

varjolintu:
Fixes the following reconnect issues:
- Reconnect didn't work if KeePassXC was started after the browser
- Clicking the popup was not reconnecting the extension properly
- Error message was shown if reconnected database was already open

FYI: When reconnect is active the extension consumes memory, but the garbage collector will clean it when the reconnect happens.

Fixes #523.
Fixes #544.

### DIFF
--- a/keepassxc-browser/background/event.js
+++ b/keepassxc-browser/background/event.js
@@ -152,16 +152,12 @@ kpxcEvent.onGetStatus = function(callback, tab, internalPoll = false, triggerUnl
     }
 };
 
-kpxcEvent.onReconnect = function(callback, tab) {
-    // Add a small timeout after reconnecting. Just to make sure. It's not pretty, I know :(
-    setTimeout(() => {
-        keepass.reconnect(callback, tab).then((configured) => {
-            browser.tabs.sendMessage(tab.id, {
-                action: 'redetect_fields'
-            });
-            kpxcEvent.showStatus(configured, tab, callback);
-        });
-    }, 500);
+kpxcEvent.onReconnect = async function(callback, tab) {
+    const configured = await keepass.reconnect(callback, tab);
+    browser.tabs.sendMessage(tab.id, {
+        action: 'redetect_fields'
+    });
+    kpxcEvent.showStatus(configured, tab, callback);
 };
 
 kpxcEvent.lockDatabase = function(callback, tab) {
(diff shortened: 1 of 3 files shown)


Bruceforce:
When you say "consumes memory" does that mean a user will run out of RAM if he doesn't open keepassxc and the reconnect can't happen during a long period of time?

varjolintu:
@Bruceforce The consumption is not that big. And the GC will trigger eventually even if the reconnect doesn't happen. Of course this is not an ideal solution, but it's the best so far.

In long term, we would like to handle any reconnects in keepassxc-proxy but it's still under dev.

droidmonkey:
Need to fast track proxy reconnect

PR #561, opened 2019-05-31
Remove popup reconnect

### LINKED ISSUE #562: 1.4.5 completely breaks on Windows
## Expected Behavior
1. The extension initializes `keepass.currentKeepassXC` as a blank string.
2. The extension enables automatic reconnects before connecting even once.
3. Eventually, a connect occurs because the extension realizes it's not connected.
4. Everyone is happy!

## Current Behavior
1. The extension initializes `keepass.currentKeepassXC` as a blank string.
2. The extension attempts to enable automatic reconnects before connecting even once.
3. Some legacy code, intending to block out old versions of KeePassXC on Windows, determines that the empty string is a lower version number than 2.3.4.
4. Therefore, automatic reconnects are never actually enabled, and the extension never connects.
5. Everyone is very sad.

## Introduced by Commit
https://github.com/keepassxreboot/keepassxc-browser/commit/9a5408e366c6422b557b0802704a4aa80664a447

## Possible Solution
Either remove the block-out code in `keepass.enableAutomaticReconnect`, or perform one manual connect before enabling auto-reconnect, like was done in earlier versions of the extension.
Or if you want to get really fancy, re-write the block-out code to work with the new reconnect logic. Though I can't tell if this is overkill, since I'm not sure why it's there in the first place.

## Steps to Reproduce
Try to use the extension version 1.4.5 on Windows.

## Debug info

KeePassXC - 2.4.1
KeePassXC-Browser - 1.4.5
Operating system: Win
Browser: Firefox

varjolintu:
- Removes reconnect from the popup (causes more problems than any good)
- Sets Auto-Reconnect to disabed as default
- Add a warning message to Settings page
- Fixes a connecting issue with KeePassXC versions earlier than 2.4.0

Fixes #562.

### DIFF
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -419,16 +419,6 @@ keepass.testAssociation = async function(callback, tab, enableTimeout = false, t
         page.tabs[tab.id].errorMessage = null;
     }
 
-    if (!keepass.isKeePassXCAvailable && !page.settings.automaticReconnect) {
-        try {
-            await keepass.reconnect();
-        } catch (err) {
-            keepass.handleError(tab, kpErrors.PUBLIC_KEY_NOT_FOUND);
-            callback(false);
-            return false;
-        }
-    }
-
     keepass.getDatabaseHash((dbHash) => {
         if (!dbHash) {
             callback(false);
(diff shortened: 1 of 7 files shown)


keepassxc-browser/background/keepass.js
```diff
@@ -1145,6 +1135,8 @@ keepass.enableAutomaticReconnect = function() {
             }
         }, 1000);
     }
+
+    return;
```
droidmonkey: This seems unnecessary

keepassxc-browser/background/page.js
```diff
@@ -11,7 +11,7 @@ const defaultSettings = {
     showNotifications: true,
     showLoginNotifications: true,
     saveDomainOnly: true,
-    automaticReconnect: true,
+    automaticReconnect: false,
```
droidmonkey: Recommend we rename the setting entirely so that people who had it enabled previously will not keep it enabled. Recommend ```autoReconnect```.

MaxXor:
Thanks, finally reconnect is working again! Imho it should be enabled by default.

varjolintu:
It's disabled by default because it can cause problems on some systems and launching multiple processes when the proxy is disabled.

PR #566, opened 2019-06-07
Improve password change detection

### LINKED ISSUE #334: New username/passwd detection fails in Firefox
KeepassXC Browser addon fails to detect new username password for some websites. Websites which it fails consistently : 
- https://www.kohls.com
- https://www.doordash.com/

## Expected Behavior
It should show a notification for the new username password which could be saved.

## Current Behavior
It doesn't show any notification for new username password.

## Possible Solution
Unclear, I could not debug this issue. Need some information about I can debug this addon.

1. Goto a website where it doesn't work for e.q. kohls.com
2. Login using a username password which is not stored in keepassxc
3. Note that the addons fails to show notification to store new credentials

## Context
Irritatingly I need to store this password manually for those cases. 

## Debug Info
KeePassXC-Browser Version: 1.3.0 
KeePassXC - Version 2.3.4
Revision: 6fe821c

Libraries:
- Qt 5.9.5
- libgcrypt 1.8.1

Operating system: Ubuntu 18.04.1 LTS
CPU architecture: x86_64
Kernel: linux 4.15.0-39-generic

Enabled extensions:
- Auto-Type
- Browser Integration
- Legacy Browser Integration (KeePassHTTP)
- SSH Agent
- YubiKey

### LINKED ISSUE #421: Password not updated when changed.
## Expected Behavior

When changing password on many websites, the extension offers to update password in KeePassXC. This should work.

## Current Behavior

On more than one occasion, the password wasn't actually updated. Sometimes I recovered by noticing it in time, and pasting it manually in password manager, but sometimes I noticed it too late, forcing me to use "reset password" functionality, which is less than optimal.

1. Register on hackaday.io and save password in KeePassXC.
2. Change password (on https://hackaday.io/settings?tab=password)
3. Extension asks to update password for empty URL and empty login. Clicking the update button does nothing.

## Possible Solution

I don't know why the URL was set to an empty string there, but it should not. The login is often not needed in "change password" flows, so it is often either missing or detected as original password, required to change it. A possible solution would be to offer the user to select entry to be updated (from the ones matching current URL). Related bug is that I have saved, as imported from other password manager, URL with some extra path (such as /login.php). The extension correctly detects it when trying to log in, but seemingly not when trying to update the password.

## Debug info

KeePassXC - 2.3.4
KeePassXC-Browser - 1.3.2
Operating system: Linux
Browser: Chromium

### LINKED ISSUE #515: KPXC browser doesn't detect new login credentials Smartsheet
## Expected Behavior
Plugin button/icon i/t top bar shows bubble with question to save new credentials or not.

## Current Behavior
Plugin button/icon i/t top bar doesn't do anything; stays green (sometimes turns grey). I had another site where new credentials were recognised correctly.

1. Go to https://app.smartsheet.com/
2. Log in with correct credentials

## Debug info

KeePassXC - 2.4.1
KeePassXC-Browser - 1.4.3
Operating system: Win32
Browser: Mozilla Firefox 66.0

varjolintu:
- Improves the form submit detection by adding a listener to the login button
- Improves the login button detection

Fixes #334.
Fixes #421.
Fixes #515.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -105,6 +105,11 @@ kpxcForm.init = function(form, credentialFields) {
         form.setAttribute('kpxcForm-initialized', true);
         kpxcForm.setInputFields(form, credentialFields);
         form.addEventListener('submit', kpxcForm.onSubmit);
+
+        const submitButton = kpxc.getSubmitButton(form);
+        if (submitButton !== undefined) {
+            submitButton.addEventListener('click', kpxcForm.onSubmit);
+        }
     }
 };
 
@@ -127,8 +132,9 @@ kpxcForm.setInputFields = function(form, credentialFields) {
 };
 
 kpxcForm.onSubmit = function() {
-    const usernameId = this.getAttribute('kpxcUsername');
-    const passwordId = this.getAttribute('kpxcPassword');
+    const form = this.nodeName === 'FORM' ? this : this.form;
+    const usernameId = form.getAttribute('kpxcUsername');
+    const passwordId = form.getAttribute('kpxcPassword');
 
     let usernameValue = '';
     let passwordValue = '';
@@ -137,7 +143,7 @@ kpxcForm.onSubmit = function() {
     const passwordField = _f(passwordId);
 
     if (usernameField) {
-        usernameValue = usernameField.value;
+        usernameValue = usernameField.value || usernameField.placeholder;
     }
     if (passwordField) {
         passwordValue = passwordField.value;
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -104,7 +104,14 @@ kpxcForm.init = function(form, credentialFields) {
     if (!form.getAttribute('kpxcForm-initialized') && (credentialFields.password || credentialFields.username)) {
         form.setAttribute('kpxcForm-initialized', true);
         kpxcForm.setInputFields(form, credentialFields);
+        form.removeEventListener('submit', kpxcForm.onSubmit);
         form.addEventListener('submit', kpxcForm.onSubmit);
+
+        const submitButton = kpxc.getSubmitButton(form);
+        if (submitButton !== undefined) {
+            submitButton.removeEventListener('click', kpxcForm.onSubmit);
+            submitButton.addEventListener('click', kpxcForm.onSubmit);
```
droidmonkey: Not sure we need these remove/add pairs. See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Multiple_identical_event_listeners

varjolintu: Thanks for the info. Then those pairs are really not needed.

droidmonkey: Yah just the "add" part is needed. It'll replace whatever is already attached with the same function signature.

PR #567, opened 2019-06-13
polished links a little bit to bring them in line

DJCrashdummy:


droidmonkey:
That sentence should actually be deleted. We don't have that code in the baseline anymore.

PR #578, opened 2019-07-09
Focus to input field after keyboard fill

### LINKED ISSUE #571: Keyboard focus not set after autofill via keyboard shortcut, blocking subsequent keyboard shortcuts
I love keepassxc-browser, but this UI glitch trips me up.  I'm a heavy keyboard-shortcuts user, and generally work without the mouse.

## Expected Behavior

After hitting ctrl-shift-u, my username and password are autofilled.  After the autofill, I expect to hit [return] and have the login form submitted (i.e. i'm logged in). 
## Current Behavior

I cannot hit [return] to log in, because the autofilled fields don't have the keyboard focus.

## Possible Solution
I think the keepassxc-browser menu modal dialog should be dismissed when autofill is done successfully, and the autofilled field should reliably grab the keyboard focus (as if the user had typed by hand).

1. Make sure keepassxc is not running.
1. Go to a login web page that can be autofilled.
1. Launch keepassxc and open the database.
1. Click on the keepassxc-browser menu icon, click the [reload] button.
1. Type shift-ctrl-u to autofill, then look at the page to check that autofill was done correctly
1. Hit [return] to log in.

Oops, the log in doesn't happen because the keepassxc-browser modal dialog is still up and still has the keyboard focus.

## Workaround

After hitting autofill, find the mouse, grab it, and click on the password field of the web page. Then you can hit return or click the submit button.  So much for keyboard shortcuts.

## Debug info

KeePassXC - 2.4.3
KeePassXC-Browser - 1.4.6
Operating system: Mac
Browser: Firefox 67.0.4

varjolintu:
After a keyboard fill is made, focus to the input field (to password field if it's filled alone) to enable a form submit.

Fixes #571.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -1110,7 +1110,13 @@ kpxc.fillInFromActiveElement = function(suppressWarnings, passOnly = false) {
     const el = document.activeElement;
     if (el.tagName.toLowerCase() !== 'input') {
         if (kpxcFields.combinations.length > 0) {
-            kpxc.fillInCredentials(kpxcFields.combinations[0], false, suppressWarnings);
+            kpxc.fillInCredentials(kpxcFields.combinations[0], passOnly, suppressWarnings);
+
+            // Focus to the input field
+            const field = _f(passOnly ? kpxcFields.combinations[0].password : kpxcFields.combinations[0].username);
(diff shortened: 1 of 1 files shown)


varjolintu:
Yes. The boolean value was not passed correctly and it could cause username filling when only password was requested.

brad2014:
Thank you!

PR #612, opened 2019-09-27
Update messages.json

Stephan-P:
Manual transfer from TransiFex file to Github repository


varjolintu:
The translation files are updated from Transifex when making a new build. We don't update them via GitHub.

PR #614, opened 2019-09-27
Fixes for the latest major release

### LINKED ISSUE #605: Unable to disable 'fill credentials from keepassxc' icon
Is it possible to disable this icon?

https://imgur.com/a/r1dJzb5

It clutters the login field and provides no new functionality. I  prefer the drop down menu from which the record can be chosen.

### LINKED ISSUE #607: New release seems to break some pages
## Expected Behavior
This is[ the page where I first noticed this problem](https://indico.cern.ch/event/763185/).
image

## Current Behavior
Stuff which rendered OK before looks just broken when the KeePassXC extension is enabled.

image

## Possible Explanation
I imagine that the plugin is injecting some HTML or CSS into the webpage, which breaks things.

The link provided above should be enough...

## Debug info
KeePassXC - 2.4.3
KeePassXC-Browser - 1.5.0
Operating system: Mac
Browser: Firefox

varjolintu:
Fixes a few annoyances from the major 1.5.0 update:
- Username field icon is now optional (enabled by default)
- Fix CSS separator
- Scale the username field icon dynamically
- Hide the username field icon when input field is too small in width or height

Fixes #605.
Fixes #607.


keepassxc-browser/content/username-field.js
```diff
@@ -54,7 +54,16 @@ const createIcon = function(target, databaseClosed) {
 
     const field = target;
     const className = getIconClassName(databaseClosed);
-    const size = (field.offsetHeight > 28) ? 24 : 16;
+
+    // Size the icon dynamically, but not greater than 24 or smaller than 14
+    const size = (field.offsetHeight > 28) ? 24 : Math.max((field.offsetHeight - 4), 14);
```
droidmonkey: This could be simplified to: `Math.max(Math.min(24, field.offsetHeight - 4), 14)`

keepassxc-browser/background/page.js
```diff
@@ -10,6 +10,7 @@ const defaultSettings = {
     autoRetrieveCredentials: true,
     showNotifications: true,
     showLoginNotifications: true,
+    showUsernameIcons: true,
```
droidmonkey: I'm on the fence for this variable name, I feel like it should be `showLoginFormIcon` (not plural).

droidmonkey:
Should we wait to figure out why we are having DB connection issues? Certainly wasn't just one person.

varjolintu:
@droidmonkey Yes. Let's wait a bit. I still don't know why it happens for some, and not for me. I've tried it with macOS and Linux VM with multiple different browsers.

droidmonkey:
I too cannot replicate. Even moving from the develop branch to the official release everything still worked perfectly.

varjolintu:
@droidmonkey I made the fixes you suggested. Thanks for those. I think we could merge this one and try to solve the database problem after this one. Reproducing it seems to take some more time.

PR #617, opened 2019-09-30
Fix handling multiple icons

### LINKED ISSUE #608: Bug if keepassxc icons are showing for non-visible fields?
I created bug https://github.com/huashengdun/webssh/issues/90 in a very good web-based ssh client. I am shown a login page, and then I am using my ssh terminal (see screenshot in that bug).

The problem is that the terminal is covered in keepassxc icons.

Should the icons still be visible in this case? If not, what is the expected way of hiding them in other applications?

(It seems to me that they should not be visible since the field is not visible)

varjolintu:
Previously, handling multiple icons on a site was not working correctly. The icon Object(s) always referred to the last icon created, so hiding/showing them would only affect to the last icon.

Improvements:
- All icons are now created dynamically instead of a single Object.
- Added UsernameFieldIcon and PasswordIcon classes.
- Changed `kpxcPassword` to `kpxcPasswordDialog`, removing the icon handling from that Object
- Moved `IntersectionObserver` handling to `kpxcUI` and new `Icon` class instead of copying the same code.

To test this, you can create the following local HTML file:
```HTML
<!DOCTYPE html>
<html>
<head>
</head>
<body>
<button id="nav1">Hide</button>
<button id="nav2">Show</button>
<section id="sec1" style="display:none">
<h1>Hidden</h1>
</section>
<section id="sec2">
<h1>Visible</h1>
<input type="password" />
<input type="password" />
<input type="password" />
</section>
<script>
function fun(e) {
    for (let section of document.getElementsByTagName("section")) {
        if ("nav" + section.id.substring(3) == e.target.id) {
            section.style.display = "block";
        } else {
            section.style.display = "none";
        }
    }
}

document.getElementById("nav1").addEventListener("click", fun);
document.getElementById("nav2").addEventListener("click", fun);
</script>
</body>
</html>
```

Without the fix, only the last icon will hide, and only its state is affected when a database is opened/closed.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -60,7 +60,7 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
             kpxc.settings = response;
             kpxc.initCredentialFields(true);
         } else if (req.action === 'show_password_generator') {
-            kpxcPassword.trigger();
+            kpxcPasswordDialog.trigger();
         } else if (req.action === 'add_username_only_option') {
             kpxc.addToSitePreferences();
         }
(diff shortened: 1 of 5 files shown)


droidmonkey:
Very nice

PR #620, opened 2019-10-01
Fix update button from Credential Banner

### LINKED ISSUE #619: Password change on Github -> "Update is grayed out"
If I change my password on Github, then login with the new password, the top bar shows New | Update | Dismiss. But if I try and click Update it grays out before I have clicked it.

If I am very very quick I can click the Update button, but it does not update the database.

varjolintu:
Fixes the Update button disabling from Credential Banner.

Previously, after a page submit, even if `oldCredentials` had the credentials `kpxc.credentials` were used instead because it existed, but was an empty Array. Now `oldCredentials` are always used instead if it's defined.

Fixes #619.


varjolintu:
It doesn't.

PR #625, opened 2019-10-02
Add TOTP icon filling and improve icon handling

varjolintu:
This is an extension to #617.

New features:
- Adds a new OTP icon for filling
- Adjusts icon saturation based on database lock status (password and OTP icons are grey)

When DB is locked:
closed

When open:
open

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -167,6 +167,14 @@
         "message": "Username field icon",
         "description": "Username field icon text."
     },
+    "totpFieldText": {
+        "message": "Fill TOTP from KeePassXC",
+        "description": "OTP field icon hover text."
+    },
+    "totpFieldIcon": {
+        "message": "TOTP field icon",
+        "description": "OTP field icon text."
+    },
     "defineDismiss": {
         "message": "Dismiss",
         "description": "Dismiss button text when choosing custom login fields."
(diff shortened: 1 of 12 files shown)


droidmonkey:
We are introducing a key icon again?

varjolintu:
Nope. It's just enabled for the screenshot. Still disabled by default.

droidmonkey:
Some findings for various popular websites that use different id's for their TOTP field:

* Twitter: `challenge_response`
* Google: `totpPin`
* Turbotax: `ius-mfa-soft-token`
* Amazon AWS: `mfaCode`

I found that sometimes the OTP field was misidentified as a username field when you have the site setup as username-only detection. There should be some workflow tracking to not propose another username fill AFTER the password is filled. Even a simple flag of "hasPasswordBeenFilled" would suffice to not re-propose a username fill.

varjolintu:
@droidmonkey Good findings. It's clear that we cannot support all possible ID's there are, but at least fields containing `challenge`, `mfa`, `2fa` or `token` could be added to the list. And you are correct about the another username fill happening. I will make some changes to this PR. I don't want to rush it.

droidmonkey:
Yah good idea, that that should cover the majority of cases.

varjolintu:
Changes made. I tested the TOTP field using Username-only Detection enabled for GMail and Mastodon. Both worked fine.

grenzor:
Been checking some websites I visit and noticed OTP field IDs that contain `code` and `auth` more than once (with differences in uppercase/lowercase, not sure if that matters here). Could they be added to the list? Seems to me like they could be somewhat common.

varjolintu:
@grenzor Those could be added also. Just noticed the comparison is currently not made with lowercase. Gotta fix that too.

varjolintu:
Plus, I will make showing the icon optional (enabled by default), just like the other icons.

grenzor:
`2fa` removed and `otp` listed twice. Mistake?

varjolintu:
Ah, bad copy paste. Fixed.

PR #628, opened 2019-10-04
Improvement/Feature for HttpAuth with multiple entries -> Specify exact entry with KPH: HttpAuth Flag

brainfoolong:
Hi. I use KeepassXC a lot and ever since there is one thing bothering me every day.
I have a some sites, that have a basic authentication in the first line and also a second/third/fourth later on on the website. The part with websites work flawless, Keepass can provide all credentials in the login forms.

With HttpAuth on the other hand, it's cumbersome. If you have multiple matching entries when dealing with HttpAuth you get aa annoying popup (if activated) and notification (if activated) as well. The popup where you can choose the credentials does nothing for HttpAuth, the browser stall until you goto the KP Icon and select the right entry again. The notification just say you have to click the KP icon. All in all, you have to click multiple times in different locations for HttpAuth.

So, i decided to improve the browser extension to be able to use one specific entry, out of many, for your HttpAuth and applying that automatically if possible.

For that we have the new additional attribute `KPH: HttpAuth`, which you can apply to your entry in the desktop app under the "Advanced" Tab. When it is set to `1` than the algorithm takes the first where this condition is met and authenticate you on the site with this.

Result: You need less clicks for the same result.

Problems that still exist with this because it does require changes in the desktop program of KeepassXC, which i can't do

* The popup to choose for credentials still pops up, even if KP can you login without further user actions. This seems to be hardcoded in the desktop app. Someone have to integrate the logic for the new `KPH: HttpAuth` into that, to prevent the popup if there is a dedicated HttpAuth entry.
* If you have different HttpAuth for one site, you still need to manually choose the correct entry as before

The really best end solution would be, which is not part of my PR
* Eliminate the credential pick popup for HttpAuth, as it does nothing and stall the browser
* Provide the user a popup that works when having multiple HttpAuth entries, or just open the options from the KP icon automatically
* show tips how to use this KPH new flag

### DIFF
--- a/keepassxc-browser/background/httpauth.js
+++ b/keepassxc-browser/background/httpauth.js
@@ -73,16 +73,37 @@ httpAuth.processPendingCallbacks = async function(details, resolve, reject) {
 };
 
 httpAuth.loginOrShowCredentials = function(logins, details, resolve, reject) {
-    // At least one login found --> use first to login
+    // At least one login found
     if (logins.length > 0 && page.settings.autoFillAndSend) {
-        if (logins.length === 1) {
+        let loginToUse = null
+        if(logins.length === 1){
+            // if we only have one login, use that
+            loginToUse = logins[0]
+        }else{
+            // if we have more than one than check for specific kph flag to use that as http auth
+            loginLoop: for (let i = 0; i < logins.length; i++) {
+                let login = logins[i]
+                if(login.stringFields){
+                    for (let j = 0; j < login.stringFields.length; j++) {
+                        let stringField = login.stringFields[j]
+                        if(typeof stringField['KPH: HttpAuth'] !== 'undefined' && stringField['KPH: HttpAuth'] === "1"){
+                            loginToUse = login
+                            break loginLoop
+                        }
+                    }
+                }
+            }
+        }
+        if (loginToUse !== null) {
+            // if we have a dedicated login, use it
             resolve({
                 authCredentials: {
-                    username: logins[0].login,
-                    password: logins[0].password
+                    username: loginToUse.login,
+                    password: loginToUse.password
                 }
             });
         } else {
+            // if we have no dedicated login, show notes and popups
             if (page.settings.showNotifications) {
                 showNotification(tr('multipleCredentialsDetected'));
             }


varjolintu:
Thanks for the PR and the idea!

> The really best end solution would be, which is not part of my PR
> 
>     * Eliminate the credential pick popup for HttpAuth, as it does nothing and stall the browser
> 
>     * Provide the user a popup that works when having multiple HttpAuth entries, or just open the options from the KP icon automatically
> 
>     * show tips how to use this KPH new flag

This is where it gets problematic. If there are multiple credentials present, the only way to select the correct credentials is to use the popup. Then HTTP Auth triggers in a browser, there's no DOM available so you cannot write any credential choosing dialog to the screen. Also, the popup cannot be opened automatically because all browser API's does not support it.

So the best solution would be that this attribute would be returned automatically by KeePassXC, as you suggested.

brainfoolong:
Ok, i understand. With that said, will this request probably merged or does you pass that to the desktop devs.

When passed, than instead of a String Field a dedicated checkbox to say "It's a HttpAuth Credential" would be even better.

I've choosen a string field because there are already some special flags, like other input fields, you can define with this additional attributes, that are only handled by the browser extension.

varjolintu:
FYI, I'm also in the desktop dev team, making most of the browser extension related changes. Instead of an attribute, the option could be added to this new feature: https://github.com/keepassxreboot/keepassxc/pull/3444

brainfoolong:
Yes, this would be cool to have a checkbox for this HttpAuth PR in your new browser extension settings from keepass dekstop. Than this PR would be obsolete.

varjolintu:
I'm trying to do necessary fixes for KeePassXC 2.5.1.

brainfoolong:
Looking forward for this feature. The new 2 flags in KPXC are cool but i miss this one :)

keepassxc-browser/background/httpauth.js
```diff
@@ -73,16 +73,37 @@ httpAuth.processPendingCallbacks = async function(details, resolve, reject) {
 };
 
 httpAuth.loginOrShowCredentials = function(logins, details, resolve, reject) {
-    // At least one login found --> use first to login
+    // At least one login found
     if (logins.length > 0 && page.settings.autoFillAndSend) {
-        if (logins.length === 1) {
+        let loginToUse = null
+        if(logins.length === 1){
```
hardfalcon: Please don't do this. Just because there is only a single entry iny my KeepassXC database for a given site, this doesn't mean that I'm fine with the browser extension sending those credentials automatically to the server without asking me first.

Please *only* send credentials for which the HttpAuth flag has been set, regardless of how many entries there are for the same site.

varjolintu:
Closing this as the solution will be implemented to KeePassXC side.

PR #630, opened 2019-10-08
Fix connection issues

### LINKED ISSUE #615: keepassxc-browser loses connection with keepassxc when browser restarted
I have had Keepassxc-browser working with keepassxc in Mozilla Firefox and an openSuse Leap Linux system for a few weeks. After a plugin upgrade, upon closing and reopening the web browser (without closing keepassxc), connection between the browser plugin and Keepassxc is lost, and "cannot decrypt message" appears on the plugin icon window. Repeatedly clicking the "reopen database" button has no effect.  

## Expected Behavior

keepassxc.browser should be able to reconnect with keepassxc after browser restart.

## Current Behavior

As described above, no reconnection is possible without further steps.

## Possible Solution

The problem seems to be in the keepassxc-proxy package, or the interaction of keepassxc-browser with it. 

I have checked (ps -ef | grep keepass) that keepassxc-proxy remains alive when I close and reopen the browser. However, no reconnection is possible without taking further steps. Two things work for me:

- Totally close and reopen keepassxc. I believe this works because it also restarts the keepassxc-proxy process.

- kill and manually restart keepassxc-proxy in a terminal:
 killall keepassxc-proxy
keepassxc-proxy &

After this, the connection between keepasxc and the browser plugin can be restored by just clicking the plugin icon in the browser and then "Reconnect". In fact, as a workaround for the time being I have just created a shortcut to a script in my desktop that kills and restarts keepassxc-proxy with a single click.  

The "Automatically reconnect to KeePassXC" option is DISABLED in the plugin, btw.

1. Go to the login site of a web page already known by keepassxc and keepasxc-browser.
2. With the keepassxc database open, verify that keepassxc-browser can fill the login/password fields.
3. Completely close Firefox, while keeping keepassxc open. 
4. Reopen firefox and go to the same page: keepassxc-browser fails to fill the login/password fields, and "cannot decrypt message" appears in the plugin window.

## Debug info
KeePassXC - 2.4.3
KeePassXC-Browser - 1.5.1
Operating system: Linux, Opensuse Leap 15.0
Browser: Mozilla Firefox 60.8.0 esr 64 bit.

I am submitting this report under "keepassxc-browser" because, although keepassxc-proxy belongs to the keepassxc package, this behavior only started yesterday after upgrading the browser plugin to 1.5.1. Still, I should mention that I cannot rule out other changes in my system being responsible for this. Specifically, I was playing with other versions of keepassxc (compiled and installed keepassxc-feature-multiple_urls and played with it for a few minutes), but I believe I cleaned the system afterwards, and reinstalled the "official" 2.4.3 version from openSuse repositories. In any case, I have verified that the current keepassxc-proxy binary is the one provided by the official openSuse package.

varjolintu:
This PR aims to fix the recent connect issues, and it contains the following changes:
- `keepass.sendNativeMessage()` now allows to set the timeout value
- First connection timeout is increased to 5 seconds
- `keepass.reconnect()` now returns correctly if error situation happens
- A bonus: the ugly throbber.gif is replaced with a loader CSS in the extension popup

Fixes #615 

Probably also fixes https://github.com/keepassxreboot/keepassxc/issues/3598.


derula:
Is this the same as "test build 2" over in #615? Because I tried that build and it didn't fix the connection issues I'm having rn. But it's possible that I just "did it wrong."

varjolintu:
@derula Yes this is the same one. I suggest you try it with https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/.

varjolintu:
Tested this with both Firefox (using `web-ext`) and Chromium. Works just fine.

But because I've been unable to reproduce the original issue, any other testers will be needed.

apeyrat:
Having quite often errors with "can't decrypt message" in keepassxc-browser, I'm trying this commit since yesterday, it works and I no longer have the error.

droidmonkey:
Nice work!!

PR #632, opened 2019-10-11
Fix await with TOTP fill, code cleaning

### LINKED ISSUE #569: github totp hotkey not work
## github totp not work, but dropbox is normal.

image

varjolintu:
- Fix a missing `await`
- Fix errors with `define.js`
- General code cleaning

Fixes #569.

### DIFF
--- a/keepassxc-browser/content/define.js
+++ b/keepassxc-browser/content/define.js
@@ -274,11 +274,7 @@ kpxcDefine.prepareStep3 = function() {
     $('#kpxcDefine-btn-confirm').style.display = 'inline-block';
 };
 
-kpxcDefine.skip = function(e) {
-    if (!e.isTrusted) {
-        return;
-    }
-
+kpxcDefine.skip = function() {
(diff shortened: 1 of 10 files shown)


varjolintu:
@droidmonkey I'll make a new release after this one is approved.

keepassxc-browser/content/define.js
```diff
@@ -290,32 +286,20 @@ kpxcDefine.skip = function(e) {
     }
 };
 
-kpxcDefine.again = function(e) {
-    if (!e.isTrusted) {
```
droidmonkey: Does removing these checks expose us to any security issues?

varjolintu: No, because `isTrusted` is already handled in the main function `kpxcDefine.keyDown()`. This is why these are removed.

droidmonkey: Perfect, this is ready then

PR #642, opened 2019-10-16
Import/Export settings

### LINKED ISSUE #377: Export/Import settings option
Can there be added a way to export/import keepassxc-browser settings so that it can be used for other computers or backups?

varjolintu:
Imports and exports the current settings file as JSON.

Screen Shot 2019-10-16 at 15 09 22

Fixes #377.


droidmonkey:
Nice feature!

PR #646, opened 2019-10-18
Clean functions returning Promise

### LINKED ISSUE #645: Strange TOTP behaviour on login.xing.com with MacCtrl+Shift+T
## Expected Behavior
I would expect, that the input shortcut for TOTP would work as on all other sites, but keepassxc-browser fails to input the current TOTP without any error and leaves the selected input field blank.

## Current Behavior
keepassxc-browser seems to do nothing, I need to copy and paste the TOTP from keepassxc manually.

## Possible Solution
Strangely, while using MacCtrl+Shift+U before MacCtrl+Shift+T, an error message is presented and then the TOTP will be correctly entered.

1. login to your xing.com profile
2. see the 2fa page after a successful login
3. try to insert TOTP with the adjusted hotkey
4. use the hotkey for inserting username and passwort and then again the TOTP -> which will work.

## Debug info

KeePassXC - 2.4.3
KeePassXC-Browser - 1.5.2
Operating system: Mac/
Browser: Firefox

varjolintu:
Cleans some of the functions returning Promise. In some cases the arguments were provided as an Array, but not handled as such.

Fixes #645.

### DIFF
--- a/keepassxc-browser/background/event.js
+++ b/keepassxc-browser/background/event.js
@@ -113,9 +113,9 @@ kpxcEvent.onPopStack = function(tab) {
     return Promise.resolve();
 };
 
-kpxcEvent.onGetTabInformation = function(tab) {
+kpxcEvent.onGetTabInformation = async function(tab) {
     const id = tab.id || page.currentTabId;
-    return Promise.resolve(page.tabs[id]);
+    return page.tabs[id];
 };
 
 kpxcEvent.onGetConnectedDatabase = function() {
@@ -131,7 +131,7 @@ kpxcEvent.onGetKeePassXCVersions = async function(tab) {
         return { 'current': keepass.currentKeePassXC, 'latest': keepass.latestKeePassXC.version };
     }
 
-    return Promise.resolve({ 'current': keepass.currentKeePassXC, 'latest': keepass.latestKeePassXC.version });
+    return { 'current': keepass.currentKeePassXC, 'latest': keepass.latestKeePassXC.version };
 };
 
 kpxcEvent.onCheckUpdateKeePassXC = async function() {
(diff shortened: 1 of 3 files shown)


keepassxc-browser/background/event.js
```diff
@@ -160,7 +160,7 @@ kpxcEvent.onLoginPopup = function(tab, logins) {
     browserAction.stackUnshift(stackData, tab.id);
 
     if (logins.length > 0) {
-        page.tabs[tab.id].loginList = logins[0];
+        page.tabs[tab.id].loginList = logins;
```
droidmonkey: Do we need to check the length? Wouldn't a return of "no logins" result in something leftover in the page's login list?

keepassxc-browser/background/event.js
```diff
@@ -204,8 +204,8 @@ kpxcEvent.pageClearLogins = function(tab, alreadyCalled) {
     return Promise.resolve();
 };
 
-kpxcEvent.pageGetLoginId = function() {
-    return Promise.resolve(page.loginId);
+kpxcEvent.pageGetLoginId = async function() {
+    return page.loginId;
```
droidmonkey: Do we need an async function for this? Seems like a static data member access.

keepassxc-browser/background/event.js
```diff
@@ -218,8 +218,8 @@ kpxcEvent.pageClearSubmitted = function() {
     return Promise.resolve();
 }
 
-kpxcEvent.pageGetSubmitted = function() {
-    return Promise.resolve(page.submittedCredentials);
+kpxcEvent.pageGetSubmitted = async function() {
+    return page.submittedCredentials;
```
droidmonkey: same as above

keepassxc-browser/background/event.js
varjolintu: Well, not needed but async functions return Promise automatically, so it's a bit more clear to read.

keepassxc-browser/background/event.js
varjolintu: Oh yeah, we don't. Missed this one.

varjolintu: True. That function is not even called if the Array is empty.

PR #716, opened 2019-12-15
Ignore TOTP icons if the field is too small

### LINKED ISSUE #714: Small 2FA fields covered with KeepassXC symbols at GOG
It is convenient to have a symbol marking a field which can be filled by KeepassXC.

However, the one time password field for 2FA of `https://gog.com` are just so small that the symbols cover almost  the whole field:

gog

And actually they are not necessary at all here as KeepassXC is not supported for 2FA here afaik.

## Debug info
KeePassXC - 2.6.1
KeePassXC-Browser - 1.5.4
Operating system: Manjaro Linux
Browser: Firefox

varjolintu:
Fixes #714.

### DIFF
--- a/keepassxc-browser/content/totp-field.js
+++ b/keepassxc-browser/content/totp-field.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const MINIMUM_SIZE = 60;
+
 var kpxcTOTPIcons = {};
 kpxcTOTPIcons.icons = [];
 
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/totp-field.js
```diff
@@ -25,7 +27,7 @@ class TOTPFieldIcon extends Icon {
 };
 
 TOTPFieldIcon.prototype.initField = function(field) {
-    if (!field || field.getAttribute('kpxc-totp-field') === 'true') {
+    if (!field || field.getAttribute('kpxc-totp-field') === 'true' || field.offsetWidth < MINIMUM_SIZE) {
```
droidmonkey: We should also check to maximum input length of the field, if it is less than 4 ignore it.

droidmonkey: Also I am personally not satisfied with absolute number checks like this because you could (in theory) have a small field size with a small font.

varjolintu: Me neither. But then again, the GOG site here doesn't use the input length at all, so that must be also checked just in case, even if it's not 100% reliable.

PR #719, opened 2019-12-18
Move DOM elements to Shadow DOM

### LINKED ISSUE #600: Encapsulate `<div id="kpxcAutocomplete-list" ...>` in shadow DOM for privacy!
## Expected Behavior

The DOM tree that is added to a page and that contains the list of accounts you can log in with (`<div id="kpxcAutocomplete-list" ...>`, anything else?) should be hidden from the page's code, so it can't be read out.

## Current Behavior

Currently, a website can see all your accounts for that website by simply reading the DOM and connect the dots, which is a major privacy concern, especially in [Tor Browser](https://www.torproject.org/) which has a function to renew your identity from the website's point of view.

## Possible Solution

The DOM additions of this software can be hidden with a shadow DOM. See MDN article ["Using shadow DOM"](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM).

The shadow host should be completely anonymous without any `kpxc` or something like that, so that not even the information is available that KeePassXC is used. In case an `id` is somehow necessary on the shadow host as a contrast to the page's regular DOM, maybe a random string would be best, in a format that other browser plugins could adopt (for obfuscation).

In case support for older browsers without shadow DOM support is provided, there should be a warning that the website will be able to see the fields "Title" and "Username" of all your account entries.

## Debug info

KeePassXC - 2.4.3
KeePassXC-Browser - 1.4.7
Operating system: Win 10
Browser: Firefox Developer Edition

### LINKED ISSUE #717: Credentials dropbox styles should be important!
## Expected Behavior
Dropbox items should be visible

## Current Behavior

Dropbox css is overriden

image

## Possible Solution
make css `important!`

## Debug info
KeePassXC - 2.5.1
KeePassXC-Browser - 1.5.4
Browser: Chrome/Chromium 78.0.3904.108

varjolintu:
Moves all DOM elements to Shadow DOM.

Not originally meant as a security feature, but it adds a little more protection to the DOM elements extension is using. `Element.prototype.attachShadow` is overridden by the extension, so a page or a script cannot override it and make Shadow DOM visible. This makes receiving any usernames or other possibly sensitive information more difficult by malicious extensions or scripts.

Also, pages will no longer override CSS styles defined by the extension.

The downsides of this feature is that all CSS files must be included to the content scripts, and minimum Firefox version must be increased from 52 to 63. However, even the ESR release of Firefox is now at version 68.

Fixes #600.
Fixes #717.

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -6,6 +6,8 @@ kpxcAutocomplete.elements = [];
 kpxcAutocomplete.started = false;
 kpxcAutocomplete.index = -1;
 kpxcAutocomplete.input = undefined;
+kpxcAutocomplete.shadowRoot = undefined;
+kpxcAutocomplete.wrapper = undefined;
 
 kpxcAutocomplete.create = function(input, showListInstantly = false, autoSubmit = false) {
     kpxcAutocomplete.autoSubmit = autoSubmit;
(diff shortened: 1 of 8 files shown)


varjolintu:
Support for the latest dark theme changes has been added.

droidmonkey:
Works very nicely!

Issue #390, opened 2019-01-20
Add a "Fill Attribute" option to context menu


ceptonit:
## Expected Behavior
Add a option "Fill Attribute" to the context menu.
Here's a case scenario:
I have a Protonmail account that is protected by 2 passwords: a login password and a mailbox password.
Because there's no way to save 2 passwords in KeePassXC, I just added an additional attribute named "mailbox pass" and have the 2nd password saved to it.
So when Protonmail asks me for my Mailbox password, it'd be nice to just right-click the password field and "KeePassXC-Browser > Fill Attribute > mailbox pass".
If KeePassXC has more than 1 attribute for the website, display all of them one below another.
If there's no additional attributes for the site, hide the option or grey it out.

## Current Behavior
When you right click to a text field, you can choose between "Fill User + Pass", "Fill Pass Only", "Fill TOTP" "Show Password Generator Icons" and "Save credentials" but not "Fill Attribute".

## Debug info
KeePassXC - 2.3.4
KeePassXC-Browser - 1.3.2
Operating system: Linux Debian 9
Browser: Firefox 64.0.2

varjolintu:
There's a simple workaround for this: Add a second entry with the mailbox password page URL.

Alternatively you can use Fill TOTP to fill any attribute (a little hacky way). Add the second password to your entry to a new attribute and name it `KPH: {TOTP}`.

sebast889:
Can this be reopened? There is one gov site which asks me to fill in different pre-entered security question answer every time I log in. I'm sure there are more backwards banking/other gov sites out there that follow this method. "KeePassXC-Browser > Fill Attribute > attribute name" would make it easier to deal with these sites.

varjolintu:
Sure, let's reopen this. The easiest solution for your problem is to select the security question input field as a String Field. Then it will be filled normally when filling username and password.

To get it work:
- Select `Choose custom login fields for this site`
- Select username and password fields (you can also skip one or both of these)
- After that select the security question input field (it will be selected as a String Field)
- Add a new attribute to your entry that starts with `KPH: ` (include the space)
- Value of that attribute will be filled automatically with the username and password now when the String Field is used

If the security question has its own page, then just make a new entry with the different URL.

sebast889:
Thanks @varjolintu will try it out. But will that work for multiple attributes? They are asking for different security question answers every time not the same question.

varjolintu:
It won't. Sorry.

[...]

PR #722, opened 2019-12-19
[WIP] Allow filling attributes to fields from context menu

tremby:
I decided to have a shot at implementing #390.

This is tested so far only in Firefox.

I request feedback, please.

I reworked how the context menu is built slightly, to allow arbitrary context menu item options to be passed in, and to allow arbitrary arguments to be sent along with the action, and to allow reuse of the logic which converts from the terse form of an item to the options object passed to the browser.

Questions:

1. How are other languages usually populated? Should I start off by just grabbing translations from Google Translate?
2. How should I be handling multiple sets of credentials here? I'm currently assuming there's exactly one set, and so grabbing the first one. Please give some guidance.
3. Am I running the credentials retrieval and context menu rebuilding code correctly and at the right time? See `background/init.js`.
4. Right now it's using the `KPH` attributes, since these are the only ones I seem to be able to get from Keepassxc. Am I right in saying a change to Keepassxc is needed to get others? I personally would like it to offer me *all* the attributes in this context menu; I'd rather not have to prefix them to appear here, I think. If a change is necessary, can you help guide me on that, please? My C++ is beginner-level.
5. Am I handling the attributes correctly? I don't understand why they're retrieved as an array of objects, each with one key in it. It necessitates a pretty ugly reduce call to get them all into one object so I can get the one I'm looking for or loop through them conveniently.
6. Any thoughts about what the menu structure should be?
7. What does KPH stand for? I'm guessing the KP is Keepass...

To do:

- ~Add "Fill attribute…" message in locales other than English~
- [ ] Properly handle multiple sets of credentials
- [ ] Ensure I'm running the credentials retrieval and context menu rebuilding code correctly and at the right time
- ~Get *all* attributes, or ones with a special new prefix other than `KPH: `, since this isn't what those are for~
- [x] Ensure I'm handling the attributes correctly
- [x] Finalize menu structure
- [ ] Test in Chrome

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -11,6 +11,10 @@
         "message": "Fill Password Only",
         "description": "Context menu item for filling password."
     },
+    "contextMenuFillAttribute": {
+        "message": "Fill attribute…",
+        "description": "Context menu parent item for filling a custom attribute"
+    },
     "contextMenuFillTOTP": {
         "message": "Fill TOTP",
         "description": "Context menu item for filling Time-based One Time Password."
--- a/keepassxc-browser/background/init.js
+++ b/keepassxc-browser/background/init.js
@@ -72,7 +72,7 @@ browser.tabs.onActivated.addListener(async function(activeInfo) {
  * @param {integer} tabId
  * @param {object} changeInfo
  */
-browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+browser.tabs.onUpdated.addListener(async (tabId, changeInfo, tab) => {
     // If the tab URL has changed (e.g. logged in) clear credentials
     if (changeInfo.url) {
         page.clearLogins(tabId);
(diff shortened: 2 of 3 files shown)


varjolintu:
1. Translations are done via [Transifex](https://www.transifex.com/keepassxc/keepassxc-browser/)

2. This is problematic. Now you are doing an extra `retrieveCredentials` on tab activation and this is not a good thing. `kpxc.retrieveCredentialsCallback()` is the place for this. But because it's handled in the content script (I'd like to move this to the background script instead actually..), you'll have to send message to the background script where the context menu changes are made. This will automatically allow you to pass multiple credentials to the function that updates the context menu.

3. Like told above, no, you are not.

4. Attributes starting with `KPH` are the one you must use here. We have no need to change those.

5. This may not be the best implementation for this, and I agree it could be much better.

6. An menu item with a submenu containing the attributes.

7. I think it's originally just a short form for KeePassHTTP.

tremby:
> But because it's handled in the content script (I'd like to move this to the background script instead actually..), you'll have to send message to the background script where the context menu changes are made.

Is this your recommended approach for now?

If so, can you please point me in the right direction for how to get the data across -- how to send and receive the message in this direction? I may be missing something but I don't see that happening elsewhere, at least not into the `init.js` file, where the context menu stuff is.

> Attributes starting with `KPH` are the one you must use here. We have no need to change those.

OK. I'd really like to be able to pass arbitrary ones, but I suppose that can be a separate request, and of course would be against the main application's repo rather than here.

> This may not be the best implementation for [the attributes data structure], and I agree it could be much better.

I suppose I won't rock the boat for this patch, but maybe will try to improve it separately later.

> An menu item with a submenu containing the attributes.

This sounds like what I have at the moment is fine, then:

```
+----------------------------------+
|               ...                |
| Inspect element                  |
| Inspect accessibility properties |
+----------------------------------+----------------------------+
| KeePassXC-Browser              > | Fill Username and Password |
+----------------------------------+ Fill Password Only         +-----------------------+
                                   | Fill attribute...        > | Childhood best friend |
                                   | Fill TOTP                  | Dream job as a child  |
                                   | Show Password Generator    | Favourite teacher     |
                                   | Save credentials           +-----------------------+
                                   +----------------------------+
```

Let me know if not.

tremby:
I made an account on Transifex but don't seem able to (even request to) join the xeepassxc-browser team. It needs me to select a language to translate to, but I think all I want to do is add this new string. So it's not clear to me what the process is here.

Still hoping for guidance on how to listen for and respond to credentials having been retrieved, from the `background/init.js` file too.

phoerious:
You do not need to add source strings on Transifex. That will happen automatically next time we sync the translations with Transifex.

varjolintu:
Replaces by #1023.

PR #723, opened 2019-12-24
Use the autocomplete attribute as a hint

stefansundin:
The change: Do not add helpers on fields that have `autocomplete="off"` or `autocomplete="new-password"`. Identify fields with `autocomplete="one-time-code"` as TOTP fields.

Feel free to change things such as function naming, etc. I hope this PR is a good demonstration of what I am looking for. I used this MDN page: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

I think it would also be good to somehow be able to configure how aggressive the extension should be in trying to identify things. Sometimes I see the helpers on way too many fields and it would be nice to maybe have a "strict" mode option? Remember how browsers used to have (or still have?) a Quirks mode and Strict mode.

Anyhow, I hope you consider this. Thanks!

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -300,11 +300,16 @@ kpxcFields.isVisible = function(field) {
     return true;
 };
 
+kpxcFields.isAutocompleteAppropriate = function(field) {
+    const autocomplete = field.getLowerCaseAttribute('autocomplete');
+    return !(autocomplete === 'off' || autocomplete === 'new-password');
+};
+
 kpxcFields.getAllFields = function() {
     const fields = [];
     const inputs = kpxcObserverHelper.getInputs(document);
     for (const i of inputs) {
-        if (kpxcFields.isVisible(i) && !kpxcFields.isSearchField(i)) {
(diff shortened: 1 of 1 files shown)


varjolintu:
It needs to be tested that saving new credentials still works if `new-password` ignores the helpers.

Also, please use `===` as a strict comparison instead of `==`.

One possible change for a "strict" mode would be that it doesn't add any new helpers after a login input field combination has been detected, unless it's happening via MutationObserver. Still, this could be just a modification to the default behaviour.

stefansundin:
Hi @varjolintu. I made the change to `===`. I am not sure the best way to test if saving new credentials will still work for `autocomplete="new-password"`. Can you test it?

And you are able to make changes in my branch, in case more changes are necessary. Please feel free to do so. Thanks.

varjolintu:
Just tested it. Works great!

varjolintu:
Can you squash the commits? Thanks.

stefansundin:
@varjolintu Done. GitHub also has a squash and merge feature (if you press the arrow next to the merge button). It also let's you change the commit message.

varjolintu:
Thanks! I always forget that feature exists..

PR #729, opened 2020-01-01
Performance adjustments

### LINKED ISSUE #725: Don't observe mutations when not needed
## Expected vs Current Behavior

In playing with some "drag & sort" functionality for our intranet I realized that it became very sluggish (near-unusable) when shuffling around complex DOM bits which include input fields.  Profiling revealed that all of the CPU usage was in KeepassXC-Browser.

It appears that the extension is constantly monitoring the DOM, and unchecking its automation settings does not resolve the issue.  I expected that it would.

Possibly related to #368, #545 and #629.

## Possible Solution

When none of the "Activate *" and "Automatically *" settings are checked, `kpxcObserverHelper` should not monitor the DOM at all.  The DOM should be inspected only once, when the toolbar icon is clicked or when a context menu item is selected.  (Or when the context menu is created, if that's dynamic; I haven't investigated that bit.)

Profiling reveals that most of the wasted time is spent in:

* 40.2% `kpxc.initCredentialFields`
* 37.1% `kpxcFields.getAllFields`
* 21.2% `kpxcFields.isSearchField`
* 13.6% `kpxc.getForm`
* 11.4% `kpxcFields.isVisible`

…it seems to me that avoiding use of the observer when not needed would be the most effective way to eliminate that overhead.

## Debug info
KeePassXC - 2.5.1
KeePassXC-Browser - 1.5.4
Operating system: Linux 4.16.0-0.bpo.2-amd64 #1 SMP Debian 4.16.16-2~bpo9+1 (2018-06-26) x86_64 GNU/Linux
Browser: Chrome 76.0.3809.100 (Official Build) (64-bit)

### LINKED ISSUE #745: 100% CPU usage in both Firefox and Chrome on specific page
## Expected Behavior
Normal cpu usage expected on login page.

## Current Behavior
Constant 100% CPU in the tab in both Firefox and Chrome during password screen.

Visit: https://system.aliorbank.pl/sign-in
using login 8141111111

## Debug info
KeePassXC - 2.5.2
KeePassXC-Browser - 1.5.4
Browser: Mozilla Firefox 71.0/ Chrome 79.0.3945.88

varjolintu:
The following changes are made to increase the performance:
- `MutationObserver` is completely disabled if all features are disabled for a site via Site Preferences.
- `MutationObserver` returns immediately if mouse button is pressed (drag'n drop).
- 'mousemove' event is always returned if mouse button is not pressed.
- `kpxc.initCredentialsFields()` accepts input field array directly from the target inspected by `kpxcObserverHelper.handleObserverAdd()` instead of parsing the whole document again.
- Doesn't iterate all the inputs for password generator icons when the option is disabled
- Restricts maximum simultaneous mutations to 200
- Added an option to disable the observer. When disabled, credentials can be only filled using keyboard shortcuts, using the context menu or the popup.

Fixes #725.
Fixes #745.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -24,6 +24,7 @@ _called.manualFillRequested = ManualFill.NONE;
 let _singleInputEnabledForPage = false;
 let _databaseClosed = true;
 const _maximumInputs = 100;
+const _maximumMutations = 200;
 
 // Count of detected form fields on the page
 var _detectedFields = 0;
@@ -712,7 +713,7 @@ kpxcObserverHelper.handleObserverAdd = function(target) {
         if (Object.keys(kpxc.settings).length === 0) {
             kpxc.init();
         } else {
-            kpxc.initCredentialFields(true);
+            kpxc.initCredentialFields(true, inputs);
         }
     }
 };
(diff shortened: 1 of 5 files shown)


keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -843,6 +805,61 @@ kpxc.init = function() {
...
+// Detects DOM changes in the document
+kpxc.initObserver = function() {
+    const observer = new MutationObserver(function(mutations, obs) {
+        if (document.visibilityState === 'hidden' || kpxcUI.mouseDown) {
+            return;
+        }
+
+        // Limit the mutation handling
+        if (mutations.length > _maximumMutations) {
+            return mutations.slice(0, _maximumMutations);
```
droidmonkey: Do you want to return here? I would think you would want to do `mutations = mutations.slice(0, max)` or just keep a count in your for loop below and break after max mutations.

varjolintu: Oops. No return here. My bad. I'd rather just slice the array because it's rare that the maximum amount here is even met. So this is much faster than always calculate the loop count.

PR #739, opened 2020-01-06
Show the group name in autocomplete

### LINKED ISSUE #466: Folder option for autocomplete
## Expected Behavior
Show folder name in auto complete. Sometimes i have multiple logins to same service and sorted by folder. One folder per project and Login name is auto increment and name is the Software like Control Panel XY.
I found no option or even template with variables in settings.

## Current Behavior
Shows username (title)

## Possible Solution
Autocomplete with username (title / folder)

## Debug info
KeePassXC - 2.4.0
KeePassXC-Browser - 1.3.3
Browser: Mozilla Firefox 68.0

ulich:
For users that use groups and have similar names for multiple logins but organized in different groups

The group name is only added if there are matching logins from multiple groups. I read the idea in #466 about providing a way for templating but for now I decided to go a simpler route which maybe is sufficient as well.

Fixes #466
Requires https://github.com/keepassxreboot/keepassxc/pull/4111

I decided to change the order of values inside the string because that was the most intuitive way to display the group name for me. I hope you like it, but I am also open for any better idea.


varjolintu:
This should be optional, but maybe enabled by default.

ulich:
I added the option. What do you think?

I was only able to translate it to german and spanish, next to english. What about the the other languages?

varjolintu:
That option is only in the options page. It's not saved to the extension settings. So the functionality is still missing.

Also, languages are translated via Transifex, so there's no need to add any translations here, except for the main English translation file.

ulich:
> That option is only in the options page. It's not saved to the extension settings. So the functionality is still missing.

I compared this to other settings and its pretty much the same. I guess there is some generic de/serialization of all fields of the settings page from/to the storage? I also tested it and it does work as expected.

> Also, languages are translated via Transifex, so there's no need to add any translations here, except for the main English translation file.

Ok I removed them

varjolintu:
Thanks. I'll review this again.

keepassxc-browser/_locales/en/messages.json
```diff
@@ -599,6 +599,10 @@
         "message": "Activate autocomplete for username fields.",
         "description": "Activate autocomplete for username fields checkbox text."
     },
+    "optionsCheckboxShowGroupNameInAutocomplete": {
+        "message": "Display the group name of the credentials in the autocomplete list",
```
droidmonkey: Recommend this text: "Display the group name in autocomplete list when credentials are from different groups"

keepassxc-browser/_locales/en/messages.json
```diff
@@ -683,6 +687,10 @@
         "message": "Show a dropdown list containing available credentials for all username fields on a page.",
         "description": "Autocomplete Usernames option help text."
     },
+    "optionsShowGroupNameInAutocompleteHelpText": {
+        "message": "This only happens if there are matching credentials from different groups for a page.",
```
droidmonkey: Recommend: "When credentials are returned from different groups, the group name will be shown."

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -1080,19 +1080,28 @@ kpxc.preparePageForMultipleCredentials = function(credentials) {
         return;
     }
 
-    function getLoginText(credential) {
+    function getLoginText(credential, withGroup) {
+        const group = (withGroup && credential.group) ? `[${credential.group}] ` : '';
         const visibleLogin = (credential.login.length > 0) ? credential.login : tr('credentialsNoUsername');
+        const text = `${group}${credential.name} (${visibleLogin})`;
```
droidmonkey: This reverses the order of the credential display... should maintain that order: 
`${visibleLogin} (${group}${credential.name})`

droidmonkey:
I couldn't get this change to work. The credentials were still shown without the group name when placing credentials for the same site in different groups.

ulich:
Did you use a build of the keepassxc desktop app that adds the group to the messages for the browser extension (see the PR description)?

droidmonkey:
Oops!!

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -1115,19 +1115,28 @@ kpxc.preparePageForMultipleCredentials = function(credentials) {
...
```
varjolintu: Was this ordering already fixed? I cannot see @droidmonkey's message about the line anymore.

ulich: I don't remember any comment regarding this, can you add it again please?

varjolintu: While waiting for the answer, I would actually prefer that the group name would be added last. The username should be always visible as first.

ulich: In my opinion I would rather not add settings because already now there are so many and it's hard to understand all of them. They give the impression that KeepassXC is only for tech people but it wouldn't have to be. Also it adds complexity to the code.

Maybe rather come up with a good way to display these 3 informations in a sane way and do not fear the change. Maybe some people need to get used to it but it won't take long to adopt I believe.

But in the end it is your project. You decide and I will change it :)

varjolintu: I agree that there's too much different options already. 

Just a sidenote, but I'm already planning a settings page upgrade and it would definitely need some Advaced settings tab where we could put all extra settings which doesn't concern the normal user.

ulich:
Thanks for merging this. If you can, https://github.com/keepassxreboot/keepassxc/pull/4111 still needs to be merged so the feature can actually be used.

PR #758, opened 2020-01-20
Add dark theme

### LINKED ISSUE #509: Settings background color
## Expected Behavior

Set a font and background color or nothing

## Current Behavior

Set a font color with System background color
image

No background with inherit font color
image

## Possible Solution

Set background color OR use bootstrap intended form-control class

1. Open setting with a dark system theme

### LINKED ISSUE #715: Respect user set system color theme
## Expected Behavior
Use dark theme if user set to use dark theme, or light if the set to light.

## Current Behavior

Just provides a single light theme

## Possible Solution

I don't quite know the web ecosystem but [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) could used to see what the user set, then set the theme accordingly. See also [here](https://davidwalsh.name/prefers-color-scheme)

1. Click extension icon
2. Or input credentials and wait for 'save credentials bar'

## Debug info

KeePassXC - `2.5.1`
KeePassXC-Browser - `1.5.4`
Operating system: Linux
Browser: Firefox

varjolintu:
Respects user set dark theme and modifies the Settings and Shortcuts page CSS accordingly. Applies also to content scripts. Dark theme is also enabled for Autocomplete, Password generator and Credential banner.

The theme can be set manually with following values:
- System
- Dark
- Light

Fixes #509.
Fixes #715.


droidmonkey:
Looks great and worked immediately on Edge dark mode.

varjolintu:
I'm adding the manual selection also a bit later.

varjolintu:
The manual setting is now available. I have only tested it with a system that doesn't allow setting a dark theme globally. Need to test it with another system with a global dark theme.

w1kl4s:
Shouldn't this only apply to preferences page? After latest update it turns out it applies to every page, colliding with for example TheLounge --link-color var and other sizable chunk of variables. 

screenshot-2020-03-22_17:05:44

varjolintu:
@w1kl4s That has been already fixed with 1.6.1 (and it's released). Firefox seem to handle CSS resources in content scripts in a different way when compared to Chromium.

PR #762, opened 2020-02-02
Remove duplicate entries

### LINKED ISSUE #761: One Entry in KeePassCX menu for each URL
## Summary
I recognized that for entries where I have multiple URLs in KeePassXC stored, there are also multiple entries in the KeePassXC menu (that is shown in the login fields).

## Details
In KeePassXC I have some entries that have multiple URLs in one entry.
For example the website "Meine Allianz".
Here I have on the page "Entry" the following URL stored: https://meine.allianz.de/
On the page "Browser Integration" I have the following URLs stored:
- https://allianz.de
- https://cim.allianz.de/

When I open the URL https://meine.allianz.de/ in Chrome then the URL changes to:
https://cim.allianz.de/ui/login/de/allianz/start?....
I truncated the URL at the end.

If I click on the KeePassXC icon in the field "Username" I see the same entry two times:

image

For Amazon I also see two entries for my account.
For the Amazon Account of my wife I see 3 entries.

image

## Expected Behavior
I expect that I see only one entry for each account.
So for Allianz I see no popup menu (where I select the account) because there is only one account available.
For Amazon I expect to see only two entries: my account and the account of my wife.

## Debug info
KeePassXC - Version 2.5.3
Revision: f8c962b

Qt 5.13.2
Debugging mode is disabled.

Operating system: Windows 10 (10.0)
CPU architecture: x86_64
Kernel: winnt 10.0.18363

Enabled extensions:
- Auto-Type
- Browser Integration
- SSH Agent
- KeeShare (signed and unsigned sharing)
- YubiKey

Cryptographic libraries:
 libgcrypt 1.8.5

KeePassXC-Browser - 1.5.4
Browser: Chrome Version 79.0.3945.130 (Offizieller Build) (64-Bit)

varjolintu:
If Additional URL's are defined for the same site as the entry URL, duplicate entries returned from KeePassXC.

This fix removes any duplicate entries. Please note that this is only a workaround. The fix must be also done to KeePassXC where it must not return duplicates.

EDIT: Related PR for KeePassXC 2.6.0 https://github.com/keepassxreboot/keepassxc/pull/4288.

Fixes #761.

### DIFF
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -255,7 +255,7 @@ keepass.retrieveCredentials = async function(tab, args = []) {
             keepass.setcurrentKeePassXCVersion(parsed.version);
 
             if (keepass.verifyResponse(parsed, incrementedNonce)) {
-                entries = parsed.entries;
+                entries = removeDuplicateEntries(parsed.entries);
                 keepass.updateLastUsed(keepass.databaseHash);
                 if (entries.length === 0) {
                     // Questionmark-icon is not triggered, so we have to trigger for the normal symbol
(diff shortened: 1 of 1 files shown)


varjolintu:
And FYI: JavaScript totally lacks a function to do a Set() from array of Objects.

droidmonkey:
Wouldn't it be sufficient to just compare the UUIDs?

varjolintu:
@droidmonkey Actually, that could be enough.

PR #766, opened 2020-02-05
Fix #765, always refresh crendentials before inserting them

### LINKED ISSUE #765: Password is not being refreshed after creating a new one in KeePassXC
I am changing passwords on existing accounts and using the KeePassXC-GUI password generator (instead of the one of keepassxc-browser) because I want to configure the generation settings. So here are my steps:
1. Login with current credentials through keepassxc-browser
2. Open "change password"-dialogue
3. Fill the "current password" form through keepassxc-browser
4. Switch to KeePassXC-GUI
5. Generate a new Password for this entry in KeePassXC-GUI
6. Switch to Browser
7. Fill the "new password" form through keepassxc-browser

## Expected Behavior
The freshly generated password is inserted in the "new password" form.

## Current Behavior
The old password is inserted in the "new password" form.
This might remain unnoticed until trying to login again.

## Possible Solution
1. Automatically send a refresh from KeePassXC-GUI to keepassxc-browser.
2. Automatically refresh keepassxc-browser every time a password is being inserted.
3. Add a refresh button to keepassxc-browser.
4. Allow change of generation settings in keepassxc-browser.

I guess solution 1 would be preferred but a bit harder to implement then solution 2.
I have not inspected the code yet but I imagine keepassxc-browser currently grabs all matching entries when the website loads, right?

## Workaround
A possible but unpleasant workaround is to perform steps 1 - 6, then lock the database via keepassxc-browser and unlock it again.

## Debug info
KeePassXC - 2.3.4
KeePassXC-Browser - 1.5.4
Operating system: Linux
Browser: Firefox

Beneter:
Fix #765 Password is not being refreshed after creating a new one in KeePassXC
Implements solution 2
kpxc.fillInCredentials now always requests all credentials from KeePassXC

I have tested it once with a good case. It seems to be working fine.
I have not tested any side-effects yet.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -1222,20 +1222,17 @@ kpxc.fillInCredentials = async function(combination, onlyPassword, suppressWarni
         kpxc.p = p;
     }
 
-    if (kpxc.url === document.location.origin && kpxc.submitUrl === action && kpxc.credentials.length > 0) {
-        kpxc.fillIn(combination, onlyPassword, suppressWarnings);
-    } else {
-        kpxc.url = document.location.origin;
-        kpxc.submitUrl = action;
+    kpxc.url = document.location.origin;
+    kpxc.submitUrl = action;
 
-        const credentials = await browser.runtime.sendMessage({
-            action: 'retrieve_credentials',
-            args: [ kpxc.url, kpxc.submitUrl, true ] // Sets triggerUnlock to true
-        });
+    const credentials = await browser.runtime.sendMessage({
+        action: 'retrieve_credentials',
+        args: [ kpxc.url, kpxc.submitUrl, true ] // Sets triggerUnlock to true
+    });
 
-        await kpxc.retrieveCredentialsCallback(credentials, true);
-        kpxc.fillIn(combination, onlyPassword, suppressWarnings);
-    }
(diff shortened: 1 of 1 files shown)


droidmonkey:
I don't believe this is a proper fix and feel like it opens is up to issues.

varjolintu:
I agree with the comment above. This may fix one problem but causes more.

Beneter:
I agree, this PR was just a quick shot. I also did not look at HTTP Basic Auth.

But even after further analysis apart of some overhead due to frequent refreshes I do not see any issues with removing the if clause.
I have traced back its origin to https://github.com/lsgd/passifox/commit/386574177e3a91fc75e70b4717ac28d170ec528f#r37142657. The commit-message contains: 

> ctrl + shift + u sends new request du keepasshttp **if no credentials are found**

This enforces my assumption that its sole purpose is to reduce overhead.

droidmonkey:
I like the option to add a "Refresh" button in the pop-up dialog that requests the site's credentials again from KeePassXC regardless of whether it has a set already or not.

varjolintu:
@Beneter Your changes causes an extra retrieving of credentials every time you fill username/password for example using the autocomplete menu. The whole purpose of that condition is to just fill the previously received credentials if those are found.

The extra retrieval would also cause an access confirmation dialog to appear from KeePassXC, every time.

Beneter:
@varjolintu Yes, you are right, I did not think of the access confirmation dialog.

PR #767, opened 2020-02-06
Redirect allowance

### LINKED ISSUE #627: How to configure RedirectAllowance?
## Expected Behavior
I'd expect to be able to save passwords even when the site redirects me. However this is not the case, as the banner disappears, so I have to be quick or do it via the desktop client ( annoying UX).

## Current Behavior
I'm having same issue as https://github.com/keepassxreboot/keepassxc-browser/issues/172 

## Possible Solution

I'd accept the proposed solution there (some RedirectAllowance config/slider), but I cannot find anywhere to configure it. I don't see a reason for the dialog to close at all, unless I close it or choose one of the actions there.

## Debug info
KeePassXC - 2.4.3
KeePassXC-Browser - 1.5.1
Operating system: Mac
Browser: Firefox

varjolintu:
Allows setting _Redirect allowance_ for the new Credential Banner.

The banner will disappear after the allowance count is met. Values can be set between 1-10. One is the default (form submit is count as one). Maximum is infinite, where the banner disappears only when user dismisses it.

A new permission `webNavigation` is required for the extension to identify redirect events.

Screen Shot 2020-02-06 at 20 30 24

Fixes #627.

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -619,6 +619,14 @@
         "message": "Show a banner on the page when new credentials can be saved to the database.",
         "description": "Show login notifications checkbox text."
     },
+    "optionsRedirectAllowance": {
+        "message": "Number of allowed redirects: $1",
+        "description": "Redirect allowance range input text."
+    },
+    "optionsRedirectAllowanceHelpText": {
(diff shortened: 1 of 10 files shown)


droidmonkey:
Can we set the default to at least 2? I think 3 would be the best. This would catch 99% of all outliers.

ForsakenHarmony:
You should add that this is why you need the new permission to the release notes

PR #780, opened 2020-02-20
Fixes for 1.5.5

varjolintu:
Final fixes for the next release:
- Don't show Credential Banner if the active database is not connected.
- Remove `autocomplete="off"` check added from #723. It causes problems with a lot of pages.
- Add unique ID's dynamically detected fields.
- Switch undefined `.length` check to `.size`.


droidmonkey:
You added extra new line at end but otherwise nice

PR #786, opened 2020-02-25
TOTP icon improvements

### LINKED ISSUE #768: Improvement: No "Fill TOTP from KeePassXC" in postal/zip_code fields
See
image

When there is a `<input name=postal_code id=zipcode>` where `code` is included in name or id, the "Fill TOTP from KeePassXC" is triggered.

## Expected Behavior
No "Fill TOTP from KeePassXC" on zipcode or postalcode fields.

## Current Behavior
"Fill TOTP from KeePassXC" on postal or zipcode fields.

See https://github.com/keepassxreboot/keepassxc-browser/blob/b01ab1e8352d16ef7ea17faca1bdb05d6c856c2a/keepassxc-browser/content/keepassxc-browser.js#L9-L17
and
https://github.com/keepassxreboot/keepassxc-browser/blob/b01ab1e8352d16ef7ea17faca1bdb05d6c856c2a/keepassxc-browser/content/keepassxc-browser.js#L996-L998

## Possible Solution
Change to regex with some negative check:
```javascript
/^(((?!zip|post).)*code.*)|(.*(2fa|auth|challenge|mfa|otp|token).*)$/i
```
The `otp` and `mfa` are very short identifiers, see:
```bash
$ aspell -d en dump master | aspell -l en expand | grep -iE 'otp|mfa'
```
Maybe a score calculation is possible:
* higher score if it's a separate word
* higher score if `type=number`
* higher score if `maxlength=6` (or 8)

Or skip this all and just rely on [`autocomplete=one-time-code`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)

## Debug info

KeePassXC - 2.5.3
KeePassXC-Browser - 1.5.4
Browser: Mozilla Firefox 74.0

varjolintu:
Ignores TOTP field icons when input field type is `password` or input field name/id begins with `zip`/`postal` and ends with `code`.

Fixes #768.

### DIFF
--- a/keepassxc-browser/content/totp-field.js
+++ b/keepassxc-browser/content/totp-field.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const MINIMUM_SIZE = 60;
+const ignoreRegex = /^(zip|postal).*code$/i;
 
 var kpxcTOTPIcons = {};
 kpxcTOTPIcons.icons = [];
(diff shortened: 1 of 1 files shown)


droidmonkey:
I have seen some TOTP fields that are password fields, for some dumb reason. I cannot recall an example right now, but be prepared if this breaks someone's workflow. I think the password field might be too broad of a discriminator.

varjolintu:
@droidmonkey I'm aware of that. Sadly it's expected that something always breaks because the sites just doesn't use the standard option(s) for these.

PR #791, opened 2020-02-28
Add a copy button for version

xf-:
image

after pressing
image

and ctrl+v
```
KeePassXC - 2.5.3
KeePassXC-Browser - 1.5.4
Browser: Mozilla Firefox 75.0
```

### DIFF
--- a/keepassxc-browser/options/options.html
+++ b/keepassxc-browser/options/options.html
@@ -592,10 +592,12 @@
                     <a target="_blank" href="https://github.com/keepassxreboot/keepassxc-browser/issues/new" tabindex="14"><span data-i18n="optionsAboutCreateNewIssue"></span></a>
                     <br /><span data-i18n="optionsAboutDebugInfo"></span>:
                   </p>
-                  KeePassXC - <span class="kpxcVersion"></span>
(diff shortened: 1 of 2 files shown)


droidmonkey:
Is it not possible to add the text directly to the clipboard instead of doing a select/copy operation? At least we should deselect when we are done.

varjolintu:
That's what the password generator currently does. Just uses `.select()` for the input and then `document.execCommand('copy')`.

xf-:
@droidmonkey not really. select is only possible on a input/textarea

droidmonkey:
Can you deselect once done copying?

varjolintu:
Btw, you could just use the following:
```
$('#copyVersionToClipboard').on('click', function () {
    const copyText = document.getElementById('versionInfo');
    navigator.clipboard.writeText(copyText.textContent);
});
```

xf-:
adjusted it and innerText uses no spaces

now
```javascript
    $('#copyVersionToClipboard').on('click', function () {
        const copyText = document.getElementById('versionInfo').innerText;
        navigator.clipboard.writeText(copyText);
    })
```

PR #792, opened 2020-02-28
Use bootstrap classes

xf-:
avoid overwriting some bootstrap stuff

Bootstrap uses min-width and not max because larger screen have usually more high-end PC and not a small smartphone. d-none and d-size-display overwrites it ad min-width.

```css
@media (min-width:768px){
```

Adjusted the max mediaquery, because it would match with min-width 768 from bootstrap on taht exact pixel.

## Scollbars
Shrink you current git state to 780px or mobile size and you will get scrollbars because 160+col-md-10 has more width than 100% of your current window.

## Table
image
(Scrollbar is OS related - can be styled in last versions for windows and Linux. Not visible on some touch devices, but works with normal gestures)

### DIFF
--- a/keepassxc-browser/options/options.css
+++ b/keepassxc-browser/options/options.css
@@ -1,244 +1,111 @@
 body {
-    background-color: var(--background-color);
     font-size: 14px;
-    padding-top: 20px;
-    padding-bottom: 60px;
+    max-width: 1440px;
 }
 
 a {
-    color: var(--link-color);
-    text-decoration: none;
-}
-
-a:hover {
-    color: var(--link-hover-color);
-    text-decoration: none;
-}
-
-.table td {
-    vertical-align: middle !important;
-    width: 100%;
-    word-break: break-all;
-}
-
-table td:last-child {
-    white-space: nowrap;
-    width: 1%;
-}
-
-input {
-    background-color: var(--input-background-color);
+    text-decoration: none !important;
 }
 
 input:invalid {
     border: 2px solid red;
 }
 
-input[type=range] {
-    width: 20%;
-}
-
-input[type=text], input[type=url] {
-    max-width: 66.666667%;
-}
-
-main {
-    background: var(--background-color);
+input[disabled] {
+    opacity: .3!important;
 }
 
 footer {
     bottom: 1rem;
-    color: #cbcfcb;
-    font-size: .75rem;
-    position: absolute;
-    text-transform: uppercase;
 }
 
-select {
-    background-color: var(--input-background-color) !important;
-    color: var(--text-color) !important;
-    min-width: 100px !important;
-    width: auto !important;
+.table tbody td {
+    vertical-align: middle;
 }
 
-h2, hr {
-    margin-top: 0;
+#tab-site-preferences td:nth-of-type(2),
+#tab-site-preferences td:nth-of-type(2) select {
+    width: 200px;
 }
 
-tr.clone {
-    display: none;
-}
-
-td.identifier, td.key, td.last-used, td.created {
-    white-space: nowrap;
-}
-
-button.collapsed {
-    display: none;
+#tab-site-preferences td:nth-of-type(3),
+table td:last-of-type {
+    width: 1px;
 }
 
 .fa {
     margin-right: 0.2rem;
 }
 
-.card {
-    background-color: var(--card-background-color);
-    border: 1px solid rgba(0,0,0,.125);
-    margin-bottom: 2rem;
-    min-width: 720px;
-    width: 75%;
-}
-
-.card-header {
-    background-color: var(--card-header-color);
-    color: var(--text-color);
-    font-size: 1rem;
-    font-weight: 500;
-}
-
-.card-body {
-    background-color: var(--background-color);
-    color: var(--text-color);
-}
-
 .sidebar {
-    background-color: var(--sidebar-background-color) !important;
+    height: 50px;
     bottom: 0;
-    min-width: 160px !important;
     position: fixed;
     top: 0;
     z-index: 100;
 }
 
 .sidebar a {
-    color: #cbcfcb;
+    border: 1px solid transparent;
 }
 
-.sidebar a:hover {
-    color: #a0a0a0;
-}
-
-.nav-item, .sidebar-heading {
+h1,
+.nav-item,
+.sidebar-heading {
     font-size: .75rem;
-    text-transform: uppercase;
 }
 
 .active > a {
-    border: 1px solid #a0a0a0;
-    border-radius: 2px;
-    color: #a0a0a0;
+    border-color: currentColor;
+    color: #a0a0a0 !important;
 }
 
-.sidebar-heading {
-    color: #f2f2f2;
-    padding-bottom: 1.5rem;
-}
-
-.tab {
-    display: none;
-}
-
-.tab-header {
-    margin-bottom: 2rem;
-}
-
-.table, tbody, thead, tr, th, td {
-    border: 1px solid var(--table-border-color) !important;
-}
-
-.alert, #manualUrl {
-    width: 75%;
-}
-
-.collapse:not(.show) {
+.collapse.show {
     display: block !important;
 }
 
 .invalid-feedback {
-    color: var(--link-color);
-    display: block;
     font-size: 95% !important;
 }
 
-.invalid-feedback code {
-    color: var(--link-hover-color) !important;
-}
-
-#defaultGroup {
-    width: 50%;
-}
-
-#redirectAllowanceLabel {
-    font-weight: normal;
-}
-
-#colorTheme {
-    width: 20%;
-}
-
-#dangerousSettings {
-    display: none;
-}
-
-#padded-helptext {
-    margin-bottom: 1rem;
-    padding-left: 1.25rem;
-}
-
-@media (max-width: 768px) {
-    main {
-        padding-top: 50px;
-    }
-
-    h6.d-flex {
-        float: right !important;
-    }
-
-    button.collapsed {
-      display: block;
-      padding-top: 0.5rem !important;
-    }
-
-    .collapse:not(.show) {
-        display: none !important;
-    }
-
-    .d-flex {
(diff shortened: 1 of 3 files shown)


varjolintu:
This messes up the table view. Now the width is always 100% of the page, plus the elements inside are no longer centered. You can clearly see that in the picture. It's preferred that the width is not 100% unless the screen is too small to contain everything.

Screen Shot 2020-02-29 at 9 57 10

xf-:
On current state, the UI has some min-width even with sidebar menu you can get a horizontal scrollbar
image

I removed the min with and also tried to get rid of multiline table headers. Maybe translations in some languages are even more lines. I also removed the break-all e.g.
image

For the scrollbar i made the tables responsive. Easier to navigate.

keepassxc-browser/options/options.css
```diff
@@ -1,114 +1,68 @@
 body {
-    background-color: var(--background-color);
     font-size: 14px;
     padding-top: 20px;
     padding-bottom: 60px;
+    max-width: 1440px;
```
xf-: added a max-width like we wrote in irc

keepassxc-browser/options/options.css
```diff
@@ -1,114 +1,68 @@
 body {
-    background-color: var(--background-color);
```
xf-: removed all duplicate/overwritten values to media query

keepassxc-browser/options/options.css
```diff
@@ -1,114 +1,68 @@
...
-    color: var(--link-color);
-    text-decoration: none;
-}
-
-a:hover {
-    color: var(--link-hover-color);
-    text-decoration: none;
-}
-
-table td {
```
xf-: made the tables responsive

keepassxc-browser/options/options.css
```diff
@@ -1,114 +1,68 @@
...
-input {
-    background-color: var(--input-background-color);
-}
-
 input:invalid {
     border: 2px solid red;
 }
 
 input[type=range] {
-    width: 20%;
```
xf-: 20% depends a lot on screen size

keepassxc-browser/options/options.css
```diff
@@ -1,114 +1,68 @@
...
 }
 
-main {
-    background: var(--background-color);
-}
-
 footer {
     bottom: 1rem;
-    color: #cbcfcb;
-    font-size: .75rem;
```
xf-: using classes from bootstrap css

keepassxc-browser/options/options.css
```diff
@@ -1,114 +1,68 @@
...
-
-tr.clone {
-    display: none;
+#tab-site-preferences td:nth-of-type(1), td.identifier, td.key, td.last-used, td.created {
+    white-space: nowrap;
 }
 
-td.identifier, td.key, td.last-used, td.created {
-    white-space: nowrap;
+#tab-site-preferences td:nth-of-type(2) {
```
xf-: table width in site preferences

keepassxc-browser/options/options.css
```diff
@@ -1,114 +1,68 @@
...
-    background-color: var(--card-header-color);
-    color: var(--text-color);
-    font-size: 1rem;
-    font-weight: 500;
-}
-
-.card-body {
-    background-color: var(--background-color);
-    color: var(--text-color);
+    border-radius: 0;
```
xf-: bootstrap css again

keepassxc-browser/options/options.css
```diff
@@ -128,114 +81,40 @@ button.collapsed {
...
-
-    .collapse:not(.show) {
-        display: none !important;
-    }
-
-    .d-flex {
-        display: none !important;
-    }
-
+@media (min-width: 768px) {
```
xf-: use min-width -> bootstrap works the same way and get rid of most selectors and use existing classes

keepassxc-browser/options/options.css
```diff
@@ -251,13 +130,9 @@ button.collapsed {
         color: var(--text-color);
     }
 
-    .navbar-default {
```
xf-: moved style to upper section. no var

keepassxc-browser/options/options.html
```diff
@@ -20,41 +20,41 @@
...
-          <button class="navbar-dark navbar-toggler collapsed" type="button" data-toggle="collapse" data-target="#sidebar" aria-controls="sidebar">
+        <nav class="col-md-3 col-lg-2 sidebar">
+          <button class="navbar-dark navbar-toggler collapsed d-block d-md-none pt-2" type="button" data-toggle="collapse" data-target="#sidebar" aria-controls="sidebar">
             <span class="navbar-toggler-icon"></span>
           </button>
-          <div class="collapse" id="sidebar">
-            <h6 class="sidebar-heading d-flex px-3 mt-4 mb-1">KeePassXC-Browser</h6>
-            <ul class="nav flex-column mb-2">
+          <div class="collapse d-none d-md-block" id="sidebar">
+            <h1 class="text-uppercase text-white-50 d-none d-md-flex px-3 pb-4 mt-4 mb-1 ">KeePassXC-Browser</h1>
```
xf-: no h1 on the page before

keepassxc-browser/options/options.html
```diff
@@ -271,8 +273,8 @@ <h2 class="tab-header" data-i18n="optionsGeneralSettingsTab"></h2>
                 </div>
               </div>
 
-              <div class="card shadow">
-                <div class="card-header">
+              <div class="card my-4 shadow">
+                <div class="card-header h6">
```
xf-: h6 or h6 class is 1rem and fold - same as applied style in css before

keepassxc-browser/options/options.js
```diff
@@ -31,7 +31,7 @@ options.initMenu = function() {
         $('.sidebar:first ul.nav:first li').removeClass('active');
         $(this).parent('li').addClass('active');
         $('div.tab').hide();
-        $('div.tab#tab-' + $(this).attr('href').substring(1)).fadeIn();
+        $('div.tab#tab-' + $(this).attr('href').substring(1)).removeClass('d-none').fadeIn();
```
xf-: using d-none instead of a extra display none definition

keepassxc-browser/options/options.js
```diff
@@ -302,7 +302,7 @@ options.initConnectedDatabases = function() {
     $('#tab-connected-databases tr.clone:first .dropdown-menu:first').width('230px');
 
     const trClone = $('#tab-connected-databases table tr.clone:first').clone(true);
-    trClone.removeClass('clone');
+    trClone.removeClass('clone d-none');
```
xf-: using d-none instead of a extra display none definition

keepassxc-browser/options/options.html
```diff
@@ -599,8 +607,8 @@ <h2 class="tab-header" data-i18n="optionsAboutTab"></h2>
                 </div>
               </div>
 
-              <div class="card shadow">
-                <div class="card-header">
+              <div class="card my-4 shadow">
```
xf-: add 1.5rem space - i hope that is OK. looks the same as the space to the sidebar.

varjolintu:
There are still issues:
- When resizing the window to under 992px the sidebar width changes to a bigger size
- The table content is still not vertically centered (see screenshot)
- Extra whitespace around the checkboxes (doesn't look very nice)

Screen Shot 2020-03-01 at 14 58 09

xf-:
>     * When resizing the window to under 992px the sidebar width changes to a bigger size (from around 164px to 247 px etc.)

You used a fix min width for the menu. currently it uses bootstrap responsive design > lg the menu has 2 cols and and md to lg 3 cols for the 12 col grid. With md 2 cols the items in menu using 2 lines and it is enough space on screen for the menu. lg with 3 cols is no problem, but seems more lost again.

>     * The table content is still not vertically centered (see screenshot)
Never changed something - is this way int he develop (at least on my system)

> 
>     * Extra whitespace around the checkboxes (doesn't look very nice)
OK - i change it back to align right
> 
> 
> Screen Shot 2020-03-01 at 14 58 09

In general i would prefer cards design over the table in the future, because everything is together on screen and you don't need to follow a line to set something like username.
https://getbootstrap.com/docs/4.4/components/card/#card-decks

xf-:
@varjolintu 
image
Always 3 cols - menu gets really large and 2 cols is not enough on md (> 768).

xf-:
@varjolintu about sidebar - the current behavior or more width >= lg. This kind of behavior is used often on websites. Normal user won't resize the window.

Everything else OK?

varjolintu:
@xf- It's better now.

A few small issues:

That combo box is way too wide when compared to the initial one.
Screen Shot 2020-03-08 at 20 38 48

Is it possible to keep the spacing? See the picture below for the current layout. And I'm still not sure why the _Username-only Detection_ cell must be so wide. The current layout has a smaller one, and it looks better IMO.
Screen Shot 2020-03-08 at 20 38 58

Screen Shot 2020-03-08 at 20 39 27

xf-:
The UI is translated and depending on translation/language and line break behavior column header  can get really ugly. a fix with should keep the title in 2 lines. CZ maybe around 4 lines and other languages even more. Depending on the translation length.

About the select (combobox) - i suggest instead of fix % (like before - i used a bootstrap class) we use 100% < md and md above 3 cols. Currently it isn't really responsive on small screens.

varjolintu:
If we can, we should keep the column header width as long as the string is (like it's currently). There's no need for any extra space.

xf-:
As discussed some responsive changes

- Alerts, inputs with col widths instead of static %
- Table width more responsive. use the width if available
- Using single selector per line (more overview - i can put all in a line at the end again)

varjolintu:
The selection is still too large. Didn't you just fixed this?
Screen Shot 2020-03-12 at 9 44 53

Now there's some extra space in the delete cell too:
Screen Shot 2020-03-12 at 9 45 38

It's even worse in the Custom Login table:
Screen Shot 2020-03-12 at 9 46 25

PR #798, opened 2020-03-04
Add ignored types for TOTP field icons

### LINKED ISSUE #797: TOTP on checkbox
Example from [Uptime Robot](https://uptimerobot.com/) (related to #768):
image

```html
<input id="showQR" class="checkbox" type="checkbox" value="0" name="userAuthSetting">
```
Because of the `Auth` in the name, it's detected as TOTP field, while it is a checkbox.
## Expected Behavior
No Fill in TOTP on a checkbox.

## Current Behavior
Fill in TOTP overlay over a checkbox making it undetectable (I had to disable the addon to interact with the input).

## Possible Solution
There is a size check underway which would probably fix this specific issue, but not the underlying issue that there currently is no type checking:
https://github.com/keepassxreboot/keepassxc-browser/blob/280a561bb0276fec0ff5c5af37bd61e7cce5fb13/keepassxc-browser/content/keepassxc-browser.js#L1031-L1041

only a negative check on password:
https://github.com/keepassxreboot/keepassxc-browser/blob/280a561bb0276fec0ff5c5af37bd61e7cce5fb13/keepassxc-browser/content/totp-field.js#L30-L40

Probably only text or number or empty/blank types should be whitelisted, instead of password being blacklisted.

## Debug info
KeePassXC - 2.5.3
KeePassXC-Browser - 1.5.4
Browser: Mozilla Firefox 75.0

varjolintu:
Don't show TOTP icon on field types in the ignore list.

Fixes #797.

### DIFF
--- a/keepassxc-browser/content/totp-field.js
+++ b/keepassxc-browser/content/totp-field.js
@@ -2,6 +2,7 @@
 
 const MINIMUM_SIZE = 60;
 const ignoreRegex = /^(zip|postal).*code$/i;
+const ignoredTypes = [ 'email', 'password', 'username' ];
 
 var kpxcTOTPIcons = {};
 kpxcTOTPIcons.icons = [];
@@ -29,7 +30,7 @@ class TOTPFieldIcon extends Icon {
 
 TOTPFieldIcon.prototype.initField = function(field) {
     if (!field
-        || field.type === 'password'
+        || ignoredTypes.some(t => t === field.type)
         || field.getAttribute('kpxc-totp-field') === 'true'
         || field.offsetWidth < MINIMUM_SIZE
         || field.size < 2


keepassxc-browser/content/totp-field.js
```diff
@@ -2,6 +2,7 @@
 
 const MINIMUM_SIZE = 60;
 const ignoreRegex = /^(zip|postal).*code$/i;
+const ignoredTypes = [ 'email', 'number', 'password', 'username' ];
```
bwbroersma: Probably don't ignore `number`, sorry for my example which included some valid types.

varjolintu: True. Let's fix that.

bwbroersma: Looks good to me now.

PR #808, opened 2020-03-17
Add menu icons

xf-:
image

Smaller screen
image

Mobile screen size
image


xf-:
About the icons
- General - similar to general icon in keepassxc
- Connected is used to connect in the tab
- SignIn icon - no better idea
- Settings - a wrench or other tools are common
- About -  similar to about icon in keepassxc

PR #809, opened 2020-03-17
Bug fixes for 1.6.0.

varjolintu:
- Hide username icon properly when needed (add one extra check)
- Add missing translations
- Some code cleaning


droidmonkey:
Need to push the new translations file to transifex so we can get those in

PR #814, opened 2020-03-21
 Update datetime on reconnect

varjolintu:
If the hash already exists in the extension, the Created datetime is not updated properly. The connection is also not disconnected when the key is removed from the extension.

The manifest version bumb is made here, as this is the last PR for the new release.


varjolintu:
One thing missing.

PR #816, opened 2020-03-22
Rename root CSS names

### LINKED ISSUE #815: Extenstion changes font colors on miniflux
This is miniflux: https://github.com/miniflux/miniflux

## Expected Behavior
image

## Current Behavior
image

## Possible Solution
Looks like a CSS issue

1. Update Extention from 1.5.x to 1.6.0
2. visit self hosted Miniflux application

## Debug info
KeePassXC - 2.5.3
KeePassXC-Browser - 1.6.0
Operating system: Win32
Browser: Mozilla Firefox 74.0

This does not affect Google Chrome 80.

I found this using the "Inspect Element" Tool of firefox:
moz-extension://4ebcdf80-7448-4da2-ad3b-eba01d087839/css/colors.css Line 12 or 73

```css
:root {
    --link-color: #3a8233;
    --link-hover-color: #429f14;
}
```

Switching to light or dark theme does change the result.

varjolintu:
Firefox will collide with CSS styles even if they are put to the Shadow DOM. Renaming the variables to non-generic names will prevent it.

Fixes #815.

### DIFF
--- a/keepassxc-browser/options/options.css
+++ b/keepassxc-browser/options/options.css
@@ -82,52 +82,52 @@ h1,
 
 @media (prefers-color-scheme: dark), (prefers-color-scheme: light) {
     a {
-        color: var(--link-color);
+        color: var(--kpxc-link-color);
     }
 
     a:hover {
-        color: var(--link-hover-color);
+        color: var(--kpxc-link-hover-color);
     }
 
     body,
     pre {
-        background-color: var(--background-color);
-        color: var(--text-color);
+        background-color: var(--kpxc-background-color);
+        color: var(--kpxc-text-color);
     }
 
     main {
-        background: var(--background-color);
+        background: var(--kpxc-background-color);
     }
 
     #sidebar,
     .sidebar {
-        background-color: var(--sidebar-background-color) !important;
+        background-color: var(--kpxc-sidebar-background-color) !important;
     }
 
     input,
     input[type="checkbox"] {
-        background-color: var(--background-color) !important;
-        border-color: var(--input-main-border-color) !important;
-        color: var(--text-color) !important;
+        background-color: var(--kpxc-background-color) !important;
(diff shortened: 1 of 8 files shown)


keepassxc-browser/options/options.css
```diff
@@ -136,46 +136,46 @@ h1,
...
     }
 
     .invalid-feedback code {
-        color: var(--link-hover-color) !important;
+        color: var(--kpxc-link-hover-color) !important;
     }
 
     .invalid-feedback {
-        color: var(--link-color);
+        color: var(--kpxc-kpxc-link-color);
```
droidmonkey: got a little too happy here 😄

keepassxc-browser/options/options.css
```diff
@@ -82,52 +82,52 @@ h1,
 
 @media (prefers-color-scheme: dark), (prefers-color-scheme: light) {
     a {
-        color: var(--link-color);
+        color: var(--kpxc-kpxc-link-color);
```
droidmonkey: yipee

keepassxc-browser/options/options.css
varjolintu: Combo breaker.

LeoniePhiline:
Thanks for working on a quick fix!

However, does Firefox seriously not offer any scoping mechanism for extension code? 

I often see CSP errors in the console, where extension scripts are apparently rejected by the content security policy of the requested site. This appears to be the same issue.

varjolintu:
@LeoniePhiline I've found multiple issues with Firefox and content scripts. Where Chromium-based browsers can definitely "contain" the content scripts, but with Firefox they often collide and cause issues. For example, some element id's can work as a global variable in the content scripts (IIRC), overriding the content script's own variables etc.

In this case the stylesheets are added to the Shadow DOM but they still collide. Maybe with Firefox it's enough that the stylesheets are added to the `web_accessible_resources` list.

PR #821, opened 2020-03-23
Fix TOTP negative check too strict for (zip|postal)code

### LINKED ISSUE #820: TOTP negative check too strict (postal|zip)code problem
## Expected Behavior
Zipcode fields should not be detected as TOTP:
```html
<input name="some_zipcode">
```
See the earlier issue https://github.com/keepassxreboot/keepassxc-browser/issues/768

## Current Behavior
Zipcode & postal code fields are detected as TOTP.

## Possible Solution
Currently there is a negative check on:
https://github.com/keepassxreboot/keepassxc-browser/blob/2798380de0be9ba6b924bc1c6f9056ace0c25795/keepassxc-browser/content/totp-field.js#L4

This regex is too strict (namely `/^..$` means it will *only* remove fields that start with `postal` or `zip` and end with `code` (case insensitive).

My proposed `/^(((?!zip|post).)*code.*)|(.*(2fa|auth|challenge|mfa|otp|token).*)$/i` works better (but was a positive check). To make it a negative check just having `/(zip|postal).*code/i` would work.

## Debug info
KeePassXC - 2.5.3
KeePassXC-Browser - 1.6.1
Browser: Mozilla Firefox 76.0

bwbroersma:
Fix #820.

### DIFF
--- a/keepassxc-browser/content/totp-field.js
+++ b/keepassxc-browser/content/totp-field.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const MINIMUM_SIZE = 60;
-const ignoreRegex = /^(zip|postal).*code$/i;
+const ignoreRegex = /(zip|postal).*code/i;
 const ignoredTypes = [ 'email', 'password', 'username' ];
 
 var kpxcTOTPIcons = {};


varjolintu:
Good catch!

PR #832, opened 2020-03-29
Fix auto-reconnect with Windows

### LINKED ISSUE #650: Automatic reconnect when using the username input icon
## Expected Behavior
When using the username field icon with no active connection, a connection is automatically established in the background if possible, and the input fields are filled afterwards. Only a single click is required.

## Current Behavior
Currently, an error message is shown, and the connection needs to be (re-) established manually using the browser extension toolbar icon, and only then can the icon be used. Sometimes, I even have to reload the page. Three or even four manual steps are required.

keepassxc-browser-extension

## Debug info

KeePassXC-Browser Version: 1.5.3
KeePassXC Version: 2.4.3 
Operating system: Windows/Linux
Browser: Firefox
"Automatically reconnect to KeePassXC (experimental)" is **enabled**.

varjolintu:
Auto-reconnect fails with Windows if current KeePassXC version cannot be parsed.

Fixes #650.

### DIFF
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -1112,7 +1112,9 @@ keepass.decrypt = function(input, nonce) {
 keepass.enableAutomaticReconnect = function() {
     // Disable for Windows if KeePassXC is older than 2.3.4
     if (!page.settings.autoReconnect
-        || (navigator.platform.toLowerCase().includes('win') && !keepass.compareVersion('2.3.4', keepass.currentKeePassXC))) {
+        || (navigator.platform.toLowerCase().includes('win')
+            && keepass.currentKeePassXC
+            && !keepass.compareVersion('2.3.4', keepass.currentKeePassXC))) {
         return;
     }
 


keepassxc-browser/background/keepass.js
```diff
@@ -1112,7 +1112,9 @@ keepass.decrypt = function(input, nonce) {
 keepass.enableAutomaticReconnect = function() {
     // Disable for Windows if KeePassXC is older than 2.3.4
     if (!page.settings.autoReconnect
-        || (navigator.platform.toLowerCase().includes('win') && !keepass.compareVersion('2.3.4', keepass.currentKeePassXC))) {
+        || (navigator.platform.toLowerCase().includes('win')
+            && keepass.currentKeePassXC !== ''
```
phoerious, line 1116: I think this is good to go, but perhaps it's worth relaxing this check to anything that evaluates to false (null, undefined etc.).

varjolintu: I'm not sure if that extra check is necessary here because the variable is defined to an empty string by default, and it's only set using `keepass.setcurrentKeePassXCVersion()`. So the possible check should be made to that function instead so the comparison can remain in a simpler form.

phoerious: I would generate take advantage of weak typing where possible, but I'm fine with either solution. Whatever suits you.

varjolintu: Or maybe it's just enough to use `!keepass.currentKeePassXC` after all. It should cover all false situations, including null and undefined.

phoerious: That's what I was referring to.

PR #836, opened 2020-03-30
Ignore main HTML element from MutationObserver

### LINKED ISSUE #835: High CPU usage on Firefox 74.0
## Expected Behavior

The KPXC plugin should have a negligible impact on the performance of Firefox 74.

## Current Behavior

KPXC-Browser on some pages (e.g. etherpad) constantly increases its processor usage until FF74 grinds to a halt. This can be verified through the about:performance page

1. Have FF74 with the latest version of the KPXC-B
2. Keep FF's about:performance tab open
3. Create a document on e.g. etherpad.devol.it and type on it for a while
4. See the plugin's "Energy impact" stat increasing more and more and making the browser unusable until a) a while after stopped typing or b) the plugin has been disabled.

## Debug info

KeePassXC - 2.5.3
KeePassXC-Browser - 1.6.2
Operating system: Arch Linux
Browser: Firefox

Screenshot_2020-03-30_19-00-38

varjolintu:
Some pages make constant changes to the main HTML element via scripts. Ignoring changes directly to that element shouldn't break any sites, but some more testing might be needed.

`kpxcObserverHelper.getId()` also now generates an ID for the element based on its size, if no `classList` or `id` cannot be used.

Fixes #835.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -705,8 +705,17 @@ kpxcObserverHelper.getInputs = function(target) {
     return inputs;
 };
 
+// Gets of generates an ID for the element
 kpxcObserverHelper.getId = function(target) {
-    return target.classList.length === 0 ? target.id : target.classList;
+    if (target.classList.length > 0) {
+        return target.classlist;
+    }
+
+    if (target.id !== '') {
+        return target.id;
+    }
+
+    return `kpxc${target.clientTop}${target.clientLeft}${target.clientWidth}${target.clientHeight}`;
 };
 
 kpxcObserverHelper.ignoredElement = function(target) {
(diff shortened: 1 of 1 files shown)


droidmonkey:
This may fix the symptom but the larger problem still remains. Is there a way to more generally ignore or smooth out often changing elements? I'd imagine if an element is changing so much then we shouldn't be worried about it anyway as its likely not a username/password entry.

Just thinking out loud because we are at the point of putting our fingers in the leaky dam and running out of fingers.

varjolintu:
There's several ways this could be done. Debouncing (as suggested in the issue thread but I'm worried that it could miss some of the mutations), or using a second list for elements that doesn't allow identical additions until a certain interval is met. In this issue a same div was added and removed all the time. When user was doing something, it was really rapid, and when idle it happened maybe once or twice per second.

droidmonkey:
I think the proposed method of denouncing was extremely limited. A smarter method would only add 1 queued mutation event per element during a set collection period. Since the end state of a series of mutations is all that matters and we are operating at the speed of humans. You could collect mutations for 1 second and only add elements to the queue you haven't seen before. Then release the queue for processing while starting another collection round.

varjolintu:
That is worth trying. I'll make some tests with the Etherpad site (great for testing this).

(Sometimes I wonder if it would be easier to revert to the old method that parsed the whole page once per second. MutationObserver could be just used to trigger the parser and it could run e.g. for 3-5 seconds. Otherwise it would be disabled when idle.)

varjolintu:
I suggest we make a separate PR for the MutationObserver redesign.

PR #840, opened 2020-04-04
fix autocomplete position when body position is relative

cypx:
Hello,

I noticed autocomplete position is incorrect on https://linuxfr.org/

It seems the problem come from relative position on the body element, so I try to  fix it.

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -228,12 +228,12 @@ kpxcAutocomplete.updatePosition = function(inputField, elem) {
 
     const rect = inputField.getBoundingClientRect();
     div.style.minWidth = Pixels(inputField.offsetWidth);
-    const bodyrect = document.body.getBoundingClientRect();
-    const bodystyle = getComputedStyle(document.body);
+    const bodyRect = document.body.getBoundingClientRect();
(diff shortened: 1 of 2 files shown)


cypx:
A simple test to reproduce the bug
https://jsfiddle.net/y0atvxks/

cypx:
A testing release https://github.com/cypx/keepassxc-browser/releases/tag/1.6.2-test

keepassxc-browser/content/autocomplete.js
```diff
@@ -227,9 +227,18 @@ kpxcAutocomplete.updatePosition = function(inputField, elem) {
     }
 
     const rect = inputField.getBoundingClientRect();
-    div.style.top = Pixels((rect.top + document.scrollingElement.scrollTop) + inputField.offsetHeight);
-    div.style.left = Pixels((rect.left + document.scrollingElement.scrollLeft));
     div.style.minWidth = Pixels(inputField.offsetWidth);
+    const bodyrect = document.body.getBoundingClientRect();
```
varjolintu, line 231: Just a tiny suggestion: use `bodyRect` and `bodyStyle` instead of all-lowercase variables. Otherwise this is good to go. Nice work!

cypx: You are right, it's better.
Fixed.

PR #858, opened 2020-04-14
Fixes #857 : Extend '2fa' to '2f' and add 'pin' for matching totp fields

### LINKED ISSUE #857: Extend TOTP icon field detection
Some important websites use other field names than those tested to detect TOTP.
Atlassian JIRA and Confluence names : "2fpin"

## Expected Behavior
1. Connect to JIRA or Confluence with a 2FA account
2. TOTP field displayed with icon 

## Current Behavior
1. Connect to JIRA or Confluence with a 2FA account
2. No icon in TOTP field

## Possible Solution
Extend the matching pattern list of 'acceptedOTPFields'

## Debug info

KeePassXC - 2.5.4
KeePassXC-Browser - 1.6.2
Operating system: Win
Browser: Chrome

gmoirod:


droidmonkey:
Pin is too generic. Most banks ask for your account pin on login

gmoirod:
> Pin is too generic. Most banks ask for your account pin on login

Should i remove 'pin' so ?

droidmonkey:
The real solution to this is to define the obvious field names (which I think we did already IMO) and add the ability to choose a specific field per site. This second piece is the same process as being able to choose username / password fields for a specific site. The site + field selection prevents excessive false positive detections.

gmoirod:
and add the ability to choose a specific TOTP field per site

gmoirod:
> ... and add the ability to choose a specific TOTP field per site.

I agree. This sounds better.
Unfortunately, i do not have enough knowledge of this code to do that...

varjolintu:
> > ... and add the ability to choose a specific TOTP field per site.
> 
> I agree. This sounds better.
> Unfortunately, i do not have enough knowledge of this code to do that...

That feature has not been made yet. But it's on my TODO list.

droidmonkey:
I think this is now overcome by other features, shall we close?

varjolintu:
Yes. Let's close this one.

PR #881, opened 2020-05-05
Add a command to fill the username and password + TOTP.

m4mnux:
In some 2FA environments the password field must be filled
not only with the password but with the concatenation
of the passwword and the TOTP.

This patch allows to set up a key combination to do
do the job.

Signed-off-by: Miguel Martín <m4mnux@gmail.com>

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -44,6 +44,10 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
             _called.manualFillRequested = ManualFill.BOTH;
             await kpxc.receiveCredentialsIfNecessary();
             kpxc.fillInFromActiveElement(false);
+        } else if (req.action === 'fill_username_password_totp') {
+            _called.manualFillRequested = ManualFill.BOTH;
+            await kpxc.receiveCredentialsIfNecessary();
+            kpxc.fillInFromActiveElement(false, false, true); // appendTOTP y to true
         } else if (req.action === 'fill_password') {
             _called.manualFillRequested = ManualFill.PASS;
             await kpxc.receiveCredentialsIfNecessary();
@@ -1286,7 +1290,7 @@ kpxc.getFormSubmitButton = function(form) {
     return undefined;
 };
 
-kpxc.fillInCredentials = async function(combination, onlyPassword, suppressWarnings) {
+kpxc.fillInCredentials = async function(combination, onlyPassword, suppressWarnings, appendTOTP = false) {
     const action = kpxc.getFormActionUrl(combination);
     const u = _f(combination.username);
     const p = _f(combination.password);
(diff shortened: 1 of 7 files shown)


keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -1529,7 +1550,24 @@ kpxc.fillIn = function(combination, onlyPassword, suppressWarnings) {
         }
 
         if (pField) {
-            kpxc.setValueWithChange(pField, kpxc.credentials[combination.loginId].password);
+            let password = kpxc.credentials[combination.loginId].password;
+            if (appendTOTP) {
+                if (kpxc.credentials[commbination.loginId].stringFields && kpxc.credentials[combination.loginId].stringFields.length > 0) {
+                    const stringFields = kpxc.credentials[combination.loginId].stringFields;
```
varjolintu, line 1556: Haven't tried your PR yet, but try to get rid of this duplicate code. It's almost identical to the stuff above. Maybe a separate function (inside the function?) for it?

m4mnux: I refactored the code to use a new _'kpxc.getTOTP'_ function. I hope this looks good to you

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -1491,7 +1504,17 @@ kpxc.fillIn = function(combination, onlyPassword, suppressWarnings) {
         }
         if (pField) {
             pField.setAttribute('type', 'password');
-            kpxc.setValueWithChange(pField, kpxc.credentials[0].password);
+            let password = kpxc.credentials[0].password;
+            let totp = '';
```
varjolintu, line 1508: No need to define the variable here. You can just use `const totp = kpxc.getTOTP(kpxc.credentials[0]);` inside the if sentence.

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -1529,7 +1552,17 @@ kpxc.fillIn = function(combination, onlyPassword, suppressWarnings) {
         }
 
         if (pField) {
-            kpxc.setValueWithChange(pField, kpxc.credentials[combination.loginId].password);
+            let password = kpxc.credentials[combination.loginId].password;
+            let totp = '';
```
varjolintu, line 1556: Same here.

varjolintu:
I'm not sure if this functionality is something we even need to the extension. This is the first time I've heard some site even use that kind of login mechanism. Making a totally new feature for just one or two sites.. I'm not sure if it's worth it.

m4mnux:
How do I confirm the selection? there's no button and 'enter' key does nothing

varjolintu:
There's a Confirm button in the DOM dialog. You can also press <kbd>c</kbd> as the help text says.

m4mnux:
Pressing `c` works for me but I am not able to see any help text or any green Confirm button. :(
Not sure if the problem is the page itself or the browser version (Tried with Firefox 75.0 and Chromium 81.0.4044.92)

varjolintu:
This is how it looks for me before confirmation:
Screen Shot 2020-05-05 at 12 05 48

PR #898, opened 2020-06-04
Try to identify more form buttons

### LINKED ISSUE #897: Auto log-in broken for Discourse
## Expected Behavior
Auto-login to forum when clicking the autofill button in the username field

## Current Behavior
I get to see a Discourse page `Oops! That page doesn’t exist or is private.`

## Steps to Reproduce
I used e.g. https://discourse.opensourcedesign.net and https://forum.antennapod.org.
To try, maybe use Discourse's test instance, which also has this problem: https://try.discourse.org/

Funny enough, this issue doesn't exist on [Fairphone's forum](https://forum.fairphone.com). A difference with the other instances is that it uses a [pre-2.4 version of Discourse](https://blog.discourse.org/2020/02/discourse-2-4-released/): it doesn't yet have the Revamped User Menu. In the same blog post linked just before, also `Strict CSP On By Default` is mentioned as a new thing - might that be causing KeePassXC to not work?

## Debug info
KeePassXC - 2.5.4
KeePassXC-Browser - 1.6.3
Browser: Mozilla Firefox 76.0

varjolintu:
Sometimes form submit buttons are added outside the form, for example if a form is with ID `login-form` the button might have a property `form="login-form"` and it's not identified with a direct `querySelectorAll()` call to the form element itself.

Fixes #897.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -1283,6 +1283,16 @@ kpxc.getFormSubmitButton = function(form) {
         return buttons[0];
     }
 
+    // Try to find similar buttons outside the form which are added via 'form' property
+    for (const e of form.elements) {
+        if ((e.nodeName === 'BUTTON' && e.type === 'button')
+            || (e.nodeName === 'BUTTON' && e.type === 'submit')
+            || (e.nodeName === 'INPUT' && e.type === 'button')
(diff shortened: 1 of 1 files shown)


varjolintu:
@phoerious I can add such pages to the test PR (which has been WIP for several years already).

keunes:
Thanks @vorjolintu!

PR #899, opened 2020-06-04
Ignore observed elements for a period of time

### LINKED ISSUE #888: High CPU usage on Firefox with 1.6.x
Since the keepasxc-browser was upgraded to 1.6.0 (might be 1.6.1), in certain applications performance is really bad. Sometimes with 100% CPU usage on one core for multiple seconds and a really slow responding application. Sometimes leading to Firefox suggesting to temporarily disable the Keepassxc-browser add-on.
This CPU usage and slow response disappears when I disable the add-on.

Some alleviation of this issue has been reached in 1.6.3 (see issue #835), but the situation has not been fixed for all applications.
I experience the same issue on both Windows 10 and Fedora 31/32, which are different PC's that have a CPU capable enough and didn't present this issue < 1.6.0.

## Expected Behavior

I expect the keepassxc-browser add-on to have a minimal impact. As it has been < 1.6.0.

## Current Behavior

At the moment, in some apps, the performance impact gets so bad, I have to switch off the add-on to get back performance.

## Possible Solution

Observations:
1. This issue mainly seems to concentrate around applications that work with large numbers of dynamic rows/tables with data.
1. The CPU usage isn't attributed by Firefox to the keepassxc-browser add-on, but to the actual application tab. When disabling the add-on, though, response time and cpu usage are greatly reduced.
1. The issue increases in time: at the end of the working day, when firefox is open for longer, the issue is more present: the cpu spikes and the response delays are longer.

I still have major issues when using PhpMyAdmin, and then especially when filtering the folder list when there are a lot of foldrs in a database, for example.
I also have major issues with the [Kopano Webapp](https://demo.kopano.com/webapp/index.php) when working with larger numbers of e-mails and calendar appointments.

## Debug info

KeePassXC - 2.5.4
KeePassXC-Browser - 1.6.3
Operating system: Win & Linux
Browser: Firefox

### LINKED ISSUE #893: Extension consumes all CPU and increasing amount of memory when hovering reaction in Slack
##  Expected Behavior
Extension should never lock the tab with 100% CPU usage

## Current Behavior
Like in #888, but all CPU usage, not jus high.
I found one case where extension reliably starts consuming all CPU dedicated to the tab to the point I can only kill the process of the tab from task manager.
It happens when hovering reactions in Slack. For some reason not all reactions are affected, but some are persistently even after page reload, and those were always with reaction from one person, not multiple.
It also eats memory, about 50MiB/s, I let it consume 20GiB before I killed the process.
Disabling dynamic input field detection, as mentioned, does nothing.

## Possible Solution
I believe there must be an infinite loop somewhere.

1. Go to Slack web version
2. Hover reactions with one person reacting
3. At some point reaction hovering will cause extension to consume 100% CPU and increasing amount of memory
4. With that reaction found you can reliably trigger the issue from that point onwards (adding more people to reaction will break reproducibility)

## Debug info
KeePassXC - 2.5.4
KeePassXC-Browser - 1.6.3
Operating system: Linux
Browser: Firefox Nightly

varjolintu:
Ignores observer elements for a period of time. At this point the delay is 2 seconds.

TODO:
- Possibly switch the array push/check directly to the observer callback
- Testing

Fixes #888 
Fixes #893

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -25,6 +25,7 @@ let _singleInputEnabledForPage = false;
 let _databaseClosed = true;
 const _maximumInputs = 100;
 const _maximumMutations = 200;
+const _monitorTimeout = 2000;
 
 // Count of detected form fields on the page
 var _detectedFields = 0;
@@ -642,6 +643,7 @@ kpxcFields.useDefinedCredentialFields = function() {
     return false;
 };
 
+
 const kpxcObserverHelper = {};
 
 kpxcObserverHelper.inputTypes = [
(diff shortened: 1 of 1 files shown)


varjolintu:
FYI: The current implementation will surely not work because ironically it takes more resources to monitor all the elements that not to monitor them.

PR #906, opened 2020-06-08
Check visibility of dynamically added fields

varjolintu:
#729 adjusted the `kpxc.initCredentialFields()` so it accepts dynamically added input fields as parameters. Those fields were not checked for visibility, so this can potentially be one reason why people have reported performance issues with the latest releases. It caused all input fields to be processed even if those were invisible, and possible caused `retrieve-credentials` requests without any reason.


droidmonkey:
Just a nit, I don't like the variable name of `i` in this loop. It should be `field` or similar

PR #912, opened 2020-06-15
Final cleaning for 1.6.4

varjolintu:
Some small corrections for the next release.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -79,6 +79,7 @@
         "browser": true,
         "browserAction": true,
         "createStylesheet": true,
+        "DatabaseState": true,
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -79,6 +79,7 @@
         "browser": true,
         "browserAction": true,
         "createStylesheet": true,
+        "DatabaseState": true,
```
droidmonkey, line 82: Is "true" the right value here? Perhaps DISCONNECTED?

varjolintu: That just means Eslint can find the variable and doesn't report it as error when looking at the files.

PR #915, opened 2020-06-21
Add wildcard to the ignored URL

varjolintu:
When using the Credentials Banner and selecting _Don't ask again for this site_ always saves an absolute URL which is prone to errors. The end of the URL might change depending on the situation/token/etc.

For example `https://mail.pasteur.la/?_task=login` saved to the Site Preferences no longer works with the main site. Instead, we are now saving the URL always with a wildcard appended after the last `/` (with the remaining deleted), so the final form is `https://mail.pasteur.la/*`.

### DIFF
--- a/keepassxc-browser/content/banner.js
+++ b/keepassxc-browser/content/banner.js
@@ -107,7 +107,8 @@ kpxcBanner.create = async function(credentials = {}) {
             kpxcBanner.banner.removeChild(dialog);
         } else {
             if (ignoreCheckbox.checked) {
-                kpxc.ignoreSite([ window.top.location.href ]);
+                const ignoreUrl = window.location.href.slice(0, window.location.href.lastIndexOf('/') + 1) + '*';
+                kpxc.ignoreSite([ ignoreUrl ]);
             }
             kpxcBanner.destroy();
         }


keepassxc-browser/content/banner.js
```diff
@@ -107,7 +107,8 @@ kpxcBanner.create = async function(credentials = {}) {
             kpxcBanner.banner.removeChild(dialog);
         } else {
             if (ignoreCheckbox.checked) {
-                kpxc.ignoreSite([ window.top.location.href ]);
+                const ignoreUrl = window.location.href.slice(0, window.location.href.lastIndexOf('/') + 1) + '*';
```
droidmonkey, line 110: what happens if the URL does not end in `/`?

varjolintu: Then the URL is sliced as following:
- `https://example.com/something/testpage/login?yesplease` -> `https://example.com/something/testpage/*`
- `https://example.com/something/login.html` -> `https://example.com/something/*`
- `https://example.com/something/` -> `https://example.com/something/*`
- `https://example.com` -> `https://example.com/*`

PR #920, opened 2020-06-28
Fix high CPU usage

### LINKED ISSUE #888: High CPU usage on Firefox with 1.6.x

varjolintu:
Fixes most of the high CPU usage with pages like music.youtube.com.

Changes made:
- Some special rules are added for the Youtube site
- Fixes an issue where elements in MutationObserver's `addedNodes` or `removedNodes` were using a `nodeType` which should've been ignored. The check for the target itself was clearly not enough.
- Any filling actions etc. to ignored site is forbidden via context menu or popup.
- Many checks have been altered in a way that the fastest checks are always done first.
- `IntersectionObserver` is no longer initialized automatically.
- `kpxc.passwordFilled()` is no longer requested multiple times.
- Any input field already holding some class starting with `kpxc` is ignored and not handled again.
- `kpxc.initCredentialFields()` is now divided with a new function `kpxc.handleCredentialFields()` which doesn't have any site specific checks that are unnecessary when e.g. adding dynamic input fields for checking.

With these changes the inspected elements decreased from 942 to 241 when loading https://music.youtube.com and scrolling the page to the bottom. With https://system.aliorbank.pl/sign-in the time used for scripting is 10x smaller.

Fixes #888.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -44,6 +44,11 @@ let _documentURL = document.location.href;
 // These are executed in each frame
 browser.runtime.onMessage.addListener(async function(req, sender) {
     if ('action' in req) {
+        // Don't allow any actions if the site is ignored
+        if (kpxc.siteIgnored()) {
+            return Promise.resolve();
+        }
+
         if (req.action === 'fill_user_pass_with_specific_login') {
             kpxc.fillWithSpecificLogin(req.id);
         } else if (req.action === 'fill_username_password') {
@@ -275,6 +280,7 @@ kpxcFields.traverseParents = function(element, predicate, resultFn = () => true,
             return resultFn(f);
         }
     }
+
     return defaultValFn();
 };
 
@@ -318,20 +324,26 @@ kpxcFields.isSearchField = function(target) {
 };
 
 kpxcFields.isVisible = function(field) {
-    // Check for parent opacity
-    if (kpxcFields.traverseParents(field, f => f.style.opacity === '0')) {
+    // Check element position and size
+    const rect = field.getBoundingClientRect();
+    if (rect.x < 0
+        || rect.y < 0
+        || rect.width < 8
+        || rect.x > document.body.offsetWidth
+        || rect.y > document.body.offsetHeight
+        || rect.height < 8) {
         return false;
     }
 
     // Check CSS visibility
     const fieldStyle = getComputedStyle(field);
-    if (fieldStyle.visibility && (fieldStyle.visibility === 'hidden' || fieldStyle.visibility === 'collapse')) {
+    if (fieldStyle.visibility && (fieldStyle.visibility === 'hidden' || fieldStyle.visibility === 'collapse')
+        || fieldStyle.opacity === '0') {
         return false;
     }
 
-    // Check element position and size
-    const rect = field.getBoundingClientRect();
-    if (rect.x < 0 || rect.y < 0 || rect.width < 8 || rect.height < 8) {
+    // Check for parent opacity
+    if (kpxcFields.traverseParents(field, f => f.style.opacity === '0')) {
         return false;
     }
 
@@ -595,6 +607,16 @@ kpxcFields.getPasswordField = function(usernameId, checkDisabled) {
 };
 
 kpxcFields.prepareCombinations = async function(combinations) {
+    if (combinations.length === 0) {
+        return;
+    }
+
+    // Only request this once if there are combinations available
+    let passwordFilled;
+    if (combinations.length > 0) {
+        passwordFilled = await kpxc.passwordFilled();
+    }
+
     for (const c of combinations) {
         const pwField = _f(c.password);
         // Needed for auto-complete: don't overwrite manually filled-in password field
@@ -611,7 +633,7 @@ kpxcFields.prepareCombinations = async function(combinations) {
         // If no username field is found, handle the single password field as such
         const usernameField = c.username ? _f(c.username) : field;
 
-        if (kpxc.settings.showLoginFormIcon && await kpxc.passwordFilled() === false) {
+        if (kpxc.settings.showLoginFormIcon && passwordFilled === false) {
             kpxcUsernameIcons.newIcon(usernameField, _databaseState);
         }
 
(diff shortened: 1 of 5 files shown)


keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -318,20 +319,21 @@ kpxcFields.isSearchField = function(target) {
 };
 
 kpxcFields.isVisible = function(field) {
-    // Check for parent opacity
-    if (kpxcFields.traverseParents(field, f => f.style.opacity === '0')) {
+    // Check element position and size
+    const rect = field.getBoundingClientRect();
```
droidmonkey, line 323: What about being beyond the extent of the page as well? x > page.width()

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -745,15 +767,27 @@ kpxcObserverHelper.getId = function(target) {
```
droidmonkey: Shouldn't this be an ||?

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -896,10 +930,7 @@ kpxc.initObserver = function() {
 
         for (const mut of mutations) {
             // Skip text nodes and base HTML element
-            if (mut.target.nodeType === Node.TEXT_NODE
-                || mut.target.nodeName === 'HTML'
-                || mut.target.nodeName === 'LINK'
-                || mut.target.nodeName === 'HEAD') {
+            if (kpxcObserverHelper.ignoredNode(mut.target)) {
                 continue;
             }
```
droidmonkey, line 935: Again should be ||

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -318,20 +319,21 @@ kpxcFields.isSearchField = function(target) {
 };
 
 kpxcFields.isVisible = function(field) {
-    // Check for parent opacity
-    if (kpxcFields.traverseParents(field, f => f.style.opacity === '0')) {
+    // Check element position and size
+    const rect = field.getBoundingClientRect();
+    if (rect.x < 0 || rect.y < 0 || rect.width < 8 || rect.height < 8) {
         return false;
```
droidmonkey, line 325: What about elements that are beyond the page extents?

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -745,15 +767,27 @@ kpxcObserverHelper.getId = function(target) {
     return `kpxc${target.clientTop}${target.clientLeft}${target.clientWidth}${target.clientHeight}`;
 };
 
+kpxcObserverHelper.hasKpxcClass = function(target) {
+    if (!target.className
+        && !target.className.includes('kpxc')) {
```
droidmonkey, line 772: This should be an ||

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -1026,7 +1057,8 @@ kpxc.initCredentialFields = async function(forceCall, inputs) {
 
     _called.clearLogins = true;
 
-    if (kpxc.siteIgnored()) {
+    // No need to check this for every dynamically added inputs
+    if (inputs === undefined && kpxc.siteIgnored()) {
```
droidmonkey, line 1061: Again should be ||

keepassxc-browser/content/keepassxc-browser.js
varjolintu: Oops.

keepassxc-browser/content/keepassxc-browser.js
varjolintu: The whole point of this check is to use `kpxc.siteIgnored()` only when `kpxc.initCredentialFields()` is called without dynamically added input fields. But now when I think of it, the whole site ignored check should be done outside of this function when the page loads, not with every possible check. So making a fix.

droidmonkey: Good point! Even better

varjolintu:
@droidmonkey All the changes made. I also listed them all in the PR description.

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -1013,7 +1054,8 @@ kpxc.siteIgnored = function(condition) {
     return false;
 };
 
-kpxc.initCredentialFields = async function(forceCall, inputs) {
+// Initializes all input fields from the whole page
+kpxc.initCredentialFields = async function(forceCall) {
```
droidmonkey, line 1058: This function signature changed but don't see where any calls to it changed or where inputs was dropped from use.

varjolintu: See line 827 inside `kpxcObserverHelper.handleObserverAdd()`. The whole `inputs` parameter was added to `kpxc.initCredentialFields()` because of the input fields added by the observer.

PR #1021, opened 2020-09-26
Fix relative icon position with dir='rtl'

### LINKED ISSUE #1019: Icon obstructs input text on RTL forms that rely on text-align:right
The code here https://github.com/keepassxreboot/keepassxc-browser/blob/c90cff8f65d2ef67b30c51e38dcf1fedd620a97e/keepassxc-browser/content/ui.js#L104-L106

checks if the page has set `document.dir` to `rtl` to see if 
the inputs are RTL. But at least one major bank in israel doesn't
use that, but instead uses the `text-align:right` CSS property,
for the inputs.

So on this site, input text writes from right-to-left but keepassxc-browser 
doesn't detect that and puts the icon on the right-hand side of input, which
hides what the user types.

Website:
https://www.mercantile.co.il/MB/private (Use big green button)

varjolintu:
https://github.com/keepassxreboot/keepassxc-browser/pull/840 broke the icon position with right-to-left pages.

Also adds `text-align: right` and `dir=rtl` detection to Icon constructor.

Fixes #1019.

### DIFF
--- a/keepassxc-browser/content/ui.js
+++ b/keepassxc-browser/content/ui.js
@@ -20,7 +20,12 @@ const Pixels = function(value) {
 
 // Basic icon class
 class Icon {
-    constructor() {
+    constructor(field, databaseState = DatabaseState.DISCONNECTED) {
+        this.databaseState = databaseState;
+        this.icon = null;
+        this.inputField = null;
+        this.rtl = kpxcUI.isRTL(field);
+
         try {
             this.observer = new IntersectionObserver((entries) => {
                 kpxcUI.updateFromIntersectionObserver(this, entries);
@@ -92,7 +97,7 @@ kpxcUI.monitorIconPosition = function(iconClass) {
 
 kpxcUI.updateIconPosition = function(iconClass) {
     if (iconClass.inputField && iconClass.icon) {
-        kpxcUI.setIconPosition(iconClass.icon, iconClass.inputField);
+        kpxcUI.setIconPosition(iconClass.icon, iconClass.inputField, iconClass.rtl);
     }
 };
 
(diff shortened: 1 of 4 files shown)


droidmonkey (approved):

droidmonkey (approved):

anon324875693:
I've tested this on several RTL login pages and it does much better than current master. :+1: 

The current code will fail in the (weird case) of `<body dir=rtl>` with `text-input:left`: it would position the icon on the left, and text entry will also begin on the left side of the input.

Remember that RTL is actually about the order in which characters are displayed, and only only implicitly about text alignment. The correct icon position depends only on the text-alignment.


I've done some tests about mechanisms and relative priorities of doing RTL with HTML/CSS. 
Note that I've seen actual examples of each of these on one or more of the high-traffic sites 
I've checked.

From highest priority to lowest:
1. `text-align` value
2. `dir` attribute on `input` element
3. `dir` attribute` on `body`
4. `dir` attribute` on `html`

For example, one site has input `dir='ltr'` with `text-align:right` with the result that the text is right-aligned. So the current PR works only because it doesn't check for `dir` attribute on the input element itself (another issue). If it did and than ignored the text-align like it does with `<body dir=_>`, it would position the icon in the wrong place.

droidmonkey:
Nothing like consistency!

varjolintu:
@anon324875693 One more issue: if the input field has a parent `div` with `dir=rtl` that is not handler either. That must be fixed too.

anon324875693:
>  Lets skip that for now.

I think that's reasonable.

By and large all the sites I've seen always had an explicit text-align property. probably to preempt any cross-browser weirdness from all the possible combinations.

keepassxc-browser/content/ui.js
```diff
@@ -144,12 +144,15 @@ kpxcUI.isRTL = function(field) {
         return false;
     }
 
-    if (document.dir.toLowerCase() === 'rtl' || document.body.getLowerCaseAttribute('dir')) {
+    const style = getComputedStyle(field);
+    const fieldDir = field.getLowerCaseAttribute('dir');
+    if (style.textAlign.toLowerCase() === 'left' || fieldDir === 'ltr') {
```
anon324875693, line 149: if textalign is set to *either* right or left, the value of `dir` makes no difference.

For example, consider `textalign=right` and d `fieldDir=='ltr', then this first conditional would match
and isRTL would be set to false. But if `textalign=right` the text will be right aligned so this won't work.

keepassxc-browser/content/ui.js
```diff
@@ -134,6 +139,23 @@ kpxcUI.deleteHiddenIcons = function(iconList, attr) {
     }
 };
 
+kpxcUI.isRTL = function(field) {
```
anon324875693, line 142: `isRTL` is not the best name. input text could be RTL and be either left *or* right aligned.
Icon position simply needs to know if the effective text is aligned right or left - than position the icon
on the opposite side. So perhaps a better name would be `isRightAligned` or some such?

keepassxc-browser/content/ui.js
varjolintu: That's the priority? `text-align` always first, then the field attribute.

anon324875693: yes, I toggled things to see what takes precedence and text-align always wins, then the value of the dir property starting with the element itself if present or else inheriting from any ancestor that has `dir` attribute set, going up the ancestors until the topmost `<html>` tag.

anon324875693: How's this?
```js
kpxcUI.isRTL = function(elem) {
    if (!elem) {
        return false;
    }

    const style = getComputedStyle(elem);
    if (style.textAlign.toLowerCase() === 'left') {return false;}
    else if (style.textAlign.toLowerCase() === 'right') {return true;}

    const loopLimit = 128;
    for (let i = 0; i < loopLimit; i++) {
        const elemDir = elem.getLowerCaseAttribute('dir');
        if (elemDir === 'ltr') { return false; }
        else if (elemDir === 'rtl') { return true; }
        elem = elem.parentElement;
        if (!elem) break;
    }

    return false;
};

```

varjolintu: We already have a `kpxcFields.traverseParents()` function for these kind of situations.

anon324875693: Like this?
```js
kpxcUI.isRTL = function(element) {
    if (!element) {
        return false;
    }

    const style = getComputedStyle(element);
    if (style.textAlign.toLowerCase() === 'left') {return false;}
    else if (style.textAlign.toLowerCase() === 'right') {return true;}

    return kpxcFields.traverseParents(element, 
        (element) => ["ltr", "rtl"].includes(element.getLowerCaseAttribute('dir')),
        (element) => ({"ltr" : false, "rtl" : true})[element.getLowerCaseAttribute('dir')]
    ) 
};
```

varjolintu: Something like that yes :)

anon324875693: steal my code... *please*!

varjolintu: Works great. Thanks!

anon324875693: lgtm

anon324875693:
Merge?

PR #1040, opened 2020-10-11
TOTP ignoredTypes fix

### LINKED ISSUE #855: TOTP icon for name=author|coupon_code|postcode|bankcode|activationcode
## Expected Behavior
No TOTP icon for
```html
<input name=author>
```

## Current Behavior
TOTP icon
image

## Possible Solution
https://github.com/keepassxreboot/keepassxc-browser/blob/4a04055bb47823c6fc16d1c32f55466eda47bcad/keepassxc-browser/content/totp-field.js#L4
A fix would to add `author` to the ignore regex:
```javascript
/(zip|postal).*code|author/i
```
but that would include quite some other maybe auth-y words like `authori[sz]e` and `authori[sz]ation`:
```bash
aspell -d en dump master | aspell -l en expand | grep -iE 'author'
```
so it might ignore too much. A bit more complex regex that does work:
```javascript
/(zip|postal).*code|author([^a-z]|s|$)/i
```
But maybe I'm over-engineering now, and we should just ignore author?

## Debug info
KeePassXC - 2.5.3
KeePassXC-Browser - 1.6.2
Browser: Mozilla Firefox 77.0

### LINKED ISSUE #1030: username field misdetected as TOTP on major bank website
https://www.bankhapoalim.co.il is one of israel's largest banks.

The following html:
```html
<input formcontrol name="userCode" id="userCode" inputmode="text" type="text">
```

trips the totp detector because "Code" appears in the name/id. In fact, this `userCode` field is an alphanumeric string assigned by the bank as a username.

## Possible solution

https://github.com/keepassxreboot/keepassxc-browser/blob/c90cff8f65d2ef67b30c51e38dcf1fedd620a97e/keepassxc-browser/content/totp-field.js#L3

adjust `ignoreRegex` to exclude `userCode` (most-specific fix).

Note: was #1024

### LINKED ISSUE #1039: TOTP detected multiple times on soundcloud and "disable all features" doesn't help
## Expected Behavior

Selecting "disable all features" in the settings also disables any TOTP fields. At best all of these fields aren't mis-detected, but as  workaround it would already help to have an option disabling totp detection for that domain.

## Current Behavior

Using soundcloud.com with 2FA, if you're logged in every commentary field is detected as TOTP field. Selecting "disable all features" for soundcloud.com changes nothing regarding TOTP detection.

## Possible Solution

As stated, disable TOTP field detection with "disable all features" selection in extension settings.

1. Login to soundcloud.com
2. Go to an artist site and klick play

## Debug info

KeePassXC - 2.6.1
KeePassXC-Browser - 1.7.1
Operating system: Linux
Browser: Firefox
image

varjolintu:
Adds some more strings to ignoreRegex. Fixes invalid TOTP detections for example with SoundCloud.

Fixes #855.
Fixes #1030.
Fixes #1039.

### DIFF
--- a/keepassxc-browser/content/totp-field.js
+++ b/keepassxc-browser/content/totp-field.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const ignoreRegex = /(zip|postal).*code/i;
+const ignoreRegex = /(act|bank|coupon|postal|user|zip).*code|comment|author/i;
 const ignoredTypes = [ 'email', 'password', 'username' ];
 const MIN_INPUT_LENGTH = 6;
 const MAX_INPUT_LENGTH = 10;


keepassxc-browser/content/totp-field.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const ignoreRegex = /(zip|postal).*code/i;
+const ignoreRegex = /(act|bank|coupon|postal|user|zip).*code|comment|author/i;
```
droidmonkey, line 3: How is that regex evaluated? My read is that it would match "zipcomment" and "bankauthor" but I don't think that's the intention.

droidmonkey (changes requested):

keepassxc-browser/content/totp-field.js
droidmonkey: Related, if we are going to match a name we should make it an inclusion instead of exclusion. My thought is there are far more false positives on "****code****" then true positives like "pincode" or "totpcode"

anon324875693: How about setting up a form/issue/wikipage/json_file for users to report this kind of information for the sites they care about? That should then serve as the basis for regression tests.

For that matter, protonmail uses "twoFactorCode".

varjolintu: We are doing both inclusion and exclusion. I'd like to keep it that way. The Regex must be added to the test PR later, but must be tested properly for this PR first.

droidmonkey: ok no problems thanks!

droidmonkey (approved):

PR #1066, opened 2020-10-24
Content script tests

varjolintu:
All tests files are copied to the extension before running them. Without this action the content scripts cannot be run. This also prevents including the tests files in the extension bundle. Those files cannot be included in the release.

It can be accessed manually from: `extension://<extensionId>/tests/tests.html` if the test script file is modified and the delete test folder action is deleted.

Usage: `npm run tests`

TODO:
- [x] Tests for general functions (global.js)
- [x] Tests for input field detection (keepassxc-browser.js)
- [x] Tests for TOTP fields (totp-field.js)
- [x] Tests for search fields (keepassxc-browser.js)
- [x] Tests for password change (keepassxc-browser.js)
- [x] Run tests in headless mode with Mocha and selenium-webdriver

Screenshot:
Screenshot 2020-10-28 at 23 18 39

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -491,16 +491,18 @@ kpxcFields.isVisible = function(elem) {
     const rect = elem.getBoundingClientRect();
     if (rect.x < 0
         || rect.y < 0
-        || rect.width < 8
+        || rect.width < MIN_INPUT_FIELD_WIDTH_PX
         || rect.x > Math.max(document.body.scrollWidth, document.body.offsetWidth, document.documentElement.clientWidth)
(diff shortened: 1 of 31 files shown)


phoerious:
Can you integrate this into TeamCity?

varjolintu:
@phoerious It should be possible. I have zero experience with configuring TeamCity so it can take a while.

droidmonkey (approved):
Why is the package-lock.json file so large now?

varjolintu:
@droidmonkey Lots of dependencies in the packages 💁‍♂️

phoerious:
I hope we only need them for running the tests?

varjolintu:
@phoerious For tests yes. But those need to be downloaded also for the build script. I'm not aware of a method that could only download necessary dependencies per needed script.

PR #1085, opened 2020-11-03
Remove one TOTP ignore regex

varjolintu:
Removes `act` from `ignoreRegex` because it matches e.g. twoF**act**or, breaking Protonmail and possibly some other sites.

I'm not happy about this one. We really need to re-implement this whole accept/ignore TOTP in the future. Before that I will do any necessary tests for the Content Tests branch which includes 2FA fields from most common sites (and the ones we have had problems with), so we will have a possibility to quickly test any fixes.


droidmonkey:
I never liked that regex... wasn't act supposed to be for account, why not just expand it?

droidmonkey (approved):

PR #1086, opened 2020-11-03
Fix showing autocomplete list without input field focus

### LINKED ISSUE #1081: Dropdown list with logins is not displayed on page where login fields are not autofocused
## Expected Behavior

On a page, where the login form does not get the focus automatically on page load, and there are multiple logins available for that site, pressing `Alt+Shift+u` should display the dropdown list with available logins to select from.

## Current Behavior

Pressing `Alt-Shift-u` fills in the first available login from the available logins.
You don't get the option to select.
You have to press `Alt-Shift-u` again (or the `down` arrow) for the drop-down list to appear.

## Possible Solution

Might be related to the fix for #1053

1. Go to a page that does not set focus to the login field automatically, and for which you have more than one login saved in KeepassXC, e.g. https://www.otto.de/user/login
2. Press `Alt-Shift-u`
3. First login is automatically filled in to the login form, there is no option to select from the list of logins

## Debug info

KeePassXC - 2.6.1
KeePassXC-Browser - 1.7.2
Browser: Mozilla Firefox 82.0

philipreimer:
If input field does not have focus:
- show autocomplete list if there are multiple credentials
- otherwise fill in credentials

Fixes #1081

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -727,19 +727,22 @@ kpxc.fillInFromActiveElement = async function(passOnly = false) {
     }
 
     const el = document.activeElement;
-    if (el.nodeName !== 'INPUT') {
-        // No active input element selected -> fill the first combination found
-        if (kpxc.combinations.length > 0) {
-            kpxc.fillInCredentials(kpxc.combinations[0], kpxc.credentials[0].login, kpxc.credentials[0].uuid, passOnly);
-
-            // Focus to the input field
-            const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
-            if (field) {
-                field.focus();
-            }
+    if (el.nodeName !== 'INPUT' && kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
+        // No active input element selected -> focus to the input field
+        const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
+        if (field) {
+            field.focus();
         }
 
-        return;
+        if (kpxc.credentials.length > 1) {
+            // More than one credential -> show autocomplete list
+            kpxcAutocomplete.showList(field);
+            return
+        } else {
+            // Just one credential -> fill the first combination found
+            kpxc.fillInCredentials(kpxc.combinations[0], kpxc.credentials[0].login, kpxc.credentials[0].uuid, passOnly);
+            return;
+        }
     } else if (kpxc.credentials.length > 1 && kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
         kpxcAutocomplete.showList(el);
         return;
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -727,25 +727,26 @@ kpxc.fillInFromActiveElement = async function(passOnly = false) {
     }
 
     const el = document.activeElement;
-    if (el.nodeName !== 'INPUT' && kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
-        // No active input element selected -> focus to the input field
-        const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
-        if (field) {
-            field.focus();
-        }
 
-        if (kpxc.credentials.length > 1) {
-            // More than one credential -> show autocomplete list
-            kpxcAutocomplete.showList(field);
-            return
-        } else {
-            // Just one credential -> fill the first combination found
-            kpxc.fillInCredentials(kpxc.combinations[0], kpxc.credentials[0].login, kpxc.credentials[0].uuid, passOnly);
+    if (kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
+        const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
+        if (field && field.id !== el.id) {
+            // focus to the correct input field
+            field.focus();
+
+            if (kpxc.credentials.length > 1) {
+                // More than one credential -> show autocomplete list
+                kpxcAutocomplete.showList(field);
+                return
+            } else {
+                // Just one credential -> fill the first combination found
+                kpxc.fillInCredentials(kpxc.combinations[0], kpxc.credentials[0].login, kpxc.credentials[0].uuid, passOnly);
+                return;
+            }
+        } else if (kpxc.credentials.length > 1) {
+            kpxcAutocomplete.showList(el);
             return;
         }
-    } else if (kpxc.credentials.length > 1 && kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
-        kpxcAutocomplete.showList(el);
-        return;
     }
 
     // No previous combinations detected. Create a new one from active element
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -726,30 +726,24 @@ kpxc.fillInFromActiveElement = async function(passOnly = false) {
         return;
     }
 
-    const el = document.activeElement;
-
     if (kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
         const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
-        if (field && field.id !== el.id) {
-            // focus to the correct input field
-            field.focus();
+        // set focus to the input field
+        field.focus();
 
-            if (kpxc.credentials.length > 1) {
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -719,19 +719,22 @@ kpxc.fillInFromActiveElement = async function(passOnly = false) {
...
-        // No active input element selected -> fill the first combination found
-        if (kpxc.combinations.length > 0) {
-            kpxc.fillInCredentials(kpxc.combinations[0], kpxc.credentials[0].login, kpxc.credentials[0].uuid, passOnly);
-
-            // Focus to the input field
-            const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
-            if (field) {
-                field.focus();
-            }
+    if (el.nodeName !== 'INPUT' && kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
```
droidmonkey, line 722: Shouldn't we be checking if the active element is not the first kpxc combination field? We could theoretically be focused on a search box input field (for example) and result in the bug again.

droidmonkey (changes requested):

keepassxc-browser/content/keepassxc-browser.js
droidmonkey: My thought on this in pseudo code would be:

```
if (autocomplete_enabled)
	focus on kpxc.combinations[0]
	if (credentials > 1)
		show choice list
	else
		fill credentials
	return
endif
```

philipreimer: @droidmonkey this would simplify the code, I guess.

But the check for `kpxc.combinations.length > 0` would still be required, right? So the initial `if` would be:
`if (kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {`?

And that would enclose current lines 722 - 741.

Is the part from line 743 on the only remaining case then?

droidmonkey: Good point, yes need to ensure credentials are even available. I would put the autocomplete conditional first since that is the main purpose of the block.

varjolintu: This should just be:
```
    if (el.nodeName !== 'INPUT') {
        // No active input element selected -> fill the first combination found
        if (kpxc.combinations.length > 0) {
            // Focus to the input field
            const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
            if (field) {
                field.focus();
            }

            if (kpxc.credentials.length === 1) {
                kpxc.fillInCredentials(kpxc.combinations[0], kpxc.credentials[0].login, kpxc.credentials[0].uuid, passOnly);
                return;
            }

            kpxcAutocomplete.showList(field);
        }

        return;
    }
```

droidmonkey: Why does it matter if the current element is not an input field? Also, we only want to do this if auto complete is active.

varjolintu: @droidmonkey Because it indicates we are not focused to any input field when trying to do the fill from keyboard. But true, the check for AutoComplete option is also needed.

philipreimer: @varjolintu shouldn't `fillInCredentials` and `showList' be inside the `if (field)` block?

varjolintu: Yes.

philipreimer: I think @droidmonkey was right. It is not sufficient to check for `nodeName !== 'INPUT'`. In that case the autocomplete list would open in any input field that is currently focused, if `kpxc.credentials.length > 1`.

For example go to https://www.otto.de/user/login and click into the search field at the top. Then press `Alt-Shift-u`. If you have multiple logins the list would open beneath the search field.

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -719,22 +719,26 @@ kpxc.fillInFromActiveElement = async function(passOnly = false) {
...
-            kpxc.fillInCredentials(kpxc.combinations[0], kpxc.credentials[0].login, kpxc.credentials[0].uuid, passOnly);
-
-            // Focus to the input field
-            const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
-            if (field) {
-                field.focus();
+
+    if (kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
+        const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
+        if (field && field.id !== el.id) {
```
droidmonkey, line 725: Remove this if statement, just focus on the field no matter what.

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -719,22 +719,26 @@ kpxc.fillInFromActiveElement = async function(passOnly = false) {
...
+                return
+            } else {
+                // Just one credential -> fill the first combination found
+                kpxc.fillInCredentials(kpxc.combinations[0], kpxc.credentials[0].login, kpxc.credentials[0].uuid, passOnly);
+                return;
             }
+        } else if (kpxc.credentials.length > 1) {
+            kpxcAutocomplete.showList(el);
+            return;
         }
```
droidmonkey, line 741: Remove this block

droidmonkey (changes requested):

keepassxc-browser/content/keepassxc-browser.js
philipreimer: done

keepassxc-browser/content/keepassxc-browser.js
philipreimer: done.

This seems to be working now but it kind of breaks the login at accounts.google.com - at least for me. It still works fine for the username, but fails on the second "page" of their login flow when you have to enter the password:

* `Alt+Shift+i` does not have any effect at all
* `Alt+Shitft+u` will display the autocomplete list, but it is placed in the upper left corner of the screen.

varjolintu: Gotta check it.

philipreimer: I think the reason is that on the password page `Alt+Shift+u` also tries to focus the username field and display the autocomplete list there. But the username field does not "exist" on the password page - at least not visibly.

So this might also affect other pages with this kind of two-step login, like Paypal for instance.

varjolintu: It's logical because filling username **and** password is not working with only password fields. You'll have to use _Fill Password_ instead.

philipreimer: Unfortunately,
> Alt+Shift+i does not have any effect at all

droidmonkey (approved):

droidmonkey:
This works perfectly for me

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -726,26 +726,24 @@ kpxc.fillInFromActiveElement = async function(passOnly = false) {
...
     }
 
-    const el = document.activeElement;
-    if (el.nodeName !== 'INPUT') {
-        // No active input element selected -> fill the first combination found
-        if (kpxc.combinations.length > 0) {
+    if (kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
+        const field = passOnly ? kpxc.combinations[0].password : kpxc.combinations[0].username;
+        // set focus to the input field
+        field.focus();
```
droidmonkey, line 732: Need to check for `null` here

droidmonkey (changes requested):

keepassxc-browser/content/keepassxc-browser.js
philipreimer: can it be null if `kpxc.combinations.length > 0`?

varjolintu: The `null` check is not enough here. At the Google login page the password field is actually in the second `combination`. So if `passOnly` is used, maybe all input fields from the combinations must be checked until a password field is found. If not, then just return.

Ugly almost-like-this-code:
```
        let field;
        if (passOnly) {
            const passwordField = kpxc.combinations.find(c => c.password);
            if (passwordField && passwordField.password) {
                field = passwordField.password;
            }
        } else {
            field = kpxc.combinations[0].username;
        }

        if (!field) {
            return;
        }
```

varjolintu (approved):

philipreimer:
@varjolintu there is still one open request for change but it seems to be related to the outdated request above. I believe this has been addressed in the latest changes. Or do you want me to change anything else?

droidmonkey (approved):

droidmonkey:
Good to go!

PR #1122, opened 2020-11-21
Move isSearchField() to a better location

varjolintu:
Moves `isSearchField()` inside `kpxcObserverHelper.getInputs()` so search fields can be filtered out in one place instead of checking it in multiple functions.

Fixes username field icon with e.g. https://outlook.live.com.


droidmonkey (approved):

droidmonkey:
This fixes the search detection problem entirely!

PR #1140, opened 2020-12-12
Add supporters

varjolintu:
Adds a list of Patreon supporters to the extension About tab page.

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -1019,6 +1019,10 @@
         "message": "The following debug info can be used",
         "description": "Debug info text."
     },
+    "optionsAboutSupporters": {
+      "message": "Supporters",
+      "description": "Supporters label text."
+    },
     "optionsErrorInvalidValue": {
         "message": "Invalid value. Please check the input.",
         "description": "Error shown when input field has an incorrect value."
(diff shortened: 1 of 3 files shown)


phoerious:
Your prices are cheaper. I guess we should unify at what tier you become a listed sponsor. @droidmonkey

keepassxc-browser/options/options.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+const SUPPORTERS_URL = 'https://raw.githubusercontent.com/wiki/keepassxreboot/keepassxc-browser/Supporters.md';
```
phoerious, line 3: Should we really request a third-party service without notifying the user? This would need an entry in the privacy policy.

varjolintu: Good point. It would be strange to make this as an optional "feature", and at the same time we shouldn't make any connections to outside if the user wishes so.

varjolintu: But well. I'm all in if we want to take the familiar route and just show a hard-coded list we update in every new release.

phoerious: Just add it to the repository.

droidmonkey:
I made the requirement harder over time to favor the early donors who kickstarted everything. I only really add VIP donors at this point.

varjolintu:
> Your prices are cheaper. I guess we should unify at what tier you become a listed sponsor. @droidmonkey

Less people, less contributors, cheaper prices?

droidmonkey (approved):

PR #1142, opened 2020-12-12
Add TOTP possibility to Predefined Sites

varjolintu:
Allows specific TOTP rules for certain sites, e.g. Amazon (which has `maxLength` defined as 20, which we normally ignore).

### DIFF
--- a/keepassxc-browser/content/totp-field.js
+++ b/keepassxc-browser/content/totp-field.js
@@ -2,8 +2,8 @@
 
 const ignoreRegex = /(bank|coupon|postal|user|zip).*code|comment|author/i;
 const ignoredTypes = [ 'email', 'password', 'username' ];
-const MIN_INPUT_LENGTH = 6;
-const MAX_INPUT_LENGTH = 10;
+const MIN_TOTP_INPUT_LENGTH = 6;
+const MAX_TOTP_INPUT_LENGTH = 10;
 
 const acceptedOTPFields = [
     '2fa',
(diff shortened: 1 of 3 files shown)


keepassxc-browser/content/totp-field.js
```diff
@@ -48,6 +48,12 @@ kpxcTOTPIcons.isAcceptedTOTPField = function(field) {
     return false;
 };
 
+// Double the allowed maxLength if an exception is found (e.g. Amazon's 2FA)
+kpxcTOTPIcons.isLengthValid = function(field) {
+    const maxLength = kpxcSites.exceptionFound(PREDEFINED_TOTP) ? MAX_INPUT_LENGTH * 2 : MAX_INPUT_LENGTH;
+    return (field.maxLength > 0 && (field.maxLength < MIN_INPUT_LENGTH || field.maxLength > maxLength));
+};
```
droidmonkey, line 55: This is a very specific action (double the length) for a generic response (there is an exception). I would much prefer to not re-use the exception function which is already rather ungainly, but instead have a function such as "expectedTotpLength(site_url)" that would return the default max length in all cases except for Amazon (for now).

droidmonkey (changes requested):

keepassxc-browser/content/totp-field.js
varjolintu: Good idea. I'll make the necessary changes.

keepassxc-browser/content/totp-field.js
```diff
@@ -57,7 +63,7 @@ kpxcTOTPIcons.isValid = function(field, forced) {
         if (ignoredTypes.some(t => t === field.type)
             || field.offsetWidth < MINIMUM_INPUT_FIELD_WIDTH
             || field.size < 2
-            || (field.maxLength > 0 && (field.maxLength < MIN_INPUT_LENGTH || field.maxLength > MAX_INPUT_LENGTH))
+            || kpxcTOTPIcons.isLengthValid(field)
```
droidmonkey, line 66: These checks seem to be returning true if the field is NOT valid, same for the isLengthValid function. It should be renamed as a result, perhaps isTOTPFieldLengthInvalid

droidmonkey (changes requested):

keepassxc-browser/content/totp-field.js
droidmonkey: I would also maybe just inline the function entirely as it was before. There is no need for a separate function anymore.

varjolintu: Check.

droidmonkey (approved):

PR #1193, opened 2021-01-22
Ignore buttons with different formaction

### LINKED ISSUE #1192: Ubuntu Login (Launchpad) sends to password forgotten page, rather than hitting 'Log in' button
When trying to log in to [Ubuntu One/Launchpad](https://login.launchpad.net/) I can click on the green key icon KPXC's browser plugin. It fills in the login details, but then sends me to the password reset page.

This only seems to happen when I come from https://askubuntu.com and want to log in with my Launchpad credentials. So I'm on the page https://login.launchpad.net/4Qfm3QTOmhZ9cVA9/+decide and not on https://login.launchpad.net/ where I don't seem to have this issue.

## Expected Behavior

Fill in credentials and be logged in

## Current Behavior

Fill in credentials and open password reset page

see above

## Debug info

KeePassXC - 2.6.3
KeePassXC-Browser - 1.7.4
Operating system: Linux
Browser: Firefox

varjolintu:
Form buttons have `document.location.href` as a default `formaction`. If `formaction` is set, it will override the default `action`. Those buttons must be ignored.

Fixes #1192.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -189,8 +189,9 @@ kpxcForm.getFormSubmitButton = function(form) {
     }
 
     // Try to find another button. Select the last one.
-    // TODO: Possibly change this behavior to select the last one for only certain sites.
-    const buttons = Array.from(form.querySelectorAll(kpxcForm.formButtonQuery));
+    // If any formaction overriding the default action is set, ignore those buttons.
+    const buttons = Array.from(form.querySelectorAll(kpxcForm.formButtonQuery))
+                         .filter(b => b.formAction === document.location.href);
     if (buttons.length > 0) {
         return buttons[buttons.length - 1];
     }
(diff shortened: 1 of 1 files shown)


droidmonkey:
This looks like it will only allow buttons that are set to the document href. Is that the default setting for submit without formaction explicitly set? This code makes me nervous, is there a more explicitly way to exclude buttons that have explicitly set formaction?

varjolintu:
@droidmonkey This makes me a little nervous too because I find it odd that `formaction` could actually be the `href` every time if not set. I have no idea why it's not empty or just `undefined`. I'll inspect some other ways to make the exclusion.

varjolintu:
Ok I finally solved it. If you accessed the element directly using `element.formAction` it could return a non-null value, but using `element.getAttribute('formAction')` got the correct result. So if `formAction` is not set, `null` is returned.

droidmonkey (approved):

droidmonkey:
That's a handy function!

PR #1227, opened 2021-02-18
Reload credentials after save

### LINKED ISSUE #1225: prevent creation of duplicated database entries after initial Save credentials and banner header appearance
KeePassXC-Browser 1.7.6
KeePassXC 2.6.4

steps: fill in credentials on webpage -> right mouse -> KeePassXC-Browser -> Save credentials -> 

Then the banner header appears to Username or password changed! Save it? -> New

1. Please change this banner header when it is a new database entry, because the username or password did not change!

Now we login and the banner header appears again with the same message, there is no Update option, only New -> Press New and a second copy of the database entry will be created.

This behavior is sadly to complicated for our average users and I would love to see this being improved.

2 Please change this, after the "Save credentials" reload the webpage so the new credentials are available or use some other mechanism to not show the "Username or password changed! header again after a successful login.

3. When connecting the KeePassXC-Browser to KeePassXC  please reload the webpage as well, so the credentials become available for the current webpage.

4. When disabling the "Show a banner on the page when new credentials can be saved to the database" the right mouse -> Save credentials options does not work anymore. Can we directly add the credentials without the banner header showing?

I will donate a 150 euro bounty when these above issues are fixed in the next KeePassXC-Browser release for Firefox and Chrome.

2021-02-18-181308_1028x686_scrot
2021-02-18-181353_1477x69_scrot

Jelle de Jong

varjolintu:
After user has saved new or modified credentials using the Credential Banner, new credentials are immediately loaded. This prevents the banner to reappear after the login and seeing the recently saved credentials as a new one. It also shows the updated credentials immediately in Autocomplete Menu and the popup login menu.

- [x] Reload credentials after saving new/modified ones
- [x] Reload credentials after database connect (Firefox gives me a headache with this bug https://bugzilla.mozilla.org/show_bug.cgi?id=1665380)
- [x] Save credentials to the default group when Credential Banner is disabled

Fixes #1225.

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -268,14 +268,15 @@ page.removePageInformationFromNotExistingTabs = async function() {
 
 // Retrieves the credentials. Returns cached values when found.
 // Page reload or tab switch clears the cache.
+// If the retrieval is forced (from Credential Banner), get new credentials normally.
 page.retrieveCredentials = async function(tab, args = []) {
-    const [ url, submitUrl ] = args;
-    if (page.tabs[tab.id] && page.tabs[tab.id].credentials.length > 0) {
+    const [ url, submitUrl, force ] = args;
+    if (page.tabs[tab.id] && page.tabs[tab.id].credentials.length > 0 && !force) {
         return page.tabs[tab.id].credentials;
     }
 
     // Ignore duplicate requests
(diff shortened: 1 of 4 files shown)


droidmonkey (approved):

varjolintu:
I'm still investigating another improvement: a forced credential load after database has been connected. It works for Chromium-based browsers but Firefox still has some problems. I'll add that one to this PR too.

PR #1010, opened 2020-09-13
Add predefined sites option

varjolintu:
Adds option for predefined sites which are added to Site Preferences at the extension start. This feature is enabled by default to ensure a better compatibility with most common sites. Duplicate URL's are not appended to the list.

Another tweaks:
- Move generic scripts from the root directory to _common/_
- Remove `kpxc.initSitePreferences()` and do the same check at the extension start

Please report any common sites you know that are missing from the list.

Currently the list is:
- accounts.google.com
- www.paypal.com
- *.live.com
- amazon.com
- amazon.de
- amazon.co.uk
- signin.aws.amazon.com
- www.upwork.com/ab/
- home.personalcapital.com
- auth.services.adobe.com
- idmsa.apple.com
- soundcloud.com
- icloud.com
- signin.ebay.de
- signin.ebay.com
- login.yahoo.com
- (All other eBay domains)
- *.atlassian.com

Fixes #1033.
Fixes #1042.
Fixes #1073.

PR #1231, opened 2021-02-19
Increase reliability of custom login fields functionality

ostrolucky:
Some (possibly a lot of) websites, like https://042564560466.signin.aws.amazon.com/console nowadays are generating dynamic classNames in their forms. ClassNames there depend on multitude factors, such as which field is currently selected or if field is empty. In case of formerly mentioned websites, this looks to be the case for AngularJS framework, which injects classes like `ng-untouched`, `ng-valid`, `ng-not-empty` and so on. This means keepassxc detects these fields first time only (and sometimes again randomly). I don't know reasons why this extension went for a solution where classNames were playing major part of the role of identifying the fields, but it turned out to be unreliable. This patch removes this factor and instead relies on more stable attributes, especially input/target name. Extra/custom field associations in other password field managers like 1Password7 and Enpass detect the fields automatically, very reliably. I think this is closer to how they do it.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -432,12 +432,16 @@ kpxcFields.getCombination = async function(field, givenType) {
 
 // Gets of generates an unique ID for the element
 kpxcFields.getId = function(target) {
+    if (target.name) {
+        return `${target.nodeName} ${target.type} ${target.name} ${target.placeholder}`;
+    }
+
     if (target.classList.length > 0) {
-        return `${target.nodeName} ${target.type} ${target.classList.value} ${target.name} ${target.placeholder}`;
+        return `${target.nodeName} ${target.type} ${target.classList.value} ${target.placeholder}`;
     }
 
     if (target.id && target.id !== '') {
-        return `${target.nodeName} ${target.type} ${kpxcFields.prepareId(target.id)} ${target.name} ${target.placeholder}`;
+        return `${target.nodeName} ${target.type} ${kpxcFields.prepareId(target.id)} ${target.placeholder}`;
     }
 
     return `kpxc ${target.type} ${target.clientTop}${target.clientLeft}${target.clientWidth}${target.clientHeight}${target.offsetTop}${target.offsetLeft}`;


varjolintu:
We need to improve the Custom Login Field detection for pages which use dynamic class names. The whole reason we went to this solution was to hide the extension from the web page. Previously we marked the input fields with our own classes, which wasn't transparent enough.

Sadly, this PR is not increasing the reliability of the Custom Login Fields enough. It only decreases it, because many pages have input fields with similar classes and types, with no names or placeholders.

ostrolucky:
I don't think I have yet seen login forms where inputs don't have names assigned. But what do you propose to do instead?

varjolintu:
Identifying elements with dynamic class names and other properties is a bit tricky. The only way this could be more reliable is to detect a form first (this is easy of course) and then use the child elements directly without using any names. Meaning some kind of detection like: "this form has 8 elements, the second element contains 4 children, third one of those contains the input fields we want". Of course this would also break up if the page content would change in the future, so it's not a perfect solution either.

By the way, I haven't had any problems logging in with AWS sites. And if a problem is found with such a popular login page like AWS, we suggest making a fix using the newer [Predefined Sites](https://github.com/keepassxreboot/keepassxc-browser/pull/1010) feature which allows making exceptions for such pages.

ostrolucky:
As far as I can see, https://github.com/keepassxreboot/keepassxc-browser/pull/1010 is for forms with single input only. In AWS console, the case I'm solving is form having 3 inputs. What I expect is to be able to map _Account ID_ thanks to keepassc-browser "**Choose custom login fields**" functionality. If everything is successfully setup, autofill process then looks like following:
aws

This is a great feature of this extension and not even nowadays biggest competion of keepass - Bitwarden has it. I don't think you use this, hence why you didn't run into this issue.

> Sadly, this PR is not increasing the reliability of the Custom Login Fields enough. It only decreases it, because many pages have input fields with similar classes and types, with no names or placeholders.

I have changed my patch now so it addresses this. It checks now if input has name first and only if it doesn't, it uses other, existing strategies.

varjolintu:
Another nice improvement would be an extra check for the username and password field so it's not possible for those to be identical (like in issue https://github.com/keepassxreboot/keepassxc-browser/issues/1233). With that kind of cases, the input field size and offset should be included to the saved values.

varjolintu:
Btw, this PR is probably obsolete because of the new incoming changes in https://github.com/keepassxreboot/keepassxc-browser/pull/1243. I tested the Amazon page linked in this PR and it worked flawlessly.

varjolintu:
The fix works great and it's included in #1243 :) Shall we move the discussion there?

ostrolucky:
Sure :)

PR #1239, opened 2021-02-22
Do not allow identical field ID's with Custom Login Fields

varjolintu:
When creating ID's for Custom Login fields, always skip to the next ID method if it's a duplicate.

### DIFF
--- a/keepassxc-browser/content/define.js
+++ b/keepassxc-browser/content/define.js
@@ -386,7 +386,8 @@ kpxcDefine.confirm = async function() {
     }
 
     if (kpxcDefine.selection.password) {
-        kpxcDefine.selection.password = kpxcFields.getId(kpxcDefine.selection.password.originalElement);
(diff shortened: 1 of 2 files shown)


droidmonkey:
Why do we prefer class list over the actual ID of the field?

varjolintu:
> Why do we prefer class list over the actual ID of the field?
> 
> I also find this storage / ID method rather arbitrary and prone to random errors and minor changes to the website. We should think about better targeting methods.

This identification is a tricky one. Some pages don't use ID's at all. Some pages use only classes. Some use nothing at all. I have some ideas for a better implementation that doesn't rely on the ID at all, but it'll take time to polish. Extra points go to the sites that add classes and/or ID's dynamically to the input fields, so the existing ID methods are not working properly.

Changing this to draft. Related PR: https://github.com/keepassxreboot/keepassxc-browser/pull/1231

PR #1243, opened 2021-02-23
Use XPath for element ID creation

### LINKED ISSUE #1195: travelesolutions.com: custom login fields not working
The page `https://travelesolutions.com/app/login` has 3 fields: username, password, and company short name. Username and password can be filled successfully with no custom settings. However, once the 3rd field is set up as a string field using the custom login fields, then none of the fields work.

## Expected Behavior

KeepassXC logo shows on the username field, and clicking it results in all 3 fields (Username, password, and company short name) filled

## Current Behavior

KeepassXC logo does not show next to any fields

1. Set up an entry in KeepassXC for the page `https://travelesolutions.com/app/login` and provide some text for username and password
2. Visit the site in a browser. Observe that the browser extension recognizes the username and password fields correctly and fills in the information
3. Set up custom login fields so that the "company short name" field is selected for string field #1
4. Edit KeepassXC entry to add attribute "KPH: Company Short Name" and put data in such as "abcd"
5. Visit `https://travelesolutions.com/app/login` again and note that no fields are recognized
6. Right-clicking on any of the 3 fields to select filling in username & pw, or pw only, results in nothing happening. 
7. Right-clicking on any of the 3 fields to select filling in an attribute "company short name" works correctly.

## Debug info

KeePassXC - 2.6.3
KeePassXC-Browser - 1.7.4
Operating system: Win32
Browser: Chrome/Chromium 87.0.4280.141

### LINKED ISSUE #1233: keepassxc-browser does not fill the password in the password field
## Expected Behavior
Be able to fill the password to the password field

## Current Behavior
I try the function of Custom Login Fields, but it does not work.
image

1. link: https://www.ctbcbank.com/twrbc/twrbc-general/ot001/010
2. click button and it failed

## Debug info
KeePassXC - 2.6.4
KeePassXC-Browser - 1.7.6
Operating system: Win64
Browser: Chrome/Chromium 88.0.4324.182

### LINKED ISSUE #1259: TOTP isn't usable on Proxmox VE
I can get a demo Proxmox VE system online if anyone needs it to test against it.

But basically if you're just doing user/pass then the auto-fill is fine.

However if also doing TOTP then that comes up afterwards but it doesn't reload the page to show it (it's like an extra HTML div on top).  So the icon doesn't show to click it to fill in TOTP.

Have tried to do this with custom fields... (as I can still select user/pass fields behind that TOTP div)... but it just results in it only filling out the password oddly...

How things look if you enable the overlay to show fields...

Selection_378

How the TOTP screen looks normally...

Selection_379

How you can select the TOTP field to be used as the TOTP field...

Selection_380

Finally how if you do all that... the auto-fill just oddly fills in the password...

Selection_381

varjolintu:
Use XPath instead of a custom generated ID for HTML elements. New ID as XPath can be for example `/html/body/div/div/div[2]/div[2]/div[1]/div/div/form/div[4]/div[2]/input` so it doesn't care any names or classes, which might change dynamically, but relies on the page content and element ordering.

Fixes #1195.
Fixes #1233.
Fixes #1259.

### DIFF
--- a/keepassxc-browser/content/define.js
+++ b/keepassxc-browser/content/define.js
@@ -382,20 +382,20 @@ kpxcDefine.confirm = async function() {
     }
 
     if (kpxcDefine.selection.username) {
-        kpxcDefine.selection.username = kpxcFields.getId(kpxcDefine.selection.username.originalElement);
+        kpxcDefine.selection.username = kpxcFields.setId(kpxcDefine.selection.username.originalElement);
     }
 
     if (kpxcDefine.selection.password) {
-        kpxcDefine.selection.password = kpxcFields.getId(kpxcDefine.selection.password.originalElement);
+        kpxcDefine.selection.password = kpxcFields.setId(kpxcDefine.selection.password.originalElement);
     }
 
     if (kpxcDefine.selection.totp) {
-        kpxcDefine.selection.totp = kpxcFields.getId(kpxcDefine.selection.totp.originalElement);
+        kpxcDefine.selection.totp = kpxcFields.setId(kpxcDefine.selection.totp.originalElement);
     }
 
     const fieldIds = [];
     for (const i of kpxcDefine.selection.fields) {
-        fieldIds.push(kpxcFields.getId(i));
+        fieldIds.push(kpxcFields.setId(i));
     }
 
     const location = kpxc.getDocumentLocation();
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -430,8 +430,24 @@ kpxcFields.getCombination = async function(field, givenType) {
     return undefined;
 };
 
-// Returns an unique ID for the element
-kpxcFields.getId = function(target) {
+// Sets and returns unique ID's for the element
+kpxcFields.setId = function(target) {
+    return [ kpxcFields.getIdFromXPath(target), kpxcFields.getIdFromProperties(target) ];
+};
+
+// Returns generated unique ID's for the element. If XPath ID fails, return the fallback one.
+kpxcFields.getId = function(idArray, inputField) {
+    if (!idArray || idArray.length < 2) {
+        return '';
+    }
+
+    const elementFromXPath = kpxcFields.getElementFromXPathId(idArray[0]);
+    const fallbackId = kpxcFields.getIdFromProperties(inputField);
+
+    return elementFromXPath || (fallbackId === idArray[1] ? inputField : '');
+};
+
+kpxcFields.getIdFromXPath = function(target) {
     let xpath = '';
     let pos;
     let temp;
@@ -456,8 +472,24 @@ kpxcFields.getId = function(target) {
     return xpath;
 };
 
-kpxcFields.getElementFromId = function(id) {
-    return (new XPathEvaluator()).evaluate(id, document.documentElement, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue;
+kpxcFields.getIdFromProperties = function(target) {
+    if (target.name) {
+        return `${target.nodeName} ${target.type} ${target.name} ${target.placeholder}`;
+    }
+
+    if (target.classList && target.classList.length > 0) {
+        return `${target.nodeName} ${target.type} ${target.classList.value} ${target.placeholder}`;
+    }
+
+    if (target.id && target.id !== '') {
+        return `${target.nodeName} ${target.type} ${kpxcFields.prepareId(target.id)} ${target.placeholder}`;
+    }
+
+    return `kpxc ${target.type} ${target.clientTop}${target.clientLeft}${target.clientWidth}${target.clientHeight}${target.offsetTop}${target.offsetLeft}`;
+}
+
+kpxcFields.getElementFromXPathId = function(xpath) {
+    return (new XPathEvaluator()).evaluate(xpath, document.documentElement, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue;
 };
 
 // Check for new password via autocomplete attribute
@@ -573,9 +605,9 @@ kpxcFields.useCustomLoginFields = async function() {
     }
 
     // Finds the input field based on the stored ID
-    const findInputField = async function(inputFields, id) {
-        if (id) {
-            const input = inputFields.find(e => e === kpxcFields.getElementFromId(id));
+    const findInputField = async function(inputFields, idArray) {
+        if (idArray) {
+            const input = inputFields.find(e => e === kpxcFields.getId(idArray, e));
             if (input) {
                 return input;
             }
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -430,17 +430,34 @@ kpxcFields.getCombination = async function(field, givenType) {
     return undefined;
 };
 
-// Gets of generates an unique ID for the element
+// Returns an unique ID for the element
 kpxcFields.getId = function(target) {
-    if (target.classList.length > 0) {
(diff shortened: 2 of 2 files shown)


ostrolucky:
Isn't it more likely that element structure will change than than input names change? I really am afraid of coupling to DOM structure instead of input names and hence would prefer #1231. DOM changes all the time. If not dynamically, then with next website update, which would break whatever users saved in extension.

varjolintu:
@ostrolucky I disagree. DOM rarely changes with login fields. The site in issue #1233 is a perfect example why the previous solutions won't work with sites. That page has the following fields:
```
<input formcontrolname="custIxd" maxlength="11" type="text" class="ng-pristine ng-valid ng-touched" tabindex="1">
<input autocomplete="off" formcontrolname="userIxd" maxlength="12" type="password" class="ng-pristine ng-valid ng-touched" tabindex="2">
<input autocomplete="off" formcontrolname="pxd" maxlength="12" type="password" class="ng-pristine ng-valid ng-touched" tabindex="3">
```

With those two password fields, how do you distinquish them? Some pages might not even use `tabindex` and `formcontrolname` is probably some Angular specific stuff.

The label "more testing needed" is actually needed, because I want to find a site where this PR doesn't work.

droidmonkey:
Once the login form is shown, any DOM changes are completed. In which case the extension should be able to properly detect and tag the input fields based on the XPATH. I disagree that the DOM changes while you are using login fields.

ostrolucky:
Do you at least agree that it's quite common that layout of website changes when web developers implement various features that business requires? Coupling to xpath will mean that even if this works reliably now, saved custom fields will suddenly stop working next time website layout change is rolled out. 

I mean, xpath solution is much better than former `target.clientTop` stuff, but it's still more prone to breakages than binding to input names, when they are available. I'm not saying we don't need this PR, but that https://github.com/keepassxreboot/keepassxc-browser/pull/1231 should be preferred where possible, which isn't the case of #1233 obviously - for that we definitely need something better than current `target.clientTop` things.

droidmonkey:
The likelihood of a css change is far more likely. Targeting by id has already proven to be unreliable.

ostrolucky:
If business requirement is "add a new box here", that's not going to be a CSS change, you know that. We don't really need to talk about CSS here, since it's not related to either PR. CSS doesn't change DOM on which we rely.

Change of input names is much less likely than change of ID or classNames or order of elements, or anything like that, because it has implications on backend of websites. Most of the time, input name is actually used as a field identificator on websites backends and sent in POST. It's defacto an API for communicating between frontend and backend, so changing it is uncommon. IDs, classnames, layouts, all of that is fair game for changing because backend doesn't need to be adjusted for that, change stays on frontend. So I am not surprised if binding to IDs didn't work out. I would be very surprised if relying on input names wasn't more reliable than IDs etc.

varjolintu:
Relying to input names is going solve this problem only partially. Custom Login Fields are mostly needed on pages where it's problematic to identify the fields at all, or if user needs a String Field features. Some of the problematic pages don't uses names or id's at all with the HTML elements. Those which use, can always change the names to something else and the Custom Fields are no longer valid. As @droidmonkey said, this is a game of cat and mouse. It's quite difficult to make a generic solution that works for all possible scenarios. So far using the DOM structure has been more succesful for me when I've tried out the pages.

droidmonkey:
@varjolintu it might be good to add a notification to the user when we can no longer detect the custom fields they previously defined and offer to redefine them.

droidmonkey:
Input names don't have to be unique and for pages with multiple form submissions can be duplicated many times on the same page.

vectorjohn:
Yeah, storing the full XPath to the fields I would expect to immediately fail on some sites, and frequently fail on others. Just using name also won't work a lot of the time, but it really needs to be some combination.
Like, 99% of the time the inputs will be actual <input> elements, and they'll be inside a form, so if you used an xpath like (I don't remember xpath syntax so I'll use selector) "form[0] input[x]". I.e. username is the xth input on the first form. As a web developer, I would expect that or something like it to be very stable. I agree with ostrolucky that input names are the best and most stable but clearly there needs to be a backup in cases where they're not found. Maybe use input name as first option, and fall back to this.

If I have time I'll try to look into doing a PR myself.

varjolintu:
@vectorjohn I agree that combining these two methods would be a good solution. If the first one fails, we have a backup plan. Would you agree with this plan @ostrolucky ?

droidmonkey:
I like the backup plan idea, store both data, prefer XPath and fallback to name with form number perhaps.

ostrolucky:
Well yes, that was my suggestion too. Prefer input names and fall back to xpath.

droidmonkey:
No there is a nuance that you did not have and I want to be clear about. We store BOTH XPath and name, then when searching in the future we try finding the field with the XPath, failing that we fallback to the name method.

vectorjohn:
I admit, I was not considering the fact that it's easy to re-assign custom fields. To be clear, I was thinking the way xpath is used in this PR might be brittle because I expect it to be not uncommon for sites to change the structure of the page such that an exhaustive xpath from root to the input would break. Something looser where you only include forms and inputs in the xpath might be more resilient (e.g. 2nd input of first form).
But I don't think it would break frequently enough to be a big issue in practice, since users can just re-select the input field. In that light I'm coming around to liking this PR for the simplicity.

varjolintu:
The latest commit now generates two ID's: XPath ID and a fallback ID using @ostrolucky's fix. I tested this using a local site where I chose custom login fields and then modified the page so that the login fields are inside a `div`. The XPath ID failed in that case but the fallback ID still worked without selecting any new fields.

The notification for users to reset Custom Login Fields is still missing. One thing we might consider is to keep the legacy ID generation intact, and if there's no ID array identified the fallback goes to the legacy ID. This prevents breaking the previous settings user might have set, but when choosing the fields next time those will be saved using the new ID pair. The legacy ID could be removed later. Too much fuzz, or should be choose this method instead of showing a notification for all users?

droidmonkey:
Maybe we should keep the code that reads the property and uses it to find the element (prefix these functions with legacy perhaps?). If found, then write the XPATH/Name method moving them to the "next gen" version.

varjolintu:
@droidmonkey That could work too. We don't want to spam the notification, but only show it when it's absolutely necessary.

droidmonkey:
Prototype Code:
```
Has Custom Fields?
	Are Custom Fields "Legacy"?
		Load Legacy Custom Fields
		Are Fields Found?
			Save Fields as XPATH/Name
		Else
			Show Custom Fields Not Found Notification
	Else
		Load Modern Custom Fields
		Fields Not Found?
			Show Custom Fields Not Found Notification
```

varjolintu:
@droidmonkey I've added the necessary changed with one exception: identified legacy fields are not stored with XPath/Name. Instead those are just converted to XPath/Name on-the-fly if found. I think this is good enough. When storing new Custom Login Fields, it overrides the legacy ones.

Now the notification is also shown when fields are not found when expected.

droidmonkey:
I have a form that hides the password field until you fill in and submit the username field. If I select the custom fields using just username field it works. If I select them after submitting the username (ie, select username and password) then it fails on page reload. I think it is trying to find the password field that is currently hidden and fails entirely. Perhaps we need to add a provision to find as many fields as possible, finding 1 or more we continue. Then on page mutation we run the find operation once again to find the new field.

varjolintu:
@droidmonkey Yeah, those two page / dynamic login sites still aren't working properly with Custom Login Fields. I'll see what I can do to improve this.

droidmonkey:
We could move that phase of this to a new PR entirely. It just happened to strike on my test site. Otherwise this PR is good to go and works nicely.

droidmonkey (approved):

varjolintu:
@droidmonkey True. There's some previous issues related directly to that.

PR #1251, opened 2021-02-28
Add exception for Protonmail's mailbox password input

### LINKED ISSUE #1249: Can't fill mailbox password on app.protonmail.ch / mail.protonmail.com
## Expected Behavior

KeePassXC-Browser should fill both the login username / password and the mailbox decrypt password.

## Current Behavior

The login username / password is filled but not the mailbox decrypt password, even if you manually tell it to.

## Possible Solution

I think the bug is happening because KeePassXC-Browser is not detecting the mailbox decrypt box as a password input. This could be solved by either adding more customizability (scripting support) or pre-programming Protonmail support into the extension.

1. Open https://app.protonmail.ch or https://mail.protonmail.com.
2. Fill login details automatically or explicitly.
3. Page takes you to the mailbox decryption step, where you will be unable to fill it using KeePassXC-Browser.

## Debug info

KeePassXC - 2.6.4
KeePassXC-Browser - 1.7.6
Operating system: Linux
Browser: Waterfox G3 (basically Firefox)

varjolintu:
Fixes #1249.

### DIFF
--- a/keepassxc-browser/common/sites.js
+++ b/keepassxc-browser/common/sites.js
@@ -46,22 +46,29 @@ kpxcSites.googlePasswordFormUrl = 'https://accounts.google.com/signin/v2/challen
 kpxcSites.googleUrl = 'https://accounts.google.com';
 kpxcSites.savedForm = undefined;
 
-// Handles a few exceptions for certain sites where password form is inside a div
-// or another element that is not detected directly. Triggered by MutationObserver.
-// Returns true if an Element has a matching classList.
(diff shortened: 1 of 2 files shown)


droidmonkey (approved):

strasharo:
Is there any special guidance how to save the mailbox password in order to have it recognized? For me it looks like it's currently filling something else, I guess the main entry password instead of the custom string I named "KPH: mailboxPassword"

varjolintu:
For me it's just another entry. You can use that attribute but fill it from the context menu.

Issue #559, opened 2019-05-31
unhelpful error "Key exchange was not successful." when native messaging host not set up


khazhyk:
[TIP]:  # ( Provide a general summary of the issue in the title above ^^ )
[TIP]:  # ( DO NOT include screenshots of your actual database! )

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen )

it connects to the database

## Current Behavior
[NOTE]: # ( Tell us what actually happens )

first:
"KeePassXC-Browser has encountered an error:

Cannot connect to KeePassXC. Check that browser integration is enabled in KeePassXC settings. "

then click "reload" gives:
"KeePassXC-Browser has encountered an error:

Key exchange was not successful. "

This happens if keepassXC is open or closed, suggesting to me something broken within the browser extension 

## Possible Solution
[NOTE]: # ( Not required, but suggest a fix/reason for the bug )

## Steps to Reproduce
[NOTE]: # ( Provide a link to a live example, or an unambiguous set of steps to )
[NOTE]: # ( reproduce this bug. Include code to reproduce, if relevant )
1. start firefox with keepascx-browser
2. click the icon and press reload
3.

## Context
[NOTE]: # ( How has this issue affected you? What unique circumstances do you have? )
unusable

## Debug Info
[NOTE]: # ( Paste debug info from Help → About here )
KeePassXC - 2.4.1
KeePassXC-Browser - 1.4.5
Operating system: MacIntel
Browser: Mozilla Firefox 67.0

khazhyk:
debug console shows:

Connecting to native messaging host org.keepassxc.keepassxc_browser keepass.js:1003:13
Failed to connect: Unknown error keepass.js:979:13
Error: Could not establish connection. Receiving end does not exist.

- issue was keepassxc was enabled for chrome but not firefox...

so I guess, change this to... maybe handle this error and give more helpful error message. Perhaps... "KeepassXC is not configured, please set it up in preferences > browser integration and make sure <browser> is checked" or something

varjolintu:
Those error messages have been changed few months ago. The problem here is that the same error message is used for several situations, and there's no way to distinguish them.

tjuerges:
Independent of whatever the error message stands for, I see the same error message on MacOS with 1.4.5/2.4.2 and Chrome.

Was this ever tested?

dxdc:
^^ same here.

varjolintu:
[...]

PR #1257, opened 2021-03-05
Output the error from the port on Disconnect to the debug logs

Ferk:
Right now the error output on disconnection is taken from `browser.runtime.lastError` which might only make sense if the timming is exactly right.

According to Mozilla's documentation the onDisconnect should be receiving the port argument which contains an error field, I expect this should be the right place to look for a meaningful error to display.
See: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/Port

This is related to: #559

### DIFF
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -1015,7 +1015,8 @@ keepass.onNativeMessage = function(response) {
     }
 };
 
-function onDisconnected() {
+function onDisconnected(port) {
+    let error = (port && port.error) || browser.runtime.lastError;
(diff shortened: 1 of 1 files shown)


droidmonkey:
So does it provide more info? Please show an example before and after your coffee submission

varjolintu:
I couldn't get any extra information from it. Plus the code didn't work without some adjustments to it.

droidmonkey:
Recommend closing this

varjolintu:
Closing this as the PR doesn't work properly and the collaborator hasn't been updating it.

PR #1278, opened 2021-03-24
Ignore Auto-Submit option for Site Preferences

### LINKED ISSUE #1120: Add "ignore autofill" in the Site preferences menu
I use one website, where auto-submit somehow doesn't work and keeps on refreshing the page without actually logging into it.

## Expected Behavior
There is already the menu "Site preferences" in the KeePassXC Brwoser Settings, which works nicely. I wish there were another option in the dropdown menu to ignore auto-submit on given pages.

## Current Behavior
At the moment you are unable to disable auto-submit for specific websites.

## Possible Solution
Check page url and if "ignore-auto-submit" is set, dont use auto-submit for the login fields.

## Debug info

KeePassXC - 2.6.2 and 2.3.1
KeePassXC-Browser - 1.7.3
Browser: Mozilla Firefox 82.0

Thank you for developing such an amazing tool, that helps to speed up daily processes <3

varjolintu:
Fixes #1120.

### DIFF
--- a/keepassxc-browser/options/options.html
+++ b/keepassxc-browser/options/options.html
@@ -518,6 +518,7 @@
                             <select class="form-control form-control-sm" name="ignore" data-i18n="[title]optionsSitePreferencesSelect">
                               <option value="ignoreNothing" data-i18n="optionsSelectionNothing"></option>
                               <option value="ignoreNormal" data-i18n="optionsSelectionNormal"></option>
+                              <option value="ignoreAutoSubmit" data-i18n="optionsSelectionAutoSubmit"></option>
                               <option value="ignoreFull" data-i18n="optionsSelectionFull"></option>
                             </select>
                           </td>
--- a/.eslintrc
+++ b/.eslintrc
@@ -85,6 +85,7 @@
         "Icon": true,
         "isEdge": true,
         "isFirefox": true,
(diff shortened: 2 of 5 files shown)


keepassxc-browser/options/options.html
```diff
@@ -518,6 +518,7 @@ <h2 class="pb-3 mt-0" data-i18n="optionsSitePreferencesTab"></h2>
                             <select class="form-control form-control-sm" name="ignore" data-i18n="[title]optionsSitePreferencesSelect">
                               <option value="ignoreNothing" data-i18n="optionsSelectionNothing"></option>
                               <option value="ignoreNormal" data-i18n="optionsSelectionNormal"></option>
+                              <option value="ignoreAutoSubmit" data-i18n="optionsSelectionAutoSubmit"></option>
```
droidmonkey, line 521: I think we have outgrown a signal choice list 🤔
We can keep this change for now but we need to rethink this workflow.

droidmonkey (approved):

keepassxc-browser/options/options.html
varjolintu: I was thinking exactly the same. This is the maximum options we can have as a combo box. Anything extra would need a redesign.

PR #1280, opened 2021-03-24
Add credential sorting

### LINKED ISSUE #1217: Autocomplete sort order should take group into account
Currently there are two options for the sorting of the returned credentials:
- Sort matching credentials by title
- Sort matching credentials by username

Neither of these takes into account the group where the entry is stored. Whereas in the list of returned credentials, the entries are displayed as: [group] title (username) 
As I have multiple entries with the same title and similar usernames (e.g. when managing for multiple companies the same service, and having a normal and an admin user) I get now a list shown in quite a "random" order, because the entries listed start with the group name, so it is quite difficult to find the right entry. 

It would be great if there would be an option to take the group into account when sorting.
Current:
[groupC] theWebService (user1)
[groupA] theWebService ADMIN (user2)
[groupB] theWebService (user3)

WIth sort by group:
[groupA] theWebService ADMIN (user2)
[groupB] theWebService (user3)
[groupC] theWebService (user1)

e.g. have the two additional options:
- Sort matching credentials by title
- Sort matching credentials by username
- **Sort matching credentials by group & title**
- **Sort matching credentials by group & username**

varjolintu:
Adds credential sorting to the extension. This makes KeePassXC's sorting in Browser Integration tab redundant and it must be removed for 2.7.0.

This feature adds four different options for the user:
- Sort matching credentials by title
- Sort matching credentials by username
- Sort matching credentials by group & title
- Sort matching credentials by group & username

Option page now has a combo box under the Autocomplete option:
Screenshot 2021-03-24 at 17 23 35

Fixes #1217.

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -683,6 +683,10 @@
         "message": "Color theme",
         "description": "Theme selection header text."
     },
+    "optionsCredentialSortSelectionHeader": {
+        "message": "Sort matching credentials by",
+        "description": "Credential sort option header text."
+    },
     "optionsThemeSelection": {
         "message": "Select color theme",
         "description": "Theme selection title text."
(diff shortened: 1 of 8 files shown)


droidmonkey (commented):
Works great! Can we make the default Group and Title? It makes the most sense visually and the group sorting is not applicable if everything is in the same group. Win Win

varjolintu:
@droidmonkey Sure. I was thinking the same thing when trying different options. It really makes sense. Removing this option from the KeePassXC side seems to affect a bit to the domain match sorting so that side needs to be tested carefully.

droidmonkey:
understood, just highlight that when a PR is amde

droidmonkey (approved):

PR #1283, opened 2021-03-27
Make sure credentials are cleaned at tab switch

varjolintu:
Credentials are now cached in the background script. This can cause credentials to stay in the memory until database is locked or a tab is closed. Let's make sure all other credentials are cleared from other tabs after a switch.

The feature is using a timeout, which is a new option with default value of 10 seconds (range is 0-3600).


droidmonkey (approved):
Works really well, great security feature!

PR #1286, opened 2021-03-28
Fix version error message on options page

varjolintu:
Remove a previously hard-coded snapshot version from if-sentence.

### DIFF
--- a/keepassxc-browser/options/options.js
+++ b/keepassxc-browser/options/options.js
@@ -276,7 +276,7 @@ options.showKeePassXCVersions = function(response) {
     $('#tab-about span.kpxcVersion').text(response.current);
     $('#tab-general-settings button.checkUpdateKeePassXC:first').attr('disabled', false);
 
-    if (response.current.startsWith('2.6') || response.current === '2.5.3-snapshot') {
+    if (response.current.startsWith('2.6') || response.current.includes('snapshot')) {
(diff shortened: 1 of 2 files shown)


droidmonkey (approved):

keepassxc-browser/options/options.js
```diff
@@ -276,7 +276,7 @@ options.showKeePassXCVersions = function(response) {
     $('#tab-about span.kpxcVersion').text(response.current);
     $('#tab-general-settings button.checkUpdateKeePassXC:first').attr('disabled', false);
 
-    if (response.current.startsWith('2.6') || response.current === '2.5.3-snapshot') {
+    if (response.current.startsWith('2.6') || response.current.includes('snapshot')) {
```
droidmonkey, line 279: This is also wrong, we want to make sure it _at least starts with 2.6_. We need to convert the version string into a number and then compare it.

droidmonkey (changes requested):

keepassxc-browser/options/options.js
varjolintu: Ah, true.

droidmonkey:
I didn't realize we had a message already defined for this. Does that work with all versions of keepassxc?

droidmonkey (approved):

PR #1294, opened 2021-04-02
Fix autocomplete menu with keyboard

### LINKED ISSUE #1293: FireFox 87 can't use arrow-keys to select login
FireFox has just updated to v87, and an issue identical to the one below has started happening again.

https://github.com/keepassxreboot/keepassxc-browser/issues/1207

Symptoms are exactly the same, except, if I click in the login-box, then it shows credentials in a dropdown, and I can then select using the arrow keys after that.

varjolintu:
#1248 broke this partially. The Autocomplete menu cannot be opened from the keyboard. It only works if it has been opened at least once using a mouse. With multiple menus supported, the input field must always point to the target element.

Fixes #1293.

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -171,11 +171,12 @@ class Autocomplete {
         }
 
         const items = this.getAllItems();
+        const inputField = e.target;
         if (e.key === 'ArrowDown') {
             // If the list is not visible, show it
             if (items.length === 0) {
                 this.index = -1;
-                this.showList(this.input);
+                this.showList(inputField);
             } else {
                 // Activate next item
                 ++this.index;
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/autocomplete.js
```diff
@@ -197,23 +198,23 @@ class Autocomplete {
...
         } else if (e.key === 'Tab') {
             // Return if value is not in the list
-            if (this.input.value !== '' && !this.elements.some(c => c.value === this.input.value)) {
+            if (inputField.value !== '' && !this.elements.some(c => c.value === inputField.value)) {
                 this.closeList();
                 return;
             }
 
-            this.index = this.elements.findIndex(c => c.value === this.input.value);
+            this.index = this.elements.findIndex(c => c.value === input.value);
```
droidmonkey, line 206: Should this be `=== inputfield.value` ?

droidmonkey (changes requested):

keepassxc-browser/content/autocomplete.js
varjolintu: Yes. Fixed.

droidmonkey (approved):

PR #1317, opened 2021-04-29
Allow file:// protocol in site preferences

### LINKED ISSUE #1337: URLs start with file:/// should be ignored
When I login while viewing a local file, the extension prompts for saving password. If I choose never ask again, the extension breaks **for all sites** because URLs starting with `file:///` is unexpected, the label in **Settings** page explicitly says `The URL must start with either https://, http://, or ftp://`, but this URL doesn't get into the settings through that box so there is no check.

About the case after that URL got into the settings, the extension will fails with an exception, like this:
`initContentScript error:  TypeError: "file:///THE_PATH/*" does not have a valid host`. This exception prevents the extension from providing autocompletion.

## Expected Behavior

No prompt displayed when browsing local files.

## Current Behavior

A prompt for saving password is displayed when submit form even when browsing local files.

## Possible Solution

Just check `window.location` and cancel the display if the location starts with `file:///`

Open this form in Firefox

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title></title>
</head>
<body>
    <form>
        <input id="username" type="text">
        <input id="password" type="password">
        <input type="submit">
    </form>
</body>
</html>
```

Enter something for both fields and submit. The prompt (https://i.imgur.com/DRQmjQh.png) will be shown.

## Debug info
KeePassXC - 2.6.4
KeePassXC-Browser - 1.7.8.1
Operating system: ArchLinux
Browser: Firefox

stefansundin:
Hello.

I wanted to block the extension from running on all `file://` URLs, but found that I was unable to do so. So here's a PR to make some changes to allow this.

I was actually able to just edit the input field with the browser's developer tools to add `file://.*` to the `pattern` attribute. But since `file://*` is 8 characters, I wasn't able to add it. I was able to change the `minlength` attribute, but there's also a JavaScript check that is harder to bypass (that is why I am removing the JavaScript check in this PR, since the `manualUrl.validity.valid` check already covers it). The JavaScript check actually requires 11 characters to pass (`value.length > 10`).

I was able to add a slightly longer URL though, `file:///users/*`, which should cover most of the local files I am going to load anyway. But this made me think.. I can imagine that it may be nice to be able to ignore `http://*`, so I decided to reduce `minlength` to `5` to allow for this and perhaps other things that I haven't thought of. Maybe we should just remove `minlength` since the pattern will already ensure that the field is not empty?

Please let me know what you think! Thanks!

P.S. It may be time to remove `ftp://` soon. https://blog.mozilla.org/addons/2021/04/15/built-in-ftp-implementation-to-be-removed-in-firefox-90/

Fixes #1337.

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -992,7 +992,7 @@
         "description": "Label for adding site manually on Site preferences tab."
     },
     "optionsSitePreferencesManualAddHelp": {
-        "message": "The URL must start with either <code>https://</code>, <code>http://</code>, <code>file://</code>, or <code>ftp://</code> and must be at least 10 characters long.",
+        "message": "The URL must start with either <code>https://</code>, <code>http://</code>, <code>file://</code>, or <code>ftp://</code> and must be at least 5 characters long.",
         "description": "Help text for adding site manually on Site preferences tab."
     },
     "optionsSitePreferencesConfirmation": {
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -992,7 +992,7 @@
         "description": "Label for adding site manually on Site preferences tab."
     },
     "optionsSitePreferencesManualAddHelp": {
-        "message": "The URL must start with either <code>https://</code>, <code>http://</code>, or <code>ftp://</code> and must be at least 10 characters long.",
+        "message": "The URL must start with either <code>https://</code>, <code>http://</code>, <code>file://</code>, or <code>ftp://</code> and must be at least 10 characters long.",
         "description": "Help text for adding site manually on Site preferences tab."
     },
     "optionsSitePreferencesConfirmation": {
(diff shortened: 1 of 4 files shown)


stefansundin:
😂 Although now I'm getting this error on all websites, which I didn't notice since my entry did technically stop the icons from showing up on file:// URLs.

> TypeError: "file:///users/*" does not have a valid host

So need to add some more code there.

Maybe we should just blanket stop the extension from running on `file://` URLs instead? That may just be easier. 🙃

varjolintu:
> 😂 Although now I'm getting this error on all websites, which I didn't notice since my entry did technically stop the icons from showing up on file:// URLs.
> 
> > TypeError: "file:///users/*" does not have a valid host
> 
> So need to add some more code there.
> 
> Maybe we should just blanket stop the extension from running on `file://` URLs instead? That may just be easier. 🙃

I actually use a lot of `file://` pages for internal testing.

stefansundin:
Ok, I pushed two more commits. `file://` is now a special case and is handled separately. It only allows for an asterisk at the end, or if there is no asterisk then it will be an exact match.

Please let me know if you'd like any changes. Thanks!

stefansundin:
Found a couple of other minor issues and fixed them.

varjolintu:
Works great! Thank you for the contribution.

varjolintu (approved):

PR #1321, opened 2021-05-03
TypeError: identifier.contains is not a function

pykaso:
This PR solves a problem I experienced when trying to fill in my login details on the https://appstoreconnect.apple.com/login website.

It is a crash in the `kpxcSites.exceptionFound` method that causes autocomplete to fail. Specifically, `TypeError: identifier.contains is not a function`. In most calls, the type of `identifier` is DOMTokenList. An exception occurs when `string` is passed.

Since I don't have the necessary insight into this add-on, this is a simple fix that resolves only the wrong parameter type. 
With this fix autocomplete works fine.


droidmonkey:
This totally changes the function behavior and is not a fix

varjolintu:
I'll check that site and make a fix for it without breaking any functionality.

PR #1329, opened 2021-05-16
Call password generator directly if 2.7.0 or newer

varjolintu:
Supports launching KeePassXC's password generator popup directly from the extension. The extension's own password generator dialog still exists for backwards compatibility, but it will be removed later in some future release.
This feature will be supported only with KeePassXC 2.7.0 and newer.

Related KeePassXC PR: https://github.com/keepassxreboot/keepassxc/pull/6529

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -187,7 +187,10 @@ page.switchTab = async function(tab) {
     browser.contextMenus.update('fill_attribute', { visible: false });
 
     // Clears all logins from other tabs after a timeout
-    clearTimeout(page.clearCredentialsTimeout);
+    if (page.clearCredentialsTimeout) {
+        clearTimeout(page.clearCredentialsTimeout);
+    }
+
     page.clearCredentialsTimeout = setTimeout(() => {
         for (const pageTabId of Object.keys(page.tabs)) {
             if (tab.id !== Number(pageTabId)) {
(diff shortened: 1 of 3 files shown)


keepassxc-browser/background/page.js
```diff
@@ -187,7 +187,10 @@ page.switchTab = async function(tab) {
     browser.contextMenus.update('fill_attribute', { visible: false });
 
     // Clears all logins from other tabs after a timeout
-    clearTimeout(page.clearCredentialsTimeout);
+    if (page.clearCredentialsTimeout) {
```
droidmonkey, line 190: This seems like an unrelated change. Are there guarantees as to the life cycle of this object? This check implies to me that we've had trouble making sure it is valid.

keepassxc-browser/content/pwgen.js
```diff
@@ -339,3 +376,16 @@ const disableButtons = function() {
...
+    const response = await browser.runtime.sendMessage({
+        action: 'get_keepassxc_versions'
+    });
+
+    const result = await browser.runtime.sendMessage({
+        action: 'compare_version',
+        args: [ '2.7.0', response.current ]
+    });
+
+    return result || response.current === '2.7.0-snapshot';
```
droidmonkey, line 390: We should bake a snapshot check into the version check function instead.

droidmonkey (approved):

PR #1335, opened 2021-05-24
add namespace to `slidein` animation

mwcz:
I ran into a conflict on a site which also used a CSS animation named `slidein`, so I'd suggest this change or something like it to namespace the animation.  The `slidein` animation from this extension caused elements on the site to not appear.

### DIFF
--- a/keepassxc-browser/css/banner.css
+++ b/keepassxc-browser/css/banner.css
@@ -1,6 +1,6 @@
 div.kpxc-banner {
     animation-duration: 0.2s;
-    animation-name: slidein;
+    animation-name: kpxc-slidein;
     background-color: var(--kpxc-background-color);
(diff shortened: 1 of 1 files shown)


varjolintu:
Which site has this problem? Thanks for the contribution, btw :)

mwcz:
> Which site has this problem? Thanks for the contribution, btw :)

Happy to contribute!  The site is https://arcade.redhat.com, which I created, so I've already made the fix there as well.  But my concern is that this problem will probably occur in numerous other places because `slidein` is the name used in the MDN examples for CSS documentation here https://developer.mozilla.org/en-US/docs/Web/CSS/animation

That's how I wound up with the name `slidein`.

If you'd like an example page to test against, I'd be happy to create one for you.

varjolintu (approved):

PR #1381, opened 2021-07-22
Adds ability to resolve totp field by parent selector

noahtkeller:
In some cases I would have to open KP to get the TOTP because it wasn't detecting fields in SSO provider. This extends the TOTP search to check if the input has any preceding element with the specified selectors.

Screen Shot 2021-07-22 at 9 39 52 AM


droidmonkey:
Good idea!

varjolintu:
I like this one. Thank you!

noahtkeller:
Thanks, glad to contribute.

Issue #1250, opened 2021-02-27
Autofill doesn't work for popup login boxes


throrfj:
## Expected Behavior
On the following website, there is a 'Login' button which opens a floating window. https://qsuper.qld.gov.au/
The Keypass icon shows in the Username field to offer autofill.
The username and password should fill when I click the Keypass icon.

## Current Behavior
Instead, when I click the icon, the floating window closes. Strangely, the Keypass icon continues to show in the same spot on the underlying page - where the login field would be if the floating box was still open. Clicking it again does nothing either.

## Debug info
KeePassXC - 2.6.4
KeePassXC-Browser - 1.7.6
Operating system: Win64
Browser: Chrome/Chromium 88.0.4324.182

varjolintu:
You can use context menu as a workaround until this is solved. Right-mouse click on the username field -> KeePassXC-Browser -> Fill Username and Password.

shoeper:
At least on https://id.atlassian.com/ not even "Fill password only" works.

varjolintu:
A Finnish site https://www.veikkaus.fi/ has the same problem also.

varjolintu:
This is caused by the click to the Autocomplete Menu item or Username Icon. The web page thinks we are clicking outside the floating window/panel, thus closes it.

varjolintu:
Now tracked in #1358.

PR #1388, opened 2021-08-03
Fix filling dynamic popups

varjolintu:
Use `stopPropagation()` to prevent event bubbling to the main DOM. These events should be only handled inside the extension.

Fixes filling for sites like https://qsuper.qld.gov.au/ and https://www.veikkaus.fi with dynamic popups/panels which normally close when they see a mouse click outside of the popup/panel.

Related (already closed) issue #1250.


droidmonkey (approved):
Nice! This will help with login workflow picker too eventually

PR #1390, opened 2021-08-08
Custom Login Fields banner selector

### LINKED ISSUE #454: "Choose custom login fields" creates one overlay per frame
## Expected Behavior
After clicking "Choose custom login fields for this page", I should be able to select the username field on https://appleid.apple.com

## Current Behavior
I am unable to select the username field on https://appleid.apple.com. Instead, the Info-screen is displayed twice.
grafik
After pressing <kbd>Esc</kbd>, one overlay (in the frame with the last click?) gets dismissed.

## Possible Solution
Only show the info-screen in the outermost frame.
Allow selection of input fields in all frames.

1. Open https://appleid.apple.com
2. Choose the option "Choose custom login fields for this page"
3. Press <kbd>Esc</kbd>

Alternatively:
1. Open https://appleid.apple.com
2. Click into username field
3. Choose the option "Choose custom login fields for this page"
4. Press <kbd>Esc</kbd>
5. Notice that the overlay of the inner frame gets dimissed.

## Debug info
KeePassXC - 2.4.0 
KeePassXC-Browser - 1.3.3 
Operating system: MacIntel 
Browser: Chrome/Chromium 74.0.3729.28

### LINKED ISSUE #670: Keep site interactible while choosing custom login fields
## Expected Behavior

Many websites have an SSO login flow, so the email input is on a different screen than the password input. I should be able to move along in this flow when I choose custom login fields.

## Current Behavior

The site becomes non interactable, which makes it impossible to set both username and password fields for autofill.

## Possible Solution

Maybe add a button to the window which toggles between the input selection mode and the normal webpage view?

1. Go to amazon.com, google.com or any website which uses an SSO login flow
2. Open the KeePassXC browser extension and click on "choose custom login fields"
3. Try to select both the right username as well as the right password field

## Debug info

KeePassXC - 2.5.0
KeePassXC-Browser - 1.5.3
Operating system: Win32
Browser: Mozilla Firefox 70.0

### LINKED ISSUE #1149: Two steps credentials plus custom fields (better use of "KPH: inputName")
Here is a link to page 1 of credential. You won’t be able to see page 2 if you don’t have an account in this administration (sorry for the inconvenience) : [https://wwwd.caf.fr](https://wwwd.caf.fr/wps/portal/caffr/login/!ut/p/a1/04_Sj9CPykssy0xPLMnMz0vMAfGjzOID_A3dPbyDDdz9A1yNDTxdzQNDXJ19DS0CjYAKIoEKDHAARwNC-sP1o8BK8JhQkBthkO6oqAgArtbX2Q!!/dl5/d5/L2dBISEvZ0FBIS9nQSEh/#/authmotdepasse)

Page 1 is currently properly filled by extension with this values (according to their alphabetical order and when _choose own credential fields_ is set)
image

This is page 2 with password input undetected because of custom credentials saved in browser storage, I guess.
image

Username html code:
image

## Current Behavior
_Domain (https://wwwd.caf.fr/) added with **detect username only** option with no success._

Currently, with _choose own credential fields_ button I can set _**keepassxc-browser**_ and **_keepassxc_** to fill username + KPH in page 1. But password input in page 2 (same url) won’t be detected and cant not even be filled via context menu. 

If I don’t customize _own credential fields_, extension won’t detect the username in page 1 (because exotic input name: **_numeroAllocataire_** I guess) but will detect password (because of its type I guess) in page 2.

So no way to have them both autofilled/detected. Either one or the other.

## Expected Behavior
Extension should always check `KPH: inputName` (that matches the input html code instead of using them alphabetically) and fill any fields that match **inputName** and url. There should be no need to play with _choose own credential fields_ button. Anyway, not until it’s not able to create/update proper `KPH: inputName` in database instead of using browser's localStorage that user can not edit.

And this auto read `KPH: inputName` shouldn’t prevent two steps credentials.

### Manual edition of entry, add additional attributes.
* Save custom field in `KPH: inputName`.
* Even save the username and password whose value would be {username} or {password} but protected. (Would it be a security breach ? I hope not.)
* Of course, password should only fill input with `type="password"`

image
image

* Add a setting section in extension to custom input name include/exclude list like **Kee - Password Manager** extension. This would make the `KPH: inputName` for username non-mandatory.
image

## Debug info

KeePassXC - 2.6.2
KeePassXC-Browser - 1.7.3
Operating system: Win32
Browser: Vivaldi 3.5.2115.81 (Stable channel) (64 bits) (Chromium 87.0.4280.107)

### LINKED ISSUE #1406: keepass xc mozilla plugin don't  work for grafana username and password field
Hello,

i am using "Grafana v5.3.2 (commit: 0d821d0)" . Username and password not filling auto. I cant define or re define username and password fields

I am sure url in db ( i have a lot of addresses )
https://pasteboard.co/uVU69sNXIpMk.png

When I drag the cursor over the boxes, no blue boxes or selection boxes appear. 
https://pasteboard.co/OPYvIMcWIeZe.png

Operating System: macOS Bigsur 11.6
Browser : Mozilla Firefox 92.0.1 (64 bit)
Plugin vers: .7.9.1 (22-07-2021)

KeePassXC - Version 2.6.6
Revision: 386b79a

Qt 5.15.2
Debugging mode is disabled.

Operating system: macOS 11.6
CPU architecture: x86_64
Kernel: darwin 20.6.0

Enabled extensions:
- Auto-Type
- Browser Integration
- SSH Agent
- KeeShare (signed and unsigned sharing)
- YubiKey
- TouchID

Cryptographic libraries:
- libgcrypt 1.9.3

varjolintu:
Replaces the old Choose Custom Fields implementation with a backdrop and replaces it with a new banner solution that also keeps the page interactive.
It's now possible to select single fields that are added to the current settings. This ensures it's possible to add for example a TOTP field without removing username/password selections. Using the Custom Fields selector in the "old" way still rolls it from the beginning to the end: Username -> Password -> TOTP -> String Fields.
Keyboard shortcuts still work, but the information/help dialog is only visible when hovered on the question mark icon.

When new fields are detected when there's some interaction with the page, possible fields for selection should be updated.

Iframes are also supported.

One good page for testing this: https://ib.airbank.cz/

Good sites for testing the banner with iframes:
https://appleid.apple.com/#!&page=signin
https://danskebank.se/privat/kundservice?n-login=pbmobid
https://www.bbva.com.ar/

Fixes #454.
Fixes #670.
Fixes #1149.
Fixes #1406.

A normal state:
Screenshot 2022-05-29 at 20 52 57

After a username selection it changes to:
Screenshot 2022-05-29 at 21 03 05


varjolintu:
This probably doesn't work with the page interactive the whole time, but you'll need to dismiss the banner (or apply changes) and then open it again. I'm thinking a way to keep it more interactive, or at least keep the banner open when selecting single fields, so the Apply doesn't close it.

varjolintu:
Founds some more bugs. Fixing..

varjolintu:
Moved to 1.8.0. I want to release 1.7.10 asap.

varjolintu:
Rebased.

droidmonkey:
Looking at this now, I am not a fan of the Google colors on the buttons. Perhaps just make the username/password/etc buttons a gray color or uniform muted color.

Some more thoughts:
1. Instead of "Again" should be "Reset"
2. "Show more" of what? This button doesn't lend itself it understanding what it does.
3. "Skip" this button should be to the left of "Confirm". But with this bar what are we skipping? I think this button is not necessary anymore.

varjolintu:
@droidmonkey It's too M&M yeah. Gotta tune the colors.

About the suggestions:
1. Good idea.
2. Maybe we should just show all possible stuff by default and remove the whole More button.
3. We are skipping the normal/familiar selection flow here: username -> password -> TOTP -> String Fields.
4. "Discard" sounds good.

droidmonkey:
For 3, if we are still in a linear workflow why do we have persistent choices of username/password/totp? Will have to play with the implementation to see how it works.

varjolintu:
@droidmonkey Those selections are for single selections, mainly for those second page password/TOTP fields that can now be selected after a username has been inserted because the page can be kept interactive. I'm all open for better ideas that still keeps the "normal" selection flow intact but allows the single selections too.

droidmonkey:
Maybe ditch the workflow and just let the user pick and choose. Click username, enter pick mode, pick username field, username button goes into "selected already" display. If you click the username button again it resets the selection for that field and enters pick mode again.

varjolintu:
That could work + "Show more" options are always enabled. In some way I still like the interactive method. Gotta think this through.

varjolintu:
Working on this now. Some more things that needs to be fixed/added:
- The selections should now be "global" and all previous selections should be visible (and non-selectable) unless the banner is closed or selections are confirmed
- If Custom Login fields are already chosen for the site, opening the banner should show the current selections that are set instead of blank selections

droidmonkey:
Nice will give it a drive

varjolintu:
No idea when the commit is ready, so hold your horses..

droidmonkey:
One thing is the keyboard shortcuts prevent me from interacting with the forms (like typing an email or password) to advance the process. The shortcuts should probably require pressing CTRL or ALT or remove them all together. 

Also, it would be nice to be able to hide the picker overlays so you can advance the login form without them in the way.

varjolintu:
I'm going to remove the keyboard shortcuts, except for number selection and closing the banner.

varjolintu:
@droidmonkey Please check the current implementation if you have time :) Thanks.

droidmonkey:
Looks great, but some functionality issues:

1. If you choose a field but don't click on an input area then you cannot confirm. You basically can't cancel choosing a field, you have to start over again.
2. When in field choosing sequence you cannot type any numbers or shift+numbers. 
3. There is no indication in the top bar when a field has already been chosen for the site. The original button colors remain even if the field is defined.

varjolintu:
> Looks great, but some functionality issues:
> 
>     1. If you choose a field but don't click on an input area then you cannot confirm. You basically can't cancel choosing a field, you have to start over again.
> 
>     2. When in field choosing sequence you cannot type any numbers or shift+numbers.
> 
>     3. There is no indication in the top bar when a field has already been chosen for the site. The original button colors remain even if the field is defined.

Made the changes.
1. You can now cancel the selection process with <kbd>Esc</kbd> or by pressing the button again, or some other button. Confirm status should change accordingly.
2. I can type numbers without any problems with all browsers I'm using for testing. Shift+numbers are not used currently.
3. The button colors now change to gray if there's already some selected. The status should stay the same even if the selection was made on the previous "page".

droidmonkey:
I'll resume testing this after we merge other bigger PR's

droidmonkey:
This works better but still suffers from the problem that I cannot actually complete the login process while picking the fields... so it is somewhat impossible to do multi-page selections. Add a button next to "Reset" to go back to "Interactive Mode" or similar. This would hide the selection fields but still remember what you selected.

varjolintu:
"Confirm" should remember what you have selected, even if you must pick another field later using the banner. In that case the "Clear saved data" is shown. Maybe "Reset" should be changed to something that could describe the behavior better. But it's actually clearing the current selections and starting over, not clearing any previously saved data. I'd like to keep those things separate.

droidmonkey:
For reset: my view is that if I am selecting new custom fields I don't need to delete ones I already saved. Or another way to put it, i wouldnt go to select new custom fields to delete the saved data. I can just hit reset and confirm if thats my intention. You delete saved custom fields in the extension settings. 

Maybe I have bias from the current process, but hitting confirm seems final to me and won't let me continue on the next page (like I will be starting over). Maybe add "Save and continue" which keeps the bar visible but the selections not overlayed so you can perform the login. Confirm could be changed to simply "Finish" which saves and hides the selection bar. 

Is that possible or am I overthinking this?

varjolintu:
"Clear saved data" is useful (at least for me) because it shows the user this page already has some previously saved data, and new selection will override those. So there's a possibility to clear all of them before selecting new ones. It's also good-to-have when user closes the banner and selects some other field later.

"Save and continue" could work a little better, or just "Save".

droidmonkey:
Just 'save' is fine. You could also indicate with a text label in the middle that there are saved fields already, could say "Editing existing fields" and when you press reset it clears that label and clears the fields. If you then press 'save' it would clear it for real in the settings. That workflow makes slightly more sense to me given the reason I have the selection bar visible at all. Again, maybe overthinking this (I'm also trying to reduce the number of buttons). @phoerious thoughts?

phoerious:
I'm not quite sure I follow what the problem is, correct me if I misinterpret things. 

To me, "Dismiss" is very ambiguous. What does the button do, will I lose data? I think when there is saved data, the fields should simply be marked already, so I can delete them individually. As a convenience feature, we can offer a "Delete all" button in that case. The other buttons should be "Save" and "Cancel".

droidmonkey:
Reset is equivalent to Delete All

varjolintu:
Updated the screenshot. It actually has a "Close" button which changes to "Cancel" button when some input field has been selected.

droidmonkey:
I like that!

phoerious:
The buttons are way too colourful. There is no visual hierarchy or consistency. Everything competes for the user's attention and the colours lose all meaning. The only coloured buttons should be Cancel and Save.

droidmonkey:
I think the username / password / totp buttons could use an icon to give them some pop and visual distinction. They need a color or indicator of some form to show "this one has been configured already"

varjolintu:
The buttons are colourful because the color describes the selection. With some pages the selection text is not visible due the styling of the input field, so the selected input field's color is the only thing for the user to see what selections have been made. Icons would definitely be a plus to the buttons.

I know those buttons are a bit candy store stuff, but those are purely because of some visual feedback. In my opinion there is a consistency, because the button and selection colors on the page are identical.

varjolintu:
> I would prefer a floating icon on the input field instead of different colours.

Doable, of course. This would be mean we'll have to copy each icon from Fork Awesome as SVG or make new ones, and add them to `web_accesible_resources`. I don't like the idea to inject the whole Fork Awesome CSS to each content script. Even adding the icons to the buttons will have to go through this process.

droidmonkey:
I'm not a fan of changing the current selection method in this PR (ie use floating buttons). Let's just fix the selection workflow itself first, then we can think about how to make selecting easier or more obvious.

droidmonkey:
Go here: https://www.w3schools.com/html/html_forms.asp

Scroll down even by a smidge, then try to choose fields... notice that the choice boxes are offset from the fields.

Also, it is hard to see the choice selection boxes (they are too transparent). I am actually starting to think that a combination of the selection boxes and an icon next to the box (indicating username/password/totp/string) would be good. I am not clear what the numbers mean in the box.

Finally, when I save custom fields... then go to pick them again.... my saved fields are not represented on the page, it pretends like I never saved them in the first place.

varjolintu:
@droidmonkey Yeah. Still need to fix many things. Without the old overlay the selection boxes are not visible enough. Hopefully the feature itself is still working as it should.

varjolintu:
@droidmonkey The number were meant for keyboard selection, but because we want to keep to page interactive, maybe it's best to remove those. Instead of icons we could user the same Username/Password/TOTP texts instead of numbers for non-selected fields when the selection step is active.

varjolintu:
The latests commits include the following changes:
- Detects background color of the input fields and adjusts colors accordingly for better contrast.
- Deletes keybindings from numbers. We want to keep the page interactive. Escape is still in use.
- Selection boxes now show the current selection type (Username/Password/TOTP/String field) instead of numbers.
- Selection boxes are now positioned correctly even after page scroll, or when scrolling.

keepassxc-browser/background/client.js
```diff
@@ -338,6 +338,8 @@ function onDisconnected() {
 }
 
 keepassClient.onNativeMessage = function(response) {
+    console.log(JSON.stringify(response));
```
droidmonkey, line 341: Did you mean to include this?

keepassxc-browser/content/ui.js
```diff
@@ -7,15 +7,20 @@ const MIN_INPUT_FIELD_OFFSET_WIDTH = 60;
 const MIN_OPACITY = 0.7;
 const MAX_OPACITY = 1;
 
-let notificationWrapper;
-let notificationTimeout;
+const BLUE_BUTTON = 'kpxc-button kpxc-blue-button';
+const GREEN_BUTTON = 'kpxc-button kpxc-green-button';
+const ORANGE_BUTTON = 'kpxc-button kpxc-orange-button';
+const RED_BUTTON = 'kpxc-button kpxc-red-button';
```
droidmonkey, line 13: Shouldn't these be defined in the custom fields picker file?

droidmonkey (approved):

keepassxc-browser/background/client.js
varjolintu: Nope. Removed.

keepassxc-browser/content/ui.js
varjolintu: Yes. Moved those.

Uchedanieln (commented):
Optional

Uchedanieln (commented):
Optional

keepassxc-browser/manifest.json
```diff
@@ -124,6 +124,7 @@
     },
     "web_accessible_resources": [
         "icons/disconnected.svg",
+        "icons/help.svg",
         "icons/keepassxc.svg",
         "icons/key.svg",
         "icons/locked.svg",
```
Uchedanieln, line 130: Approved

keepassxc-browser/content/username-field.js
```diff
@@ -111,7 +111,7 @@ UsernameFieldIcon.prototype.createIcon = function(field) {
 };
 
 const iconClicked = async function(field, icon) {
-    if (!kpxcFields.isVisible(field)) {
+    if (!kpxcFields.isCustomLoginFieldsUsed() && !kpxcFields.isVisible(field)) {
         icon.parentNode.removeChild(icon);
```
Uchedanieln, line 115: - [ ] Approved

keepassxc-browser/icons/help.svg
```diff
@@ -0,0 +1 @@
+<svg viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:serif="http://www.serif.com/" style="fill-rule:evenodd;clip-rule:evenodd;stroke-linejoin:round;stroke-miterlimit:2"><g transform="matrix(0.983029,0,0,0.983029,-2.15274,-4.69129)"><circle cx="26.604" cy="29.187" r="20.345" style="fill:#0090ff"/></g><g transform="matrix(31.8224,0,0,31.8224,16.3008,35.374)"><path d="M.169-.218C.169-.263.175-.3.186-.327.197-.354.217-.38.247-.407.276-.433.296-.454.306-.471.315-.487.32-.505.32-.523.32-.578.295-.605.244-.605.22-.605.201-.598.186-.583.172-.568.164-.548.164-.522H.022C.023-.584.043-.633.082-.668.122-.703.176-.721.244-.721.313-.721.367-.704.405-.671.443-.637.462-.59.462-.529.462-.501.456-.475.443-.451.431-.426.409-.399.378-.369L.339-.331C.314-.307.3-.28.296-.248l-.002.03H.169zm-.014.15C.155-.09.163-.108.177-.122.192-.136.211-.143.234-.143S.276-.136.291-.122c.015.014.022.032.022.054C.313-.047.306-.029.292-.015.277-.001.258.006.234.006.211.006.191-.001.177-.015.163-.029.155-.047.155-.068z" style="fill:#fff;fill-rule:nonzero"/></g></svg>
```
Uchedanieln, line 1: - [ ] Optional approved []()

keepassxc-browser/content/define.js
```diff
@@ -1,496 +0,0 @@
-'use strict';
```
Uchedanieln, line 1: - [ ] Approved

Uchedanieln (commented):
- [ ] Optional approval

PR #1396, opened 2021-08-21
Fix segmented TOTP with text type

varjolintu:
Allow segmented TOTP fields if `type` is `text` and input fields has `maxLength` set to 1.

Fixes TOTP with for example https://app.hackthebox.eu/login.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -419,6 +419,7 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
         // Use the form's elements
         addTotpFieldsToCombination(form.elements);
     } else if (inputs.length === 6 && inputs.every(i => (i.inputMode === 'numeric' && i.pattern.includes('0-9'))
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -419,6 +419,7 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
         // Use the form's elements
         addTotpFieldsToCombination(form.elements);
     } else if (inputs.length === 6 && inputs.every(i => (i.inputMode === 'numeric' && i.pattern.includes('0-9'))
+                || i.type === 'text' && i.maxLength == 1
```
droidmonkey, line 422: Missing ( ) around this logic block

droidmonkey (changes requested):

keepassxc-browser/content/keepassxc-browser.js
varjolintu: Too much C# these days.. fixed :)

droidmonkey (approved):

PR #1404, opened 2021-09-12
Switch update check default interval to Never

### LINKED ISSUE #1400: Don't check for updates by default
We deploy Firefox with some default addons (like keepassxc-browser) using group policies.
For more information on how to deploy addons see documentation at https://github.com/mozilla/policy-templates/blob/master/README.md#preferences

Keepassxc-browser's default behavior is to check for updates of keepassxc automatically. It is possible to set this check to "never" in the addon preferences. Unfortunately this is not possible in bigger environments where you deploy the addon via group policy (if you don't want to set the preference by hand on every client). This fact leads to unnecessary warning messages (and unnecessary helpdesk questions) when a keepassxc version is outdated.

In small business and enterprise environments it is not good practice (and often simply not possible) to install always the newest version in a short time period. The fact that keepassxc releases new versions quiet often doesn't make the situation easier.

Therefore I kindly ask to change the default update check setting to "never".

## Debug info

KeePassXC - 2.6.6
KeePassXC-Browser - 1.7.9.1
Operating system: Windows 10 controlled by Domain Controller
Browser: Tested with Firefox but others might face the same problem

varjolintu:
Fixes #1400.

### DIFF
--- a/keepassxc-browser/background/event.js
+++ b/keepassxc-browser/background/event.js
@@ -135,7 +135,7 @@ kpxcEvent.onCheckUpdateKeePassXC = async function() {
 };
 
 kpxcEvent.onUpdateAvailableKeePassXC = async function() {
-    return (page.settings.checkUpdateKeePassXC > 0) ? keepass.keePassXCUpdateAvailable() : false;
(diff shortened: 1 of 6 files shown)


droidmonkey (approved):
I agree with this since it essentially duplicates the update checker in keepassxc itself.

phoerious (approved):

PR #1429, opened 2021-10-18
Fix Apple login page

varjolintu:
Apple has made some changes that caused errors when trying to detect the password field. This should fix it.


alex-the-man:
I'm hitting this bug as well. Thanks for fixing.

droidmonkey (approved):

PR #1439, opened 2021-11-07
Add scrolling to autocomplete list

stefansundin:
This makes long lists easier to use.

Look, I have way too many Google accounts. Opening the autocomplete list adds a scrollbar to the window which makes the experience less than stellar. I think it is better to scroll inside the list itself.

Before:

Screen Shot 2021-11-07 at 13 03 44

With this change:

Screen Shot 2021-11-07 at 13 03 24

250px was not scientifically picked, so let me know if you prefer another value.

### DIFF
--- a/keepassxc-browser/css/autocomplete.css
+++ b/keepassxc-browser/css/autocomplete.css
@@ -1,5 +1,7 @@
 .kpxcAutocomplete-items {
     position: absolute !important;
+    max-height: 250px;
+    overflow-y: auto;
 }
 
 .kpxcAutocomplete-items div {


varjolintu:
`max-height` could be also relative to the page/frame size?

droidmonkey (approved):

stefansundin:
> `max-height` could be also relative to the page/frame size?

Good idea, although I'm not sure how to do it. Would also make sure there's a min height in that case, in case the input field is very close to the bottom.

varjolintu:
> > `max-height` could be also relative to the page/frame size?
> 
> Good idea, although I'm not sure how to do it. Would also make sure there's a min height in that case, in case the input field is very close to the bottom.

I think is OK for now. Let's fine-tune it later if necessary.

varjolintu (approved):

PR #1440, opened 2021-11-07
Settings page: load the theme earlier to prevent a white flash

stefansundin:
Load the colorTheme earlier to prevent a white flash when the Settings page is opened. This requires storing the theme somewhere where we can load it immediately before the first render (I picked localStorage for simplicity).

My setup is: OS in light mode, KeePassXC-Browser in dark mode. I have only tested this in Chrome.

With this setup, there's a white flash when the settings page is opened up (test it by refreshing the settings page, which makes it very evident). The browser renders the page before the JavaScript has started running. A workaround is to have an inline script that blocks rendering and can apply the style before the first render. I had to add a `content_security_policy` for the inline script to run.

If someone can think of a more elegant way of doing this then please let me know and I can update this branch.

If this hack is a bit too hacky then feel free to close the PR. Since I had a working fix, although not the prettiest, I figured it doesn't hurt to open the PR to start a discussion about this issue.

### DIFF
--- a/keepassxc-browser/manifest.json
+++ b/keepassxc-browser/manifest.json
@@ -146,6 +146,7 @@
         "http://*/*",
         "https://api.github.com/"
     ],
+    "content_security_policy": "script-src 'self' 'sha256-UD5WN2QEhRDEZ6vkLdHKFlkHwzbemzW2ppJiOLooI8o='; object-src 'none'",
     "applications": {
        "gecko": {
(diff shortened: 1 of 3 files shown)


varjolintu:
I tried a few tricks but this seems to work fine. Loading the setting from localStorage at the beginning of `options.js` did not work either.

varjolintu (approved):

PR #1441, opened 2021-11-07
Sort the page uuid first in the list

stefansundin:
Sort the page uuid first in the list. This is useful for pages that make you enter the username on one page and then the password on a second page (e.g. Google). You do not have to search for the same entry in the list again since it will conveniently be placed on top.

Since there are existing settings for sorting this list, perhaps a setting should be introduced for this behavior?

I've noticed that the TOTP list only lists one credential. I don't know exactly how that works under the hood, but perhaps it would be useful to only list the selected credential in this case too (but only when the password field is autocompleted)?

Please let me know if there's anything I should change.

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -1,8 +1,8 @@
 'use strict';
 
 const defaultSettings = {
-    afterFillSorting: SORT_BY_DEFAULT,
-    afterFillSortingTotp: SORT_BY_RELEVANT,
+    afterFillSorting: SORT_BY_MATCHING_CREDENTIALS_SETTING,
+    afterFillSortingTotp: SORT_BY_RELEVANT_ENTRY,
     autoCompleteUsernames: true,
     showGroupNameInAutocomplete: true,
     autoFillAndSend: false,
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -2,7 +2,7 @@
 
 const defaultSettings = {
     afterFillSorting: SORT_BY_DEFAULT,
-    afterFillSortingTotp: SORT_BY_DEFAULT,
+    afterFillSortingTotp: SORT_BY_RELEVANT,
     autoCompleteUsernames: true,
     showGroupNameInAutocomplete: true,
     autoFillAndSend: false,
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const defaultSettings = {
+    afterFillSorting: SORT_BY_DEFAULT,
+    afterFillSortingTotp: SORT_BY_DEFAULT,
     autoCompleteUsernames: true,
     showGroupNameInAutocomplete: true,
     autoFillAndSend: false,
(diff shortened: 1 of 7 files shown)


droidmonkey:
Concur on the TOTP "list the relevant entry only" approach.

varjolintu:
Need to test this with a case where TOTP is retrieved from another database.

stefansundin:
@varjolintu Feel free to modify this branch as needed. Let me know if you want me to change anything. Thanks! :)

varjolintu:
I think a separate setting for this could be nice. TOTP could have its own section. Let's move this PR to the next release for now.

stefansundin:
I rebased this to fix the merge conflict. Haven't tested it again though.

varjolintu:
Added new sort options. The same could be used for #1477.
Screenshot 2022-05-09 at 17 26 06

The left margins are a bit disturbing. Maybe I should level those to the same position with other options. However, the Color Theme selector is also at the left border.

keepassxc-browser/content/autocomplete.js
```diff
@@ -105,7 +110,13 @@ class Autocomplete {
                 item.addEventListener('mousedown', e => e.stopPropagation());
                 item.addEventListener('mouseup', e => e.stopPropagation());
 
-                this.list.appendChild(item);
+                // If this page has an associated uuid and it matches this credential, then put it on top of the list
+                if (this.afterFillSort === SORT_BY_RELEVANT && c.uuid === pageUuid) {
+                    this.list.prepend(item);
+                }
+                else {
```
droidmonkey, line 117: Style fix needed

keepassxc-browser/background/page.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 const defaultSettings = {
+    afterFillSorting: SORT_BY_DEFAULT,
+    afterFillSortingTotp: SORT_BY_DEFAULT,
```
droidmonkey, line 5: Shouldn't TOTP sort by relevant by default?

keepassxc-browser/background/page.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 const defaultSettings = {
+    afterFillSorting: SORT_BY_DEFAULT,
```
droidmonkey, line 4: What does "Sort By Default" mean? Need to make this explicit on what we are sorting by.

droidmonkey (changes requested):

droidmonkey:
I am not sure why you would ever _not want_ to have the subsequent password and TOTP fields sorted by relevant?

keepassxc-browser/background/page.js
varjolintu: The help text under the option says: "When Relevant credential first option is selected, the credential for previously filled login is set as first in Autocomplete Menu's listing." Should we add some clarification what the default is, if that is not obvious?

keepassxc-browser/background/page.js
varjolintu: Sure. Let's do that.

keepassxc-browser/background/page.js
droidmonkey: No I meant make it explicit in code. SORT_DEFAULT  is meaningless. Needs to have a better name.

varjolintu: Got it.

varjolintu: Not sure about the variable name. When default, everything is sorted by the "Sort matching credentials by" setting, not alphabetical or natural order. Maybe `SORT_BY_MATCHING_CREDENTIALS` could describe it better.

droidmonkey (approved):

varjolintu:
Conflicts resolved. Had to check both options.

droidmonkey:
Should be good to merge

PR #1447, opened 2021-11-17
Fix interfering mouse events when dragging content

### LINKED ISSUE #1446: Google calendar issue after KeepassXC firefox extension update
Starting today I am unable to drag google calendar appointments to new days/times. I followed Firefox's troubleshooting procedure and found that the KeepassXC extension for Firefox was causing the issue, and it seems the extension was updated today.

varjolintu:
Broken by https://github.com/keepassxreboot/keepassxc-browser/pull/1388. Seems `stopPropagation` interferes with page content. Just keeping it with `mouseUp` event works fine. Let's fix this asap.

Fixes #1446.

### DIFF
--- a/keepassxc-browser/content/ui.js
+++ b/keepassxc-browser/content/ui.js
@@ -274,8 +274,6 @@ document.addEventListener('mousemove', function(e) {
         return;
     }
 
-    e.stopPropagation();
-
     if (kpxcPasswordDialog.selected === kpxcPasswordDialog.titleBar) {
         const xPos = e.clientX - kpxcPasswordDialog.diffX;
         const yPos = e.clientY - kpxcPasswordDialog.diffY;
@@ -302,7 +300,6 @@ document.addEventListener('mousedown', function(e) {
         return;
     }
 
-    e.stopPropagation();
     kpxcUI.mouseDown = true;
 });
 
@@ -311,7 +308,6 @@ document.addEventListener('mouseup', function(e) {
         return;
     }
 
-    e.stopPropagation();
     kpxcPasswordDialog.selected = null;
     kpxcDefine.selected = null;
     kpxcUI.mouseDown = false;


droidmonkey (approved):

droidmonkey:
We are likely to break something by blocking the mouse up as well. We shouldn't do any ops like that at the document or window level.

varjolintu:
> We are likely to break something by blocking the mouse up as well. We shouldn't do any ops like that at the document or window level. The scoped blocks when clicking on the icon or selection list are much better.

I'll remove that one too before releasing a fix.

tomasklaen:
Don't prevent or stop propagation of any default events on window or any global element!

And please release this ASAP, it's breaking everything that relies on mouseup/down events and users are complaining. Had to spend a lot of time figuring out it's this extension.

varjolintu:
@tomasklaen Just released an update. Sorry for the trouble.

jamz216:
> @tomasklaen Just released an update. Sorry for the trouble.

Thank you!

varjolintu:
In a perfect world you could assume that any event modifications to global elements would be scoped _only_ inside the extension, not actually affecting outside of it. Sadly that's not the case, and mistakes like this can happen.

droidmonkey:
I certainly consider this a bug in the browser. It should not be possible for an extension to cause this disruption.

killroy42:
Thanks for sorting it out. Was scratching my head for a bit. My affected site is/was app.crackingthecryptic.com

varjolintu:
This did not go as planned with Chrome Web Store. I got en email hour ago which informed that they decided to take our extension down because it has an "excessive and / or irrelevant keywords" in the description. The violating content is "Fix".

Didn't know release notes were a bad thing. Oh well, in the future there's just a link to the changelog. New version is already submitted for re-view.

PR #1470, opened 2021-11-21
Add minimum font size to autocomplete menu

varjolintu:
Adds a minimum font size to the Autocomplete Menu. The default is still `.9em` but if it's smaller than `10px` we will use the latter. Fixes a site for example site https://www.teamblind.com/ where the font size is way too small to see properly.


droidmonkey (approved):

stefansundin:
Great! Wouldn't it be a good idea to also set the font family? Thanks!

varjolintu:
> 

Not sure what would be the right choice for that. I personally like that the font-family is identical with the site, which works with most cases.

PR #1472, opened 2021-11-21
Add support for downloading favicons after saving a credential

### LINKED ISSUE #86: Automatically download favicon for new entries
See https://github.com/smorks/keepasshttp-connector/issues/79

varjolintu:
Adds a new command parameter for supportin favicon download after a credential has been saved.
An implementation is needed to the KeePassXC side. Let's keep this WIP until then.

Related KeePassXC PR: https://github.com/keepassxreboot/keepassxc/pull/7179

Fixes #86.

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -14,6 +14,7 @@ const defaultSettings = {
     credentialSorting: SORT_BY_GROUP_AND_TITLE,
     defaultGroup: '',
     defaultGroupAlwaysAsk: false,
+    downloadFaviconAfterSave: false,
     redirectAllowance: 1,
     saveDomainOnly: true,
     showLoginFormIcon: true,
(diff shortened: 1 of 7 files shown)


keepassxc-browser/background/page.js
```diff
@@ -14,6 +14,7 @@ const defaultSettings = {
     credentialSorting: SORT_BY_GROUP_AND_TITLE,
     defaultGroup: '',
     defaultGroupAlwaysAsk: false,
+    downloadFaviconAfterSave: false,
```
droidmonkey, line 17: I didn't realize this was disabled by default and thought the code was broken. Perhaps we should enable this by default instead?

droidmonkey (approved):

keepassxc-browser/background/page.js
varjolintu: On the other hand I would assume users don't want to make any network connections from KeePassXC by default.

ChrTall:
I am using the current Browser Plugin, but can not find the option to enable it.

PR #1477, opened 2021-11-24
Autocomplete: username detection for Google

stefansundin:
Hello, it's me again. This is a little bit similar to #1441, but while that works for all websites, this change is specific to Google and the primary difference is that it helps in the case when you don't have to select the username at all.

While it is specific for Google right now, it can easily be extended to work with more websites.

Sometimes, for various reasons, Google forces you to confirm that you are you, and they do that by having you enter your password again. In this case you don't have to enter the username, which is why #1441 does not help in this case. They simply make you enter the password.

Since the username is on the page, we can use it to help the user by putting the correct entry on the top of the list.

google-profileIdentifier


varjolintu:
Any site specific code to `sites.js` if possible, yes please.

stefansundin:
Moved the code. 👍

keepassxc-browser/common/sites.js
```diff
@@ -65,6 +65,19 @@ kpxcSites.googlePasswordFormUrl = 'https://accounts.google.com/signin/v2/challen
...
+ * Tries to detect a username from the page.
+ * @returns {string}   Returns the detected username if there is one, undefined otherwise.
+ */
+kpxcSites.detectUsername = function() {
+  if (document.location.origin === 'https://accounts.google.com') {
+    const profileIdentifier = document.getElementById('profileIdentifier');
+    if (profileIdentifier) {
+      return profileIdentifier.textContent.trim();
+    }
+  }
```
varjolintu, line 78: There's no default return in this function.

stefansundin: JavaScript implicitly returns undefined in this case. I can add a return statement if that's preferred.

stefansundin:
I pulled the latest changes to fix the merge conflict. Haven't tested it again though.

keepassxc-browser/common/sites.js
```diff
@@ -69,6 +69,20 @@ const kpxcSites = {};
 kpxcSites.googlePasswordFormUrl = 'https://accounts.google.com/signin/v2/challenge/password';
 kpxcSites.savedForm = undefined;
 
+/**
+ * Tries to detect a username from the page.
+ * @returns {string}   Returns the detected username if there is one, undefined otherwise.
+ */
+kpxcSites.detectUsername = function() {
```
droidmonkey, line 76: Prefer this to be named `detectUsernameFromPage` (or similar) to be more explicit about where the information is coming from.

droidmonkey (changes requested):

keepassxc-browser/common/sites.js
stefansundin: Done.

stefansundin:
I pulled the latest changes from the develop branch, and this time I also re-tested this on the Google sign-in page and had to change some other things because that page no longer used the id `profileIdentifier`. But it works again now. 👍

droidmonkey (approved):

PR #1490, opened 2021-12-05
Show gray icon when no credentials are available

varjolintu:
Changes the icon to grayscale when there's no credentials to fill.


tofuSCHNITZEL:
This is great!

keepassxc-browser/content/username-field.js
```diff
@@ -47,6 +47,12 @@ class UsernameFieldIcon extends Icon {
         this.icon.classList.remove('lock', 'lock-moz', 'unlock', 'unlock-moz', 'disconnected', 'disconnected-moz');
         this.icon.classList.add(getIconClassName(state));
         this.icon.title = getIconText(state);
+
+        if (kpxc.credentials.length === 0 && state === DatabaseState.UNLOCKED) {
+            this.icon.style.filter = 'saturate(0%)';
+        } else {
+            this.icon.style.filter = 'saturate(100%)';
+        }
```
droidmonkey, line 55: Shouldn't this be shared amongst the three icon types (username, password, TOTP)? I guess especially shared with TOTP. In any case where there is not a credential to provide, the icon should go gray.

droidmonkey (changes requested):

keepassxc-browser/content/username-field.js
varjolintu: Fixed it. The Username Icon still has to implement a slightly different version of the comparison because otherwise the saturation would be 0% when database is locked, showing a gray lock icon.

droidmonkey (approved):

PR #1498, opened 2021-12-10
Improve polish translations.

mjamro:
Change translations to more commonly used and more intuitive.


droidmonkey:
This needs to be done over on Transifex. https://transifex.com/keepassxc/keepassxc

Issue #478, opened 2019-04-21
Filter username while typing in username field


sebast889:
It is currently possible to filter usernames using the icon popup but is there a way to do this while typing in the username field? It would be a easier way than always going to the popup first.

varjolintu:
That's currently missing. I'll fix that soon.

varjolintu:
Just giving an update: I've been working this for a while and it already works quite nicely. Still need to refactor a lot of code for it, so this will be delayed a little bit.

andreasmihm:
Any updates on this one? This missing feature is a blocker for me

varjolintu:
@andreasmihm Sorry but no. I ran to issues and the Autocomplete Menu has been refactored so I'll have to start from a scratch.

amrecords:
cool, when can I expect this to be released?

varjolintu:
I'm expecting to release 1.8.0 in mid-April or early May.

PR #1511, opened 2022-01-02
Autocomplete search feature

stefansundin:
This refactors the autocomplete list code quite a bit. Sorry I didn't split it into well defined commits but a big reason for that is these changes kinda snowballed as I went along. I kept finding things that I wanted to change.

I was mostly annoyed at the first item always being selected, which makes submitting user-entered text more annoying than it should be. I don't want the enter key to pick from the autocomplete list if the text I entered doesn't match. So this turns the text you enter into a search string that is used to find the element you want. It will even `scrollIntoView` the item if the list is scrolled. If the text you enter doesn't match any item, then it will deselect the item in the list and the enter key will not interfere.

Summary of changes:

- You can now type in the text field to "search" for the entry you want. Items starting with the text will be prioritized, but will fall back to search for any item that contains the text. This way you can search a substring if it is faster.
- If you continue typing and what you type is not a substring of any item, then it will automatically "deselect" the item in the list and allow you to submit using the enter key and it will submit what you typed instead of overwriting with the item that was selected.
- The auto-submit footer will now always be displayed, even if the list is scrolled.
- The mousemove event handler is now used to select an item with the mouse. This prevents an item from being selected if the mouse cursor is where the list appears when the list opens.
- The DOM elements are now kept around. This means that if the list is opened, then scrolled, then closed, and then opened again, the scroll position is kept. This also fixes leaking DOM elements since it was possible to continually recreate the list and its DOM elements while the old ones were not cleaned up properly.
- Remove all uses of `.select()` to select the contents of the text field.

I really like how the autocomplete list and the search field interacts after these changes. It just feels a lot better in my opinion. I hope you like it too. Let me know if you want any specific changes or have questions about certain changes.

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -9,15 +9,16 @@ class Autocomplete {
         this.elements = [];
         this.index = -1;
         this.input = undefined;
-        this.shadowRoot = undefined;
         this.wrapper = undefined;
+        this.shadowRoot = undefined;
+        this.container = undefined;
     }
 
     clear() {
         this.elements = [];
     }
 
-    async click(e, input) {
+    async click(e) {
 
     }
 
@@ -37,8 +38,9 @@ class Autocomplete {
         this.autoSubmit = autoSubmit;
 
         if (!this.autocompleteList.includes(input)) {
-            input.addEventListener('click', ev => this.click(ev, input));
-            input.addEventListener('keydown', ev => this.keyPress(ev));
+            input.addEventListener('click', e => this.click(e));
+            input.addEventListener('keydown', e => this.keyDown(e));
+            input.addEventListener('keyup', e => this.keyUp(e));
             input.setAttribute('autocomplete', 'off');
             this.autocompleteList.push(input);
         }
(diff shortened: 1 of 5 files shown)


varjolintu:
Thank you! This works great. However, I'm not sure why `.select()` is removed here. It causes the following problems:
- If credentials are filled from the context menu, username field is not selected -> user cannot just press <kbd>Enter</kbd> anymore to log in
- Same happens if user has filled credentials from the popup

The username field (password field if it's the only one) should stay selected after a fill so just pressing <kbd>Enter</kbd> is enough for logging in.

Btw, this solves #478.

stefansundin:
Thanks for testing it out. The reason why I removed `select()` was because I wanted the ability to click at different places in the text field and insert text, but since it was selecting everything it just got very hard. I think using `focus()` should solve the points you presented without changing the behavior that is expected from a normal text field.

stefansundin:
I actually was not able to reproduce either of the problems that you mentioned. It seems to be that filling from the popup isn't currently selecting any of the inputs in the form (tested released version on Firefox). I added some code to select the password or username field, please let me know if there's a better way of doing that.

And when using the context menu, by virtue of right clicking on the input field to use "Fill Username and Password", the right click itself selects the DOM element. Unless there's another way of using the context menu that I'm not aware of?

varjolintu:
@stefansundin I got the bug reproduced from context menu when there were multiple credentials. Triggering the Fill Username and Password shows the Autocomplete Menu, but after selection no input fields are focused/selected.

stefansundin:
I wasn't able to quite reproduce for some reason. But I think I added back the code necessary for it to work again.

varjolintu:
Now it seems to work like it should. Is there a reason it focuses to the password field now instead of username field? We should keep the current focus behavior. Otherwise this is ready to go :)

stefansundin:
Is this the focus line you are referring to?

https://github.com/stefansundin/keepassxc-browser/blob/2ba83bd3e24af9918bfbfce95cb0c28662fc7d33/keepassxc-browser/content/keepassxc-browser.js#L1114

I can change it to `combination.username.focus();`. Only problem with this is that it won't select the password field when you fill on sign-in where username and password are entered on separate pages (e.g. Google sign-in). So maybe just swapping the order makes more sense? `(combination.username || combination.password).focus();`

varjolintu:
@stefansundin Sounds good. Username field should be the first priority if found.

varjolintu:
Works good. Can you squash the commits? Thanks.

stefansundin:
I could, but I think it would be better if you use the Squash and Merge feature that GitHub provides. It will also let you edit the final commit message to say what you prefer.

I think that's better because it would preserve the commits and progression of the work here for anyone coming to look later.

varjolintu:
Oh yeah. I forgot that feature exists.

varjolintu (approved):

PR #1517, opened 2022-01-07
Uptated Italian Translation

MichelePezza:
Updated translation


droidmonkey:
These are done through transifex.

PR #1525, opened 2022-01-12
Fix Google password fill from popup

### LINKED ISSUE #1523: Autofill of password using extension popup does not work at accounts.google.com
## Expected Behavior
User should be able to fill password field using extension popup without having to refresh a page

## Current Behavior
Password field stays empty after described above action.

## Possible Solutions
1. Refresh page with password field using F5. After that autofill from popup starts working.
2. Use dropdown list 
Screenshot (2)

## Steps to Reproduce

1. Have multiple gmail accounts, log out of all of them
2. Head to https://mail.google.com
3. Click _Sign in_ in page header, page with username field opens
4. Click extension icon and select any account from filter
Screenshot (3)_LI

5. Proceed to password page
6. Try to repeat step 4

Password field should be filled but it's not.

## Debug info

KeePassXC - 2.6.6
KeePassXC-Browser - 1.7.11
Operating system: Win64-10
Browser: Vivaldi 5.0.2497.35 (Chromium 96.0.4664.137)/fresh Edge 97.0.1072.55

varjolintu:
Filling Google's login page password from the popup fails because the password field is actually in a recently detected combination. If there's more than one combination on Google's page, we must use the second one for a successful fill.

Fixes #1523.

### DIFF
--- a/keepassxc-browser/common/sites.js
+++ b/keepassxc-browser/common/sites.js
@@ -162,3 +162,16 @@ kpxcSites.formSubmitButtonExceptionFound = function(form) {
 
     return undefined;
 };
+
+/**
+ * Handles a few exceptions for certain sites where popup fill is not working properly.
+ * @param {object} form     Form element
+ * @returns {boolean}       True if exception found
+ */
+kpxcSites.popupExceptionFound = function(combinations) {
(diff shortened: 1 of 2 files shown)


droidmonkey (approved):
This is ugly but if it fixes a problem...

varjolintu:
Ugly pages need ugly fixes.

PR #1540, opened 2022-01-22
Add debug logging option

varjolintu:
Adds debug logging to the extension, mainly focusing on additional information on content scripts. Debug info contains all input fields detected on the page, plus the combinations sorted by the extension.

Option is disabled by default.

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -12,6 +12,7 @@ const defaultSettings = {
     clearCredentialsTimeout: 10,
     colorTheme: 'system',
     credentialSorting: SORT_BY_GROUP_AND_TITLE,
+    debugLogging: false,
     defaultGroup: '',
     defaultGroupAlwaysAsk: false,
     downloadFaviconAfterSave: false,
@@ -91,6 +92,10 @@ page.initSettings = async function() {
 
         if (!('credentialSorting' in page.settings)) {
             page.settings.credentialSorting = defaultSettings.credentialSorting;
+        }
+
+        if (!('debugLogging' in page.settings)) {
+            page.settings.debugLogging = defaultSettings.debugLogging;
         }
 
         if (!('defaultGroup' in page.settings)) {
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -153,7 +153,7 @@ page.initSettings = async function() {
         await browser.storage.local.set({ 'settings': page.settings });
         return page.settings;
     } catch (err) {
-        console.log('page.initSettings error: ' + err);
+        showErrorMessage('page.initSettings error: ' + err);
         return Promise.reject();
     }
 };
@@ -174,7 +174,7 @@ page.initOpenedTabs = async function() {
         page.currentTabId = currentTabs[0].id;
         browserAction.showDefault(currentTabs[0]);
     } catch (err) {
-        console.log('page.initOpenedTabs error: ' + err);
+        showErrorMessage('page.initOpenedTabs error: ' + err);
         return Promise.reject();
     }
 };
@@ -210,7 +210,7 @@ page.switchTab = async function(tab) {
 
     browserAction.showDefault(tab);
     browser.tabs.sendMessage(tab.id, { action: 'activated_tab' }).catch((e) => {
-        console.log('Cannot send activated_tab message: ', e);
+        showErrorMessage('Cannot send activated_tab message: ' + e.message);
     });
 };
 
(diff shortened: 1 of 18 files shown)


keepassxc-browser/common/global.js
```diff
@@ -163,3 +163,15 @@ function tr(key, params) {
 const trimURL = function(url) {
     return url.indexOf('?') !== -1 ? url.split('?')[0] : url;
 };
+
+const showDebugMessage = function(message, extra) {
+    console.log(`${EXTENSION_NAME} - ${message}`);
```
droidmonkey, line 168: Maybe add [Debug] in there?

keepassxc-browser/background/page.js
```diff
@@ -372,9 +377,15 @@ const createContextMenuItem = function({action, args, ...options}) {
...
             }).catch((err) => {
-                console.log(err);
+                showErrorMessage(err);
             });
         },
         ...options
     });
 };
+
+const debugLog = function(message, extra) {
```
droidmonkey, line 387: I think this function redirection is not worthwhile. Can we just have showDebugMessage and check for the setting being enabled in there?

droidmonkey (changes requested):

keepassxc-browser/background/page.js
varjolintu: The problem is that `page.settings` is not directly available for content scripts, unless we send a message to background script every time we write a debug message and check it. This is why content scripts use `kpxc.settings` for checking because it's already parsed on page load. And, vice versa, `kpxc.settings` is not directly available background scripts.

keepassxc-browser/common/global.js
varjolintu: Good idea.

keepassxc-browser/background/page.js
droidmonkey: Damn I figured that is why. It's just awkward to have two very different function names for logging errors vs debug.

varjolintu: Maybe those function names could be switched after all. The internal call should be `debugLog()` and the ones we call are `showErrorMessage()` and `showDebugMessage()`.

droidmonkey:
Might also want to add a danger tag to the debug logging because it may leak more details then the user expects.

keepassxc-browser/background/page.js
droidmonkey: show or log? I prefer log cause that's what you are doing. logError, logDebug

varjolintu: Let's use `log`.

keepassxc-browser/content/fill.js
```diff
@@ -135,11 +140,13 @@ kpxcFill.fillFromTOTP = async function(target) {
 // Fill TOTP with matching uuid
 kpxcFill.fillTOTPFromUuid = async function(el, uuid) {
     if (!el || !uuid) {
+        logDebug('Error: Element or uuid is empty');
         return;
     }
 
     const user = kpxc.credentials.find(c => c.uuid === uuid);
     if (!user) {
+        logDebug('Error: No user found with uuid: ' + uuid);
```
droidmonkey, line 149: Wouldn't this be "No entry found"?

keepassxc-browser/content/fill.js
```diff
@@ -166,6 +173,7 @@ kpxcFill.fillTOTPFromUuid = async function(el, uuid) {
 // Set normal or segmented TOTP value
 kpxcFill.setTOTPValue = function(elem, val) {
     if (kpxc.combinations.length === 0) {
+        logDebug('Error: Credential list is empty.');
```
droidmonkey, line 176: Given that the code+line is not printed, it would be very helpful to add context to these messages. (Is it possible to add file+line to the debug messages??)

Context for this one would be: Totp Fill :: No credentials found :: (optional why if known)

droidmonkey (approved):

keepassxc-browser/content/fill.js
varjolintu: The problem with these messages are that now when the log function is always the same, JavaScript Console doesn't show the line where the error is coming from.. true. The information must be added manually to these messages. There are some automatic options to retrieve the info, and I'm gonna try them first.

droidmonkey: Yah I don't think we need/want a stack trace and we don't want to manually hardcode a location. Worst case we just add a little context to give a hint as to where we are in the sequencing.

varjolintu: I added a new commit that also reports the file and line number. Just a simple parsing of the current stack. Seems there's no other way to do that, but I think that's good enough.

PR #1542, opened 2022-01-23
fix "Reload" button unaccessible from pinned popup

dym-sh:
## problem
when the extension is "pinned" to the overflow menu – the popup is constrained within the smaller window, hence not all elements on it are even visibly exist, let alone accessible:
before

and only visible+accessible after pressing [Tab]:
before onKey Tab

## solution
just **_don't_** align buttons to the right for "esthetic" reasons:

after

### DIFF
--- a/keepassxc-browser/popups/popup.html
+++ b/keepassxc-browser/popups/popup.html
@@ -30,12 +30,12 @@
       </div>
 
       <div id="initial-state">
-        <div class="loader"></div> <span data-i18n="popupCheckingStatus"></span>
+         <div class="loader"></div> <span data-i18n="popupCheckingStatus"></span>
       </div>
 
       <div id="not-configured" style="display: none">
         <p data-i18n="popupNotConfigured"></p>
-        <div style="text-align: right">
+        <div>
           <button id="connect-button" class="btn btn-sm btn-primary"><i class="fa fa-link" aria-hidden="true"></i> <span data-i18n="popupConnectButton"></span></button>
         </div>
       </div>
@@ -46,7 +46,7 @@
           <code id="need-reconfigure-message"></code>
         </p>
         <p data-i18n="popupNeedReconfigureMessage"></p>
-        <div style="text-align: right">
+        <div>
           <button id="reconnect-button" class="btn btn-sm btn-primary"><i class="fa fa-refresh" aria-hidden="true"></i> <span data-i18n="popupReconnectButton"></span></button>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


droidmonkey:
This looks like a bug in edge browser

varjolintu:
Firefox also has had a similar issue where the popup size is not respected.

dym-sh:
it is firefox, and this is supposedly a feature – where popups of pinned addons are limited by the enclosing submenu

varjolintu:
I would assume a proper fix for this should be to just set the popup width to automatic, but keep the maximum width intact. Then the content is just scaled like it should be.

dym-sh:
sure, that works too:

varjolintu:
@dym-sh That won't work. Now it breaks the normal popup, plus it looks completely different in Firefox.

dym-sh:
is this enough of a compromise?

upd_20220125062450

varjolintu:
@dym-sh I'd prefer if the layout doesn't change from the current one when normal popup is used. This PR should only solve problems when the popup is not wide enough for the content.

dym-sh:
tfw> vendors change their own defaults creating obscure edge cases, and then never address the arised issues

/ref https://bugzilla.mozilla.org/show_bug.cgi?id=1373490

varjolintu:
It' still not working with Firefox:
Screenshot 2022-01-25 at 19 37 18
Screenshot 2022-01-25 at 19 37 07

Not sure if it's even fixable due the Firefox bug.

dym-sh:
so remember how i simply removed the right-align style – lets just keep that?

you also can add this line to `popup.js` and see exactly how it differs on every fractional dpi setting: 302, 194, or sometimes 0. :hankey:

```diff
diff --git a/keepassxc-browser/popups/popup.js b/keepassxc-browser/popups/popup.js
index ef20c28..cc1aa27 100644
--- a/keepassxc-browser/popups/popup.js
+++ b/keepassxc-browser/popups/popup.js
@@ -111,5 +111,5 @@ $(async () => {
     }).catch((err) => {
         console.log('Error: Could not get status: ' + err);
     }));
+    document.getElementById('reload-status-button').innerText = window.innerWidth;
 });
```

varjolintu:
As said before, that still doesn't fix the content issues with e.g. `popup_login.html`:
Screenshot 2022-01-26 at 9 51 49

dym-sh:
well this needs complete re-flow of the layout if you want to still fit it in the 300px popup, but you didnt want 2 rows of top buttons either?

varjolintu:
Yup.. If this is going to be fixed, let's fix it properly.

varjolintu:
Replaced by: https://github.com/keepassxreboot/keepassxc-browser/pull/1571.

PR #1545, opened 2022-01-24
Fix disabled OTP icons - they are never added

Ashus:
When the checkbox **Activate 2FA/OTP field icons.** in settings is deactivated, there are some cases, when green TOTP icon with bubble shows up. This fixes that behaviour.
There may be better places to put this check, but this works fine (although change in settings needs page refresh for TOTP bubble to show up).


droidmonkey (approved):

varjolintu (approved):

varjolintu:
Good catch!

PR #1547, opened 2022-01-25
Add support for traversing child elements

varjolintu:
Add support for detecting input fields inside child elements/nodes, including those inside Shadow DOM.
The new function does only a simple visibility check for the fields. Further checks are being made in `kpxcFields.getAllCombinations()`.

Fixes multiple sites mentioned in https://github.com/keepassxreboot/keepassxc-browser/issues/1358.

### DIFF
--- a/keepassxc-browser/content/observer-helper.js
+++ b/keepassxc-browser/content/observer-helper.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const MAX_CHILDREN = 100;
+const MAX_CHILDREN = 50;
 const MAX_INPUTS = 100;
 const MAX_MUTATIONS = 200;
 
@@ -285,7 +285,14 @@ kpxcObserverHelper.ignoredNode = function(target) {
 };
 
 // Traverses all children, including Shadow DOM elements
-const traverseChildren = function(target, children) {
+const traverseChildren = function(target, children, depth = 1) {
+    depth++;
+
+    // Children can be scripts etc. so ignoredElement() is needed here
+    if (depth >= MAX_CHILDREN || kpxcObserverHelper.ignoredElement(target)) {
+        return;
+    }
+
     for (const child of target.childNodes) {
         if (child.type === 'hidden' || child.disabled) {
             continue;
@@ -295,18 +302,9 @@ const traverseChildren = function(target, children) {
             children.push(child);
         }
 
-        if (children.length === 0) {
-            continue;
-        }
-
-        // Limit the maximum number children
-        if (children.length > MAX_CHILDREN) {
-            children = children.slice(0, MAX_CHILDREN);
-        }
-
-        traverseChildren(child, children);
+        traverseChildren(child, children, depth);
         if (child.shadowRoot) {
-            traverseChildren(child.shadowRoot, children);
+            traverseChildren(child.shadowRoot, children, depth);
         }
     }
 };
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/observer-helper.js
```diff
@@ -270,3 +282,17 @@ kpxcObserverHelper.ignoredNode = function(target) {
 
     return false;
 };
+
+// Traverses all children, including Shadow DOM elements
+const traverseChildren = function(target, children) {
+    for (const child of target.childNodes) {
+        if (child.nodeName === 'INPUT' && child.type !== 'hidden' && !child.disabled ) {
```
droidmonkey, line 289: If the child is hidden or disabled, we should stop traversing down (ie, return).

keepassxc-browser/content/observer-helper.js
```diff
@@ -270,3 +282,17 @@ kpxcObserverHelper.ignoredNode = function(target) {
...
+            children.push(child);
+        }
+
+        traverseChildren(child, children);
+        if (child.shadowRoot) {
+            traverseChildren(child.shadowRoot, children);
+        }
```
droidmonkey, line 296: We should set a traversal limit, perhaps 100 levels?

droidmonkey (changes requested):

keepassxc-browser/content/observer-helper.js
varjolintu: Fixed.

keepassxc-browser/content/observer-helper.js
varjolintu: Fixed. Also added one `continue` after adding inputs to the list.

keepassxc-browser/content/observer-helper.js
```diff
@@ -162,9 +169,9 @@ kpxcObserverHelper.getInputs = function(target, ignoreVisibility = false) {
         return [];
     }
 
-    // Do not allow more visible inputs than _maximumInputs (default value: 100) -> return the first 100
+    // Do not allow more visible inputs than MAX_INPUTS (default value: 100) -> return the first 100
     if (inputFields.length > MAX_INPUTS) {
-        return inputFields.slice(0, MAX_INPUTS);
+        inputFields = inputFields.slice(0, MAX_INPUTS);
```
varjolintu, line 174: This was previously returning all inputs without the checks happening at the end of the function.

keepassxc-browser/content/observer-helper.js
droidmonkey: No that doesn't solve it, that makes this function do weird things. I meant a recursion limit (depth). I feel like we are reimplementing a basic jQuery function here. Finding specific node types isn't a built in feature of the DOM?

varjolintu: Using that function is not allowed for a `NodeList`, so I'll need to make some more modifications. Also the function calls itself over and over, so a different kind of counter needs to be implemented.

droidmonkey: You'd have to add a parameter to the function that tracks depth, defaults to 0. You then check if the max depth has been reached and stop the recursion if so.

varjolintu: Count added. Needed to revert some changes because we actually need to traverse non-input elements also. But scripts etc. must be ignored, and this is why the extra `ignoredElement()` check is added.

droidmonkey (approved):

PR #1561, opened 2022-02-20
Handle requestID with error responses

varjolintu:
Solves error responses going out-of-sync with https://github.com/keepassxreboot/keepassxc/pull/7404.

### DIFF
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -5,7 +5,6 @@ keepassClient.keySize = 24;
 keepassClient.messageTimeout = 500; // Milliseconds
 keepassClient.nativeHostName = 'org.keepassxc.keepassxc_browser';
 keepassClient.nativePort = null;
-keepassClient.requestIdSize = 8;
 
 const kpErrors = {
     UNKNOWN_ERROR: 0,
@@ -184,7 +183,7 @@ keepassClient.getNonce = function() {
     return nacl.util.encodeBase64(nacl.randomBytes(keepassClient.keySize));
 };
 
-// Creates a random 8-byte data for Request ID
+// Creates a random 8 character string for Request ID
 keepassClient.getRequestId = function() {
     return Math.random().toString(16).substring(2, 10);
 };
@@ -339,8 +338,6 @@ function onDisconnected() {
 }
 
 keepassClient.onNativeMessage = function(response) {
-    console.log(JSON.stringify(response));
-
     // Handle database lock/unlock status
     if (response.action === kpActions.DATABASE_LOCKED || response.action === kpActions.DATABASE_UNLOCKED) {
         keepass.updateDatabase();
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -5,6 +5,7 @@ keepassClient.keySize = 24;
 keepassClient.messageTimeout = 500; // Milliseconds
 keepassClient.nativeHostName = 'org.keepassxc.keepassxc_browser';
 keepassClient.nativePort = null;
+keepassClient.requestIdSize = 8;
 
 const kpErrors = {
     UNKNOWN_ERROR: 0,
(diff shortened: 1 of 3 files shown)


keepassxc-browser/background/client.js
```diff
@@ -82,8 +83,12 @@ keepassClient.sendNativeMessage = function(request, enableTimeout = false, timeo
         const listener = ((port, action) => {
             const handler = (msg) => {
                 if (msg && msg.action === action) {
+                    // If the request has a separate requestID, check if it matches when there's no nonce (an error message)
+                    const isNotificationOrError = !msg.nonce
+                        && (!request.requestID && !msg.requestID || (request.requestID && msg.requestID && request.requestID === msg.requestID))
```
droidmonkey, line 88: This is mind bending, recommend splitting this apart. I actually think it is enough to just check for no nonce and equality between the requestID's. If they are both null then it still returns true (the first set of && conditions) and if not null they return true when equal (the second set of conditions).

keepassxc-browser/background/client.js
```diff
@@ -176,6 +185,11 @@ keepassClient.getNonce = function() {
     return nacl.util.encodeBase64(nacl.randomBytes(keepassClient.keySize));
 };
 
+// Creates a short random data for Request ID
+keepassClient.getRequestId = function() {
+    return nacl.util.encodeBase64(nacl.randomBytes(keepassClient.requestIdSize));
+};
```
droidmonkey, line 191: Can we just generate a random number instead of random data? Between 0 and INT_MAX, saves the base64 encoding. Alternative would be a UUID.

droidmonkey (changes requested):

keepassxc-browser/background/client.js
varjolintu: Sure.

keepassxc-browser/background/client.js
varjolintu: Fixed the both brain fogs.

keepassxc-browser/background/client.js
```diff
@@ -176,6 +184,11 @@ keepassClient.getNonce = function() {
     return nacl.util.encodeBase64(nacl.randomBytes(keepassClient.keySize));
 };
 
+// Creates a random number for Request ID
+keepassClient.getRequestId = function() {
+    return String(Math.floor(Math.random() * Number.MAX_SAFE_INTEGER));
```
droidmonkey, line 189: This could generate a large string. Sorry to go back and forth, maybe this method is better: `Math.random().toString(16).substr(2, 8); // 6de5ccda`

varjolintu: I agree. Let's do that.

varjolintu: Fixed. Though the correct format for the generation is `Math.random().toString(16).substring(2, 10);` because `substr()` is deprecated.

keepassxc-browser/background/client.js
```diff
@@ -5,6 +5,7 @@ keepassClient.keySize = 24;
 keepassClient.messageTimeout = 500; // Milliseconds
 keepassClient.nativeHostName = 'org.keepassxc.keepassxc_browser';
 keepassClient.nativePort = null;
+keepassClient.requestIdSize = 8;
```
droidmonkey, line 8: Is this used?

keepassxc-browser/background/client.js
```diff
@@ -176,6 +184,11 @@ keepassClient.getNonce = function() {
     return nacl.util.encodeBase64(nacl.randomBytes(keepassClient.keySize));
 };
 
+// Creates a random 8-byte data for Request ID
```
droidmonkey, line 187: Random 8 character string

keepassxc-browser/background/client.js
```diff
@@ -326,6 +339,8 @@ function onDisconnected() {
 }
 
 keepassClient.onNativeMessage = function(response) {
+    console.log(JSON.stringify(response));
```
droidmonkey, line 342: Did you add this for debugging or perpetual?

keepassxc-protocol.md
```diff
@@ -20,6 +20,7 @@ Encrypted messages are built with these JSON parameters:
 - message - Encrypted message, base64 encoded
 - nonce - 24 bytes long random data, base64 encoded. This is incremented to the response.
 - clientID - 24 bytes long random data, base64 encoded. This is used for a single session to identify different browsers if multiple are used with proxy application.
+- requestID (optional) - A random 8 byte alphanumeric data. Used to identify error responses. Currently used only with `generate-password`.
```
droidmonkey, line 23: Random 8 character string

droidmonkey (approved):

keepassxc-browser/background/client.js
varjolintu: This was a mistake. I was just using it for debugging some other stuff. Sigh.

Issue #1161, opened 2021-01-02
Limit Frequency of Auto-submit Login Forms


bkonia:
The Auto-submit Login Forms option is convenient. However, if you accidentally store invalid credentials, or if saved credentials expire, it can result in an infinite loop, whereby it attempts to log in, the authentication fails and the website redirects back to the login page, where the extension tries to log in again and so on... Since each attempt occurs so quickly, this could be interpreted by the website as a brute force hack attempt, resulting in your IP address being banned.

Therefore, I'd like to suggest an option that would temporarily disable the Auto-submit feature for a configurable number of seconds after each login attempt. For example, you could configure it for five seconds. So, if it tries to log in and is redirected back to the login page within five seconds, it would not perform the auto-submit a second time. It could still fill the credentials, but the user would have to click the login button manually.  The extension could simply store the time of the most recent log in attempt for each entry and if it receives another login page for the same entry within the specified number of seconds, it would not perform the auto-submit.

varjolintu:
Note to myself: This would need a implementation to the background script because content scripts will change on every login attempt.

PR #1562, opened 2022-02-25
WIP: Setting auto-submit per page

jo-so:
I would like to hear some comments on the idea to have an "auto-submit" per page. This would solve the problem that kpxc should not auto-submit everywhere, especially after an auto-submit.


varjolintu:
So this only affects when filling credentials from the extension popup? Using the icon would still use Auto-Submit normally?

I'd prefer solving the problem using https://github.com/keepassxreboot/keepassxc-browser/issues/1161 which I've already started. It could be a global fix for preventing multiple Auto-Submits.

varjolintu:
I'm going to close the PR for inactivity soon, if there's no further progress.

Issue #721, opened 2019-12-19
Popup is too small if extension is pinned to overflow menu in Firefox


FWDekker:
The popup is too small to be useful after pinning it to the overflow menu in Firefox.

## Expected Behavior
In Firefox I can [pin extensions to the overflow menu](https://support.mozilla.org/en-US/kb/customize-firefox-controls-buttons-and-toolbars). That way, they don't take up space when I don't need them to. When I click the KeePassXC icon while it's in the overflow menu I expect it to behave exactly like it would when it isn't pinned to the overflow menu. Observe the normal behaviour in the following screenshot.

![The size of the popup when it is not pinned to the overflow menu.](https://user-images.githubusercontent.com/13442533/71154686-62408880-223c-11ea-9923-50e9a615f9f7.png)

## Current Behavior
After pinning KeePassXC-Browser to the overflow menu, the popup is much narrower than before, as shown in the screenshot below. As a result, the "Connect" button is now invisible and cannot be pressed. Horizontal scrolling (that is, scrolling while holding <kbd>Shfit</kbd>) or dragging my text selection to the right both don't seem to work.

![The size of the popup when it is pinned to the overflow menu.](https://user-images.githubusercontent.com/13442533/71154706-6cfb1d80-223c-11ea-9d3e-578bebcea190.png)

## Possible Solution
The best solution would be to make the popup the same size before and after pinning it to the overflow menu. However, an alternative would be to prevent the overflow of elements to make sure elements are never hidden, but this has the drawback that the popup may become too tall. Maybe both should be done.

## Steps to Reproduce (for bugs)
1. Install the KeePassXC-Browser plugin.
2. Pin the plugin icon to the overflow menu.
3. Open the plugin popup from the overflow menu.
4. Observe its size.

## Debug info
KeePassXC - 2.5.1
KeePassXC-Browser - 1.5.4
Operating system: Windows 10
Browser: Firefox 71.0

varjolintu:
Seems like this is a Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1373490

furious:
Removing `min-width` should help...

https://github.com/keepassxreboot/keepassxc-browser/blob/b08966c6fd709e334962e661e43b40ce1345ddfe/keepassxc-browser/popups/popup.css#L8
[...]

PR #1571, opened 2022-03-11
Adjust the popup size when needed. Change upper row buttons to icons.

varjolintu:
- Changes the upper row buttons to icons for better scaling with Firefox's pinned popup. The old button texts are visible as tooltips.
- Adjusts the popup width when needed. With Chromium based browser popups have no other width than the content if it's not specified.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/721.


droidmonkey:
The hand icon is rather ambiguous. Any other choices or maybe a short text like "Custom Fields"

varjolintu:
@droidmonkey I'd prefer an icon. Any suggestions? The old icon without a text was not very informative.

droidmonkey:
How about something like this with the hand on top instead of the cursor:

https://materialdesignicons.com/icon/form-textbox-password

varjolintu:
Something like that could work.

varjolintu:
Maybe something like this? I took the TOTP icon as the base.
Screenshot 2022-03-12 at 11 34 27

droidmonkey:
That's pretty nice! Kind of mickey mouse looking thigh

varjolintu:
The transparent style:
Screenshot 2022-03-12 at 14 57 26

droidmonkey:
Personally think that's perfect

varjolintu:
Let's keep that then.

varjolintu:
Oh, one more option, if we want to keep the icon "filled". But maybe that's too much Mickey Mouse too.

Screenshot 2022-03-12 at 17 14 24

droidmonkey:
Oh the final version is hawt

varjolintu:
Or we could go dark.. difficult decisions:
Screenshot 2022-03-12 at 19 30 03

varjolintu:
Uh oh. This is totally broken after the removal of jQuery. Needs a fix.

droidmonkey:
This works well, no problems from me

droidmonkey (approved):

PR #1573, opened 2022-03-17
Fix filling TOTP to fields not in combinations

varjolintu:
Found a bug where filling TOTP might not work in cases where the field is not regognized as a part of any current input field combination. In these cases the current input field is used.
This fix should allow filling TOTP to the field user has selected. Also, `await` was missing from the fill call causing `this.input` to be undefined in some cases.

Fixes.. probably a few TOTP bugs.


droidmonkey (approved):

droidmonkey:
Oops, I bet the missing await is the cause of a ton of inconsistencies

moeffju:
Can confirm this fixed the TOTP bugs I reported previously. Can't wait for 1.8.0. :D

PR #1580, opened 2022-03-20
Allow filling password for non-detected fields and combinations

### LINKED ISSUE #1574: Fill password from context menu or keyboard should work for undetected fields.
## Expected Behavior

When filling password, input fields need to be a part of a "combination" before it can be properly filled. With TOTP it's allowed to fill manually a field that is not part of any combination. The password fill should be handled identically.

## Current Behavior

Password is not filled if the input field is not part of a combination.

## Debug info

KeePassXC - 2.7.0-beta
KeePassXC-Browser - 1.7.11 / 1.8.0-dev

varjolintu:
Allows filling password manually from context menu or keyboard shortcut to an input field even if it's not part of any previously detected combination.

Looking at the source, there has to be some logical error in the implementation when filling using a combination. It returns the whole function if a combination is not found and never goes to line 62. Now the function for filling combinations is a separate one, which allows the function to continue even if combination detection fails.

Fixes #1574.

### DIFF
--- a/keepassxc-browser/content/fill.js
+++ b/keepassxc-browser/content/fill.js
@@ -23,6 +23,7 @@ kpxcFill.fillAttributeToActiveElementWith = async function(attr) {
 
 // Fill requested from the context menu. Active element is used for combination detection
 kpxcFill.fillInFromActiveElement = async function(passOnly = false) {
+    await kpxc.receiveCredentialsIfNecessary();
     if (kpxc.credentials.length === 0) {
         logDebug('Error: Credential list is empty.');
         return;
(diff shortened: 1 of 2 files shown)


droidmonkey:
This doesn't seem to do anything for me, I still cannot fill random input fields using either Username/Password or Password Only selections. Do you have a test site for this?

varjolintu:
Missed a couple of things. Also removed the restriction to fill password to username fields.

I used this site for testing, because it's not enabled with Username-Only option by default -> there are no credentials received before the fill.
https://account.kyivstar.ua/cas/login

droidmonkey:
I tested this on twitter.com and it only worked _after_ I manually redetected fields. Then the fill icon showed up and I could fill the field with right click -> Fill username and password / Fill Password. So this works, but I think we need to "Redetect Fields" better on some sites. Also, on Twitter it keeps trying to pull my google accounts even though it is clearly a twitter site.

droidmonkey (approved):

varjolintu:
Strange. Twitter.com shouldn't even require redetecting fields. In this case the redetect shouldn't be required at all. Can you find another site where this can be seen more clearly?

I'll make another PR for the Google Account pulling. I noticed that we are making cross-origin requests to KeePassXC (in this case it's the Google login button that triggers it) even if content scripts are not allowed in those frames.

PR #1602, opened 2022-04-12
Check that maxLength is above zero

varjolintu:
Additional fix to #1596. Input field's `maxLength` can be `-1` in some cases, which passes the current check and gives an incorrect notification.


droidmonkey (approved):

droidmonkey:
Convinced that web developers are terrible people

PR #1603, opened 2022-04-15
WIP: perf: minimize `_locales` files

WofWca:
TODO: state `minimize-webext-i18n-json` as a dependency in package.json

This currently reduces the size of dist ZIPs by 14%

### DIFF
--- a/build.js
+++ b/build.js
@@ -5,6 +5,7 @@ const extra = require('fs-extra');
 const util = require('util');
 const exec = util.promisify(require('child_process').exec);
 const zaf = require('zip-a-folder');
+const { minimizeDirectory: minimizeLocalesDirectory } = require('minimize-webext-i18n-json');
 
 const DEST = 'keepassxc-browser';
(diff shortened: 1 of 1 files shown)


droidmonkey (approved):
as long as the translations still work, why not

varjolintu:
Not sure if this is really needed. Just adds another dependency. The extension size is not a problem anyway.

WofWca:
Well, that dependency is not that big either, so I'd say it's worth it. But I respect your decision.

I've still updated the branch so it's ready to be merged.

varjolintu:
Instead of adding new dependencies, we should focus removing all possible ones instead. Especially when it comes to scripts that are ran locally.

WofWca:
Off-topic, but just curious, why _especially_ the ones that are ran locally? Wouldn't client-side dependencies be more impactful?

varjolintu:
Because those local scripts would be running on my (virtual) computer, modifying files.

PR #1604, opened 2022-04-16
Show a link to the Getting Started Guide for new users

varjolintu:
Shows a link to the Getting Started Guide when there's no connection to KeePassXC. The alert can be dismissed, which also hides it permanently so it doesn't annoy anyone who is already familiar how the extension works.

Screenshot 2022-04-16 at 10 29 24


droidmonkey (approved):
Love it! Can we do same for extension troubleshooting guide if they hit refresh more than 3 times and key exchange fails?

varjolintu:
@droidmonkey Sounds like a great idea! I'll make another PR for that.

cjpoor:
Thank you for the great work!

PR #1609, opened 2022-04-18
Hide Automatic Reconnect option

varjolintu:
This option should be hidden for now. It causes more problems than actually solves.
Those who are unaware: the reconnect should happen in the proxy instead. However, mixing threads/signals and sockets causes the proxy to crash, and for now there's no good solution for it. The problem needs to be investigated further.


droidmonkey (changes requested):
We also need to disable/ ignore it if it was enabled by the user

droidmonkey (approved):

PR #1614, opened 2022-04-24
Allow filling Credit Card information

varjolintu:
- Adds basic detection for credit card forms
- Adds a new icon for filling credit card information from the name field

Currently the following KPH attributes are supported in this WIP branch:
```
KPH: CC_CCV
KPH: CC_EXP
KPH: CC_EXP_MONTH
KPH: CC_EXP_YEAR
KPH: CC_NAME
KPH: CC_NUMBER
KPH: CC_TYPE
```

TODO:
- `CC_EXP` and `CC_TYPE` are not yet used/filled.
- Attributes used above should be properly defined. What attributes are the other KeePass clients using for their templates?
- Credit Card information should be stored to one entry. When trying to fill the information using the icon, KeePassXC could ask "Do you want to allow this site to fill Credit Card information?" and then add the site URL to entry's Additional URL's list.

### DIFF
--- a/keepassxc-browser/content/credit-card-form.js
+++ b/keepassxc-browser/content/credit-card-form.js
@@ -1,0 +1,156 @@
+'use strict';
+
+const cardNameRegex = /(cc|card|cardholder).*-name|name/i;
+const cardNumberRegex = /(cc|card).*-(num|number|no)|number|card-no/i;
+const cardTypeRegex = /(cc|card|cb).*-type|brand/i;
+const cardCCVRegex = /(cc|card|security|verification).*-(code|cvv|cvc|csc)|cvv|cvc|csc/i;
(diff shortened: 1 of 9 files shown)


droidmonkey:
I dunno how I feel about extending the `KPH: ` syntax. I don't have a better alternative right now though.

varjolintu:
> I dunno how I feel about extending the `KPH: ` syntax. I don't have a better alternative right now though.

Yeah. I don't prefer it either, but at least this draft can be tested with them because it doesn't need any code changes to KeePassXC side.

droidmonkey:
That prefix is also challenged because it prefers a specific solution (KeePass2Android). The custom entry layouts will likely not prefer any prefix. The solution should be defined such that no prefix is required at all, you would map chosen attributes to their function.

varjolintu:
I'd suggest that when the template implementation is ready on KeePassXC side, `get-logins` should return an extra array `templates` with the usual `entries`. That would keep them separate from normal credentials.

PR #1621, opened 2022-05-08
Fix setting icon to all tabs

### LINKED ISSUE #1106: Icon momentarily goes green when refreshing or navigation to a page, with locked database.
## Expected Behavior
The key icon in the browser should show the lock icon when the database is locked. And only the green icon when the database is open.

## Current Behavior
When navigating to a page with the browser or refreshing the current one, the key icon will momentarily turn green indicating an unlocked database when in fact it is locked. 

## Possible Solution
Haven't really had time to dig into the code. 

1. Have KeypassXC database locked and this extension installed.
2. Navigate to a page in browser or refresh
3. Observe the icon flashing green, and then to locked.

## Debug info
KeePassXC - 2.6.2.
KeePassXC-Browser - 1.7.3
OS: Linux
Browser: Firefox 81.0.2 (64-bit)

varjolintu:
Set icon to all tabs at once, preventing a flash from a previous/default state.

Fixes #1106.

### DIFF
--- a/keepassxc-browser/background/browserAction.js
+++ b/keepassxc-browser/background/browserAction.js
@@ -10,7 +10,6 @@ browserAction.show = function(tab, popupData) {
     page.popupData = popupData;
 
     browser.browserAction.setIcon({
-        tabId: tab.id,
         path: browserAction.generateIconName(popupData.iconType)
     });
 


droidmonkey:
How often is this function called? If it's a lot, I could see this being a problem for tab hoarders.

varjolintu:
Maybe we could add a check that the function is used only when the icon is actually changed from a previous state.

droidmonkey (approved):

WhyNotHugo:
I came cross another bug on latest stable:

- Close KP
- Open a browser window
- Open KP
- Open a new browser window

The first browser window has a red-cross icon, while the second one has a padlock icon.

I _think_ this change also fixes that, since there's no longer per-tab icons.

PR #1637, opened 2022-06-04
Prevent endless loops with Auto-Submit

### LINKED ISSUE #1161: Limit Frequency of Auto-submit Login Forms

varjolintu:
If Auto-Fill Single Credentials and Auto-Submit both are enabled, prevent Auto-Submit for five seconds before it's accepted again. This prevents endless loops at the login process when user is entering a wrong password, or Auto-Submit does not regognize the submit button correctly and returns to the login page.

Fixes #1161.


droidmonkey (approved):

Staubgeborener:
great, can't wait for the new release

PR #1662, opened 2022-07-01
Fixes a bug in Firefox where the Firefox version wasn't being correctly parsed

jrc03c:
In Firefox 102.0, the version number was being incorrectly parsed as 10 rather than 102 because only 2 digits were being selected after the final "/"; so the button that enables navigation to the keyboard shortcuts page was being removed. This PR fixes that small bug.


varjolintu:
Actually, this function is no longer used so a proper fix would be to remove it totally :)

jrc03c:
Gotcha! Should I do that in this PR, or will it be handled elsewhere?

varjolintu:
@jrc03c No need to worry about it :)

varjolintu (approved):

droidmonkey (approved):

PR #1667, opened 2022-07-06
Prevent multiple simultaneous submits

### LINKED ISSUE #1656: Banner asking to save/update entry does not close anymore
## Overview

Since the last update of the Browser extension, every time you log in on a site typing your user name and password, the banner that asks whether to save/update the entry appears, but when you click for example on the *Dismiss* button, nothing happens, the banner does not go away.

I opened the developer tools to see whether an JS exception is raised when you click on the button, but no error message is displayed.

It seems that the problem is that the banner itself cannot be closed. Just for testing purposes, I clicked on the *New* button of the banner and the entry was added on my password file, the green notification that a new entry was stored also appeared. However the banner remains in place. I have to press F5 to get rid of the banner.

## Expected Behavior

The banner should go away after clicking any button

## Current Behavior
The banner does not go away

1. Go to  a page that requires login
2. Enter your user name & password
3. Click on any of the buttons (for example the *Dismiss* button)

## Debug info

KeePassXC - `2.7.1`  
KeePassXC-Browser - `1.8.0`  
Operating system: GNU/Linux  
Browser:
  - GNU/Linux on Google chrome Version 102.0.5005.115 (Official Build) (64-bit)
  - GNU/Linux on Firefox 91.10.0esr (64-bit)

**NOTE** the behaviour is present on both browsers that I tested.

varjolintu:
With some pages the extension detects both form submit and a submit button. When submit happens, two Credential Banner are created to DOM at the same time, and there's no reliable way to destroy them. The banner cannot be closed and stays on the page unless it's reloaded.

The bug can be reproduced every time when filling credentials to the login fields and pressing Sign On:
https://home.cards.citidirect.com/CommercialCard/login

Fixes #1656.

### DIFF
--- a/keepassxc-browser/content/form.js
+++ b/keepassxc-browser/content/form.js
@@ -7,6 +7,7 @@
 const kpxcForm = {};
 kpxcForm.formButtonQuery = 'button[type=button], button[type=submit], input[type=button], input[type=submit], button:not([type]), div[role=button]';
 kpxcForm.savedForms = [];
+kpxcForm.submitTriggered = false;
 
 // Returns true if form has been already saved
 kpxcForm.formIdentified = function(form) {
@@ -116,6 +117,13 @@ kpxcForm.onSubmit = async function(e) {
         return;
     }
 
+    // Prevent multiple simultaneous submits
+    if (!kpxcForm.submitTriggered) {
+        kpxcForm.submitTriggered = true;
+    } else {
+        return;
+    }
+
     const searchForm = f => {
         if (f.nodeName === 'FORM') {
             return f;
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/form.js
```diff
@@ -116,6 +117,13 @@ kpxcForm.onSubmit = async function(e) {
         return;
     }
 
+    // Prevent multiple simultaneous submits
+    if (!kpxcForm.submitTriggered) {
+        kpxcForm.submitTriggered = true;
+    } else {
+        return;
+    }
```
droidmonkey, line 125: This would be easier to understand as:

```
if (kpxcForm.submitTriggered) {
    return;
}

kpxcForm.submitTriggered = true;
```

droidmonkey (approved):

PR #1700, opened 2022-08-21
Allow autofill configuration on a per-credential basis

b73f9:
Hey, 
I want to add this small bit functionality to allow configuring auto-fill on a per-credential basis. 
Tested it locally and works as expected. This is using the KPH-prefixed advanced fields which is a but ugly, but it works pretty well to make the parameter easily syncable across multiple devices sharing a single db.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -446,10 +446,19 @@ kpxc.prepareCredentials = async function() {
         logDebug('Error: No combination found.');
         return;
     }
-
-    if (kpxc.settings.autoFillSingleEntry && kpxc.credentials.length === 1) {
-        kpxcFill.fillFromAutofill();
-        return;
+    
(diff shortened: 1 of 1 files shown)


varjolintu:
Please make the code style identical to the other functions.

b73f9:
@varjolintu Thanks for the comment! 
Tried to make it more consistent, does it look good now? If not, could you say which specific bits are problematic?

varjolintu:
The only thing I'm thinking about here is that all the other special options (use this entry only for HTTP Auth etc.) are defined in KeePassXC, but this is the first and only option that is defined by a value inside a String Field attribute. Maybe it could be more consistent way to make a new option for this also.

b73f9:
Ah, fair, I didn't realize these existed (they don't seem to show up on the kpxc.credential items available in this context). 
Yeah that'd definitely be nicer, will look into doing it that way.

PR #1704, opened 2022-08-24
Fix PayPal TOTP

### LINKED ISSUE #1703: PayPal new TOTP
Feature request :-)

PayPal has launched a new TOTP interface. Instead of one input field, there are now six individual fields for one digit each. Unfortunately, the automatic filling out doesn't work anymore. Maybe there is a way to have this field filled in automatically again in the future.

Thank you for the effort!

Screenshot_001

## My versions
KeePassXC - 2.7.1
KeePassXC-Browser - 1.8.1
Operating system: Linux
Browser: Vivaldi

varjolintu:
PayPal recently changed their 2FA page to a segmented fields. The form has six inputs plus a button as the last element. This is not currently supported.

Fix also bumps the version number.

Fixes #1703.

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -115,11 +115,31 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
         }
     };
 
+    const formLengthMatches = function(currentForm) {
+        if (!currentForm) {
+            return false;
+        }
+
+        // Accept 6 inputs directly
+        if (currentForm.length === 6) {
+            return true;
+        }
+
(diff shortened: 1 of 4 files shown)


phoerious (approved):

droidmonkey:
This is still a brittle check, but I can't think of a better way to reverse engineer every websites dumb TOTP setup without reams of code.

varjolintu:
> This is still a brittle check, but I can't think of a better way to reverse engineer every websites dumb TOTP setup without reams of code.

Gotta ignore any false-positives somehow..

PR #1708, opened 2022-09-03
Fill after reconnect

### LINKED ISSUE #1680: Reconnect & fill when using the username input icon
## Expected Behavior
If you use the username field icon without being connected, a connection should be established and the input fields should be filled in with just one click.

## Current Behavior
Two clicks are required to fill input fields. The first click reconnects to the database and a second click is required to fill the fields.

Also, thank you @varjolintu for #1676. I no longer need to press the reload button in 1.8.1.

## Debug info

KeePassXC - 2.7.1
KeePassXC-Browser - 1.8.1
Operating system: Win64
Browser: Chrome/Chromium 103.0.0.0

varjolintu:
After reconnecting to KeePassXC when clicking the Username Icon, fill is not made automatically.

Fixes #1680.

### DIFF
--- a/keepassxc-browser/content/username-field.js
+++ b/keepassxc-browser/content/username-field.js
@@ -123,6 +123,8 @@ const iconClicked = async function(field, icon) {
         return;
     }
 
+    let reconnected = false;
+
     // Try to reconnect if KeePassXC is not currently connected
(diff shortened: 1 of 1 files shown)


droidmonkey (approved):
Never had this problem but fix is good

PR #1711, opened 2022-09-06
Fix Steam TOTP

### LINKED ISSUE #1710: Steam new TOTP
Similar to #1703 steam has a new interface for handling TOTPs. Keepassxc-browser currently does not identify the field as totp and one has to copy/paste to fill in. It would be nice if the extensions handle this field correctly. Thanks!

image

varjolintu:
Steam also uses a segmented 2FA, but with five alphanumerics instead of six numbers. Some hard-coded requirements for the six char length is changed to a more dynamic implementation when doing the fill.

Needs to be tested properly. Steam locked my account temporarily when making the fix..

Fixes #1710.

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const DEFAULT_SEGMENTED_TOTP_FIELDS = 6;
+
 /**
  * @Object kpxcFields
  * Provides methods for input field handling.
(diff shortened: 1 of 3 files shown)


keepassxc-browser/content/fields.js
```diff
@@ -92,9 +92,12 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
...
         return;
     }
-    const addTotpFieldsToCombination = function(inputFields) {
+
+    let exceptionFound = false;
+
+    const addTotpFieldsToCombination = function(inputFields, ignoreLength = false) {
         const totpInputs = Array.from(inputFields).filter(e => e.nodeName === 'INPUT' && e.type !== 'password' && e.type !== 'hidden');
-        if (totpInputs.length === 6) {
+        if (totpInputs.length === 6 || ignoreLength) {
```
droidmonkey, line 100: How about instead we add a "expectedDigits" variable and default its value to 6.

droidmonkey (approved):

keepassxc-browser/content/fields.js
varjolintu: Good idea.

droidmonkey: Ok not too concerned, you can merge

PR #1726, opened 2022-09-17
Allow filling TOTP without a combination

### LINKED ISSUE #1649: Can't fill TOTP on app.clickup.com
## Expected Behavior
Autofill TOTP fields or manually after using the keyboard shortcut.

## Current Behavior
Upon autofill/when clicking the autofill button the extension shows the "No TOTP found" dialog:
image

When using the keyboard shortcut nothing happens except a logged error in the console:
image
image

## Debug info

KeePassXC - 2.7.1
KeePassXC-Browser - 1.7.12
Operating system: Windows 10
Browser: Firefox 102

varjolintu:
At some pages there's no combination detected for the TOTP. It should still be able to fill using the context menu.

Fixes #1649.

### DIFF
--- a/keepassxc-browser/content/fill.js
+++ b/keepassxc-browser/content/fill.js
@@ -169,7 +169,7 @@ kpxcFill.fillTOTPFromUuid = async function(el, uuid) {
 
 // Set normal or segmented TOTP value
 kpxcFill.setTOTPValue = function(elem, val) {
-    if (kpxc.combinations.length === 0) {
+    if (kpxc.credentials.length === 0) {
         logDebug('Error: Credential list is empty.');
         return;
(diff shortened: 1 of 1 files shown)


droidmonkey (approved):
So this was a typo? Oopsie

varjolintu:
@droidmonkey Combination is a predefined "struct" of all detected fields. In this case there's only TOTP field and it's not detected so there's no combination available.

Previously a fix for this was done here: https://github.com/keepassxreboot/keepassxc-browser/pull/1573 but it didn't cover a case like this. However, I still need to check a few things for this.

droidmonkey:
Gotcha make sense

Issue #1727, opened 2022-09-18
Reconnect and fill from keyboard shortcuts and context menu


varjolintu:
## Expected Behavior

The extension should make an automatic reconnect and fill when using the context menu or keyboard a shortcuts for filling. The functionality would be similar to https://github.com/keepassxreboot/keepassxc-browser/pull/1676 and https://github.com/keepassxreboot/keepassxc-browser/pull/1708.

## Current Behavior

Filling from context menu or keyboard shortcut doesn't reconnect. Only clicking the Username Icon works.

## Possible solutions
Make the reconnect function more generic that can be called from multiple objects/classes.

## Debug info

KeePassXC - 2.7.1
KeePassXC-Browser - 1.8.3-dev
Operating system: Mac/Win/Linux
Browser: Chrome/Firefox/Vivaldi/Chromium

PR #1728, opened 2022-09-18
Automatically reconnect on keyboard shortcuts

3nt3:
Relevant issue #1727 

This PR adds calls to a `reconnect()` function when the following actions are fired

- `fill_user_pass_with_specific_login`
- `fill_username_password`
- `fill_password`
- `fill_totp`

The logic for `reconnect()` is identical to that which is used for the onClick listener on the username icon.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -833,28 +833,20 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
         } else if (req.action === 'clear_credentials') {
             kpxc.clearAllFromPage();
         } else if (req.action === 'fill_user_pass_with_specific_login') {
-            try {
-                await kpxc.reconnect();
-            } catch (_) {}
+            await kpxc.reconnect();
             kpxcFill.fillFromPopup(req.id, req.uuid);
         } else if (req.action === 'fill_username_password') {
-            try {
-                await kpxc.reconnect();
-            } catch (_) {}
+            await kpxc.reconnect();
             sendMessage('page_set_manual_fill', ManualFill.BOTH);
             await kpxc.receiveCredentialsIfNecessary();
             kpxcFill.fillInFromActiveElement();
         } else if (req.action === 'fill_password') {
-            try {
-                await kpxc.reconnect();
-            } catch (_) {}
+            await kpxc.reconnect();
             sendMessage('page_set_manual_fill', ManualFill.PASSWORD);
             await kpxc.receiveCredentialsIfNecessary();
             kpxcFill.fillInFromActiveElement(true); // passOnly to true
         } else if (req.action === 'fill_totp') {
-            try {
-                await kpxc.reconnect();
-            } catch (_) {}
+            await kpxc.reconnect();
             await kpxc.receiveCredentialsIfNecessary();
             kpxcFill.fillFromTOTP();
         } else if (req.action === 'fill_attribute' && req.args) {
@@ -883,7 +875,7 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
 });
 
 // automatically reconnect to KeePassXC
-// returns true if reconnected
+// returns true if connected afterwards
 kpxc.reconnect = async function() {
     // Try to reconnect if KeePassXC is not currently connected
     const connected = await sendMessage('is_connected');
@@ -891,9 +883,7 @@ kpxc.reconnect = async function() {
         const reconnectResponse = await sendMessage('reconnect');
         if (!reconnectResponse.keePassXCAvailable) {
             kpxcUI.createNotification('error', tr('errorNotConnected'));
-            throw "not connected";
         }
-        return true;
     }
-    return false;
-}
+    return await sendMessage('is_connected');
+};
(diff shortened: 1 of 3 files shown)


varjolintu:
Please revert the code cleaning from the PR and only commit relevant changes.

3nt3:
I moved the `reconnect()` function to `global.js` which I hope is a good place for it and used it in the username field logic as well. Hope this PR is okay now.

Btw it would be nice to add a `.prettierrc` or something comparable to the project since editorconfig seems to not have that many options (trailing comma, quote style etc.)

keepassxc-browser/content/username-field.js
```diff
@@ -123,19 +123,8 @@ const iconClicked = async function(field, icon) {
...
-    if (!connected) {
-        const reconnectResponse = await sendMessage('reconnect');
-        if (!reconnectResponse.keePassXCAvailable) {
-            kpxcUI.createNotification('error', tr('errorNotConnected'));
-            return;
-        }
-
-        reconnected = true;
-    }
+    let reconnected = await reconnect();
```
varjolintu, line 127: This is not working as intended. Please see the original function.

varjolintu:
> I moved the `reconnect()` function to `global.js` which I hope is a good place for it and used it in the username field logic as well. Hope this PR is okay now.
> 
> Btw it would be nice to add a `.prettierrc` or something comparable to the project since editorconfig seems to not have that many options (trailing comma, quote style etc.)

It's enough if the function is moved under `kpxc`. In `global.js` it can be accessed from background and content scripts + from HTML pages, so it's too visible there.

For the code cleaning, there are no "pretty" rules for anything right now, but still, those should be done in a different PR and not be mixed by this.

keepassxc-browser/content/username-field.js
3nt3: oof, I'll look into it again 🙃

keepassxc-browser/content/username-field.js
```diff
@@ -119,22 +119,27 @@ UsernameFieldIcon.prototype.createIcon = function(field) {
...
+    try {
+        reconnected = await kpxc.reconnect();
+    } catch (_) {
+        return;
+    }
 
-    const databaseHash = await sendMessage('check_database_hash');
-    if (databaseHash === '') {
+    const databaseHash = await sendMessage("check_database_hash");
+    if (databaseHash === "") {
```
varjolintu, line 135: Please don't change `'` to `"`.

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -833,16 +833,28 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
         } else if (req.action === 'clear_credentials') {
             kpxc.clearAllFromPage();
         } else if (req.action === 'fill_user_pass_with_specific_login') {
+            try {
```
varjolintu, line 836: Would it be easier to do one try/catch inside the actual function? If it's even needed.

3nt3: inside `reconnect`? that would defeat the purpose of being able to "return" an error and also a boolean

varjolintu: I mean.. why not returning false instead of throw?

droidmonkey: There is no sense for try catch,  remove all that. The only error you provide is 'not connected' which is obviously deduced with a boolean.

3nt3: > I mean.. why not returning false instead of throw?

because there is a difference between not connecting and there being an error connecting. [The username field checks for that and returns prematurely](https://github.com/keepassxreboot/keepassxc-browser/pull/1728/files#diff-a8d53233fe21da0571fddbcbecb85e2055b697c4e0c96328670f06c3f4deb3afR131)

3nt3: but that check also isn't really critical so I can just change `reconnect()` to return if we're connected, instead of if the connection state changed

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -869,3 +873,17 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
...
+
+// automatically reconnect to KeePassXC
+// returns true if connected afterwards
+kpxc.reconnect = async function() {
+    // Try to reconnect if KeePassXC is not currently connected
+    const connected = await sendMessage('is_connected');
+    if (!connected) {
+        const reconnectResponse = await sendMessage('reconnect');
+        if (!reconnectResponse.keePassXCAvailable) {
+            kpxcUI.createNotification('error', tr('errorNotConnected'));
```
droidmonkey, line 885: `return false;` after this line

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -869,3 +873,17 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
...
+        }
+    }
+    return await sendMessage('is_connected');
```
droidmonkey, line 888: `return true;`

droidmonkey (changes requested):

droidmonkey (approved):

varjolintu (approved):

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -869,3 +873,18 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
         }
     }
 });
+
+// automatically reconnect to KeePassXC
```
varjolintu, line 877: Just a nitpick, but capital case letter is expected here :)

3nt3:
Can this be merged?

varjolintu:
> Can this be merged?

Yes. Doing it now :) Thanks for the fix!

PR #1743, opened 2022-10-08
Update autocomplete.js

onthegocode:
Updating for loop to modern forEach method to loop over node list and remove class.


droidmonkey (approved):

onthegocode:
Thank you very much for the approval!

PR #1756, opened 2022-10-28
Update eslint rules, add `.prettierignore` file, fix some formatting, and fix some bugs

stefansundin:
Heya,

I originally was just going to add a `.prettierignore` that ignored everything (somehow my VSCode insists on formatting the code even though I'm using `"prettier.requireConfig": true`), but then I decided to also see if I could fix some formatting and maybe find some bugs.

I think I managed to find some minor things. I put an inline comment on a potential mistake. I fixed a `content_security_policy` bug in `manifest.json`.

I suspect it would be hard to adopt prettier for the whole repository. There are some issues like https://github.com/prettier/prettier/issues/3805 that makes it annoying to fully adopt. But with this setup it is easy to temporarily enable and maybe adopt some styling conventions incrementally.

Let me know if you want me to change anything, and feel free to push directly to this branch if needed.

Might be useful to review with `w=1`: https://github.com/keepassxreboot/keepassxc-browser/pull/1756/files?w=1


keepassxc-browser/content/fill.js
```diff
@@ -106,7 +106,7 @@ kpxcFill.fillFromPopup = async function(id, uuid) {
 
     // For Google password field we need to do some special handling. The password field is actually in the
     // second combination that was just detected after a username fill.
-    let combination = kpxc.combinations[0];
+    let combination = kpxc.combinations[0]; // BUG? This variable is unused since https://github.com/stefansundin/keepassxc-browser/commit/ba5a7141fba456ee409446ef32da2b19c416cee3#diff-b6dc310afe6e36d10e60a76ebd2e1d07ee0eeda631e4430aeab384fbd37cdb4cL114
```
stefansundin, line 109: Want to call this out. Either there's a bug right now or this unused variable can be removed.

varjolintu: Must check it out if it's a leftover or something else.

varjolintu:
Thanks for the PR! This has been on my TODO-list for too long.

.eslintrc
```diff
@@ -1,13 +1,14 @@
 {
     "extends": ["eslint-config-airbnb-base" ],
```
stefansundin, line 2: A very large number of rules are overridden. Perhaps there's a better base config to derive from these days. Maybe `"eslint:recommended"`?

varjolintu: I agree. Years ago those rules gave a good base, but something more general would be better. `eslint:recommended` would be fine.

keepassxc-browser/content/fill.js
varjolintu: I guess this was originally for the Auto-Submit and Google, but the current version works just fine with it. So this can be removed.

varjolintu:
I tested this quickly, and I agree some of the rules are a bit too strict (for example function calls are often changed to ones with each parameter on a separate line), but this is still better than nothing. And it's good to have those HTML errors fixed too. Thanks!

keepassxc-browser/content/fill.js
stefansundin: Removed. 👍

.eslintrc
varjolintu: Do you have plans to change this, or keep the current one?

stefansundin: Oh, sorry, no I wasn't planning on making any further changes in this PR. If I ever work in this area again I might consider making an attempt, but I think this is good for now. Please, feel free to merge.

varjolintu (approved):

PR #1758, opened 2022-10-28
Fix enter key issue on Google

stefansundin:
When you use the autocomplete feature on the Google login page, using the enter key would trigger a form submit before the username is filled in and you would get the error `Enter an email or phone number`. The effect is a bit jarring.

`stopPropagation` needs to be called as well since Google probably has a key listener above the input element. I added a call to `stopImmediatePropagation` as well for good measure.

It would make sense to put this new `cancelEvent` helper function somewhere else since I think it should probably be used in other places where `preventDefault` is called.

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -2,6 +2,12 @@
 
 const MAX_AUTOCOMPLETE_NAME_LEN = 50;
 
+function cancelEvent(e) {
+    e.preventDefault();
+    e.stopPropagation();
+    e.stopImmediatePropagation();
+}
+
 class Autocomplete {
     constructor() {
         this.afterFillSort = SORT_BY_MATCHING_CREDENTIALS_SETTING;
@@ -177,7 +183,7 @@ class Autocomplete {
 
         const inputField = e.target;
         if (e.key === 'ArrowDown') {
-            e.preventDefault();
+            cancelEvent(e);
             // If the list is not visible, show it
             if (!this.input) {
                 await this.showList(inputField);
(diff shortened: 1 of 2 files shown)


keepassxc-browser/content/autocomplete.js
```diff
@@ -193,18 +199,14 @@ class Autocomplete {
...
         } else if (e.key === 'ArrowUp' && this.list) {
-            e.preventDefault();
+            cancelEvent(e);
             const items = this.getAllItems();
             this.index = (this.index > 0 ? this.index : items.length) - 1;
             this.selectItem();
         } else if (e.key === 'Enter' && this.input) {
-            if (inputField.value === '') {
-                e.preventDefault();
-            }
```
stefansundin, line 203: I don't think this is necessary anymore. Not sure what the purpose was actually. We check below if an item is selected and if so we call `preventDefault`. I don't think we have to do it when an item isn't selected.

varjolintu: If I remember correctly, this prevented the submit when Autocomplete Menu was active and enter was pressed, but the field was still empty. Maybe. Remember to comment your code people.. :)

varjolintu:
Works great.

The only think I noticed is that now the menu has radius at the bottom before the footer.
Screenshot 2022-11-02 at 22 24 09

varjolintu:
You're correct. The latest commit fixes it.

varjolintu (approved):

keepassxc-browser/css/autocomplete.css
```diff
@@ -2,6 +2,11 @@
     display: none;
     z-index: 2147483646;
     position: absolute !important;
+    background-color: #ddd;
+    border: 1px solid rgba(0,0,0,.125);
```
varjolintu, line 6: I just noticed that this shouldn't be here. Now it adds an extra border to the whole Autocomplete Menu instead of the footer. My bad for not seeing it earlier.

PR #1770, opened 2022-11-04
Fix Autocomplete Menu border color on dark theme

varjolintu:
Introduced in https://github.com/keepassxreboot/keepassxc-browser/pull/1758 the Autocomplete Menu now has an extra border on dark theme. This removes it.

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -93,7 +93,7 @@ class Autocomplete {
             this.shadowRoot.append(styleSheet);
 
             this.list = kpxcUI.createElement('div', 'kpxcAutocomplete-items', { 'id': 'kpxcAutocomplete-list' });
-            initColorTheme(this.list);
+            initColorTheme(this.container);
 
             this.container.append(this.list);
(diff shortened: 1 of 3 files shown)


droidmonkey (approved):

stefansundin:
I didn't realize that this might look bad in dark mode. The reason why I changed this is because I think this doesn't look good:

Top:

Screenshot 2022-11-04 at 16 10 05

Bottom:

Screenshot 2022-11-04 at 16 09 48

I think having a border shows the outline of the dropdown much better and makes it look a lot better:

Top:

Screenshot 2022-11-04 at 16 10 43

Bottom:

Screenshot 2022-11-04 at 16 10 52

Is it possible to make the border look good in both dark and light mode?

varjolintu:
@stefansundin One thing that could work is just to move `background-color: #ddd;` under `#kpxcAutocomplete-container footer` instead of these fixes. At least it looks ok. The border color in dark theme comes from that and not from the `border`.

varjolintu:
Ah, the `initColorTheme()` was used to the list instead of the container class. Now it should work with a new CSS variable.

PR #1771, opened 2022-11-05
Remove special handling for Firefox overflow/hamburger menu

varjolintu:
Firefox 106 has fixed the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1786587) where the popup was smaller size than the normal one, when it was moved to overflow menu in Firefox's toolbar. The special handling for popup size is now removed.

Previously fixed in: https://github.com/keepassxreboot/keepassxc-browser/pull/1571.
Related issue: #721.

This should be merged only after Firefox ESR has the fix. Keeping the WIP state for now.

### DIFF
--- a/keepassxc-browser/popups/popup.css
+++ b/keepassxc-browser/popups/popup.css
@@ -4,10 +4,9 @@ body {
     display: flex;
     font-family: sans-serif;
     font-size: 14px !important;
-    overflow: hidden;
     padding-top: 8px;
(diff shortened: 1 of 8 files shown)


droidmonkey:
We should do a version detect and choose the right code path.

varjolintu:
This can be implemented now.

droidmonkey (approved):

PR #1773, opened 2022-11-07
Wait for stylesheets to load before displaying shadow DOM elements

stefansundin:
This fixes scrollbars appearing briefly on page load, and potentially causing the autocomplete menu to be shifted. I am experiencing this in Chrome. I tried to repro in Firefox but wasn't able to.

I have been trying to diagnose this weird offset issue a few times but haven't been able to figure it out until now. The autocomplete field would be offset like this on the Google sign-in page (but only the first time you opened the list!).

Screenshot 2022-11-06 at 19 33 35

As you can see, the list is shifted slightly to the left for some reason. Closing the list and opening it again puts it in the correct position.

Putting a `debugger` statement in `updatePosition` reveals the problem. When the position is calculated, a scrollbar has appeared on the page. If you scroll down you will see that the autocomplete list is rendered at the bottom of the page and is taking up a bunch space there, which is causing a scrollbar to be added. This scrollbar is affecting the position calculation which is why the list is shifted.

Screenshot 2022-11-06 at 19 58 27

The reason why I am adding the same code to the Username and TOTP wrappers is because they can also cause a scrollbar to appear before the stylesheet has loaded. Here's a screenshot showing that the wrapper for the username icon is 24 pixels in height and you can see a scrollbar in this screenshot (once the stylesheet has loaded this div is 0 pixels in height).

Screenshot 2022-11-06 at 19 40 03

I know there are a couple of more stylesheets being loaded but I only added this code to these three where I am positive there are negative effects. I don't really use the banners or notifications so I am not sure if that's a problem or not.

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -86,6 +86,8 @@ class Autocomplete {
             const styleSheet = createStylesheet('css/autocomplete.css');
             const colorStyleSheet = createStylesheet('css/colors.css');
             this.wrapper = kpxcUI.createElement('div');
+            this.wrapper.style.display = 'none';
+            styleSheet.addEventListener('load', () => this.wrapper.style.display = 'block');
             this.container = kpxcUI.createElement('div', 'kpxcAutocomplete-container', { 'id': 'kpxcAutocomplete-container' });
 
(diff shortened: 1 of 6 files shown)


stefansundin:
I was able to repro the same problem with the other shadow DOMs so I added the same code to them (except for pwgen).

varjolintu (approved):

PR #1781, opened 2022-11-15
Refactor how the URL is set when clicking the username-only button in the popup

stefansundin:
This is a fix for https://odysee.com/. The login page is https://odysee.com/$/signin and the username-only button adds `https://odysee.com/$/*` to the list of URLs. However, since this is interpreted as a regex, `$` is messing with the matching so the new preference doesn't actually take effect.

I'm still not sure if using regex to match the site is a good idea, but that might be a discussion for another time.

It is not clear to me if this code is currently working as intended or not:

https://github.com/keepassxreboot/keepassxc-browser/blob/2ea0fe2b97c9451c94bc517e0b7bd287556ae807/keepassxc-browser/content/keepassxc-browser.js#L46-L47

Is it supposed to convert `https://odysee.com/$/signin` to `https://odysee.com/$/*` or `https://odysee.com/*`?

I'm guessing the latter. I decided to refactor it using [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) but let me know if you prefer string manipulation instead.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -44,10 +44,13 @@ kpxc.addToSitePreferences = async function() {
 
     if (!siteExists) {
         // Add wildcard to the URL
-        site = site.slice(0, site.lastIndexOf('/') + 1) + '*';
+        const url = new URL(window.top.location.href);
+        url.pathname = '/*';
+        url.search = '';
+        url.hash = '';
 
         kpxc.settings['sitePreferences'].push({
-            url: site,
+            url: url.toString(),
             ignore: IGNORE_NOTHING,
             usernameOnly: true
         });


varjolintu:
The matching is based on [Match Patterns](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns) and previously Mozilla had a free code to use that changed these patterns to Regex. This is why it's used.

In this case the original code takes the last `/` and replaces everything after it with `*`so the path should be included.

varjolintu (changes requested):
Changes already requested in comments.

stefansundin:
Hi @varjolintu. Sorry but I don't understand what changes you want.

varjolintu:
> Hi @varjolintu. Sorry but I don't understand what changes you want.

Your change seems to replace the whole path with `/*` while the original code keeps the path in the URL and adds `*` after it.

stefansundin:
> Your change seems to replace the whole path with `/*` while the original code keeps the path in the URL and adds `*` after it.

Sorry for not responding earlier. Yes, this change was actually intended. It is the whole point of the PR. Refactoring how it is set is only because I think it is a better way to do it.

Without making this change in behavior, the Odysee website would still be broken.

If this change in behavior is not Ok then please feel free to close this PR. Maybe in that case it would be more useful to discuss options to move away from using regex to perform the matching? If it is desired to keep the regex functionality then maybe a new checkbox could be introduced to allow regex matching, and things that add to the list (like this button) would do so with the regex feature disabled. With the regex feature disabled it would still support a `*` wildcard at the end, but regex characters such as `$` would not be interpreted in a special way.

varjolintu:
After investigating this problem again, it would be enough to add the following addition to https://github.com/keepassxreboot/keepassxc-browser/blob/develop/keepassxc-browser/common/global.js#L127:
`.replace(/[!\^$\+\-\(\)@<>]/g, '\\$&');` which escapes any common extra HTML characters.

But I'd actually prefer if the whole matching could be done just using the `URL` objects, and maybe I'll refactor the whole thing later. But in the meanwhile that simple `replace()` could do the trick.

varjolintu:
This change probably fixes all the issues and makes this PR obsolete: https://github.com/keepassxreboot/keepassxc-browser/pull/1851

PR #1782, opened 2022-11-15
Try harder to prevent false-positive password change banners

stefansundin:
Try harder to identify the username used if the user have multiple entries for a website. This should help prevent false-positive banners when the user did *not* change their password, especially on multi-step login forms.

This has been a long-standing annoyance for me but I haven't really understood why it happens. Happens every time if you have multiple entries for a page _and_ the website is using a multi-page login form.

I tested this on https://odysee.com/ and I confirmed that the banner still showed up when I actually changed my password.

I was also considering changing
```javascript
if (kpxc.credentials.some(c => c.login === usernameValue && c.password === passwordValue)) {
```

to
```javascript
if (kpxc.credentials.some(c => (!usernameValue || c.login === usernameValue) && c.password === passwordValue)) {
```

But not sure if that's overkill or not.

### DIFF
--- a/keepassxc-browser/content/form.js
+++ b/keepassxc-browser/content/form.js
@@ -158,6 +158,15 @@ kpxcForm.onSubmit = async function(e) {
     } else if (kpxc.credentials.length === 1) {
         // Single entry found for the page, use the username of it instead of an empty one
         usernameValue = kpxc.credentials[0].login;
+    } else {
+        // Multiple entries found for the page, try to find out which one might have been used
+        const pageUuid = await sendMessage('page_get_login_id');
+        if (pageUuid) {
+            const credential = kpxc.credentials.find(c => c.uuid === pageUuid);
+            if (credential) {
+                usernameValue = credential.login;
+            }
+        }
     }
 
     // Check if the form has three password fields -> a possible password change form


keepassxc-browser/content/form.js
```diff
@@ -158,6 +158,15 @@ kpxcForm.onSubmit = async function(e) {
     } else if (kpxc.credentials.length === 1) {
         // Single entry found for the page, use the username of it instead of an empty one
         usernameValue = kpxc.credentials[0].login;
+    } else {
```
stefansundin, line 161: I just realized that maybe I should have used the following:

```suggestion
    } else if (kpxc.credentials.length > 1) {
```

Is there a chance that we get here when there's zero credentials?

varjolintu: In theory yes, if a login has been made in the previous page and the current page has a different URL that is not matched with the selected entry at all. A rare case though. Here the `username` value is of course only used for the banner.

If it's a first time user enters credentials to a page, there are no credentials (yet). This `else` should cover that also.

varjolintu (approved):

varjolintu:
Let's merge this. I'm not 100% sure about if this affects to some login pages, but if something strange comes up, that can be fixed later.

PR #1791, opened 2022-11-24
Update translations

### LINKED ISSUE #1790: Source code does not include Transifex settings needed to build
The source code does not include what a recipient needs to build the project.

This prevents building by a person who does not have a network connection, or whose computer cannot connect to the required online services, or who does not have accounts with the required online services.

    $ npm build
    
    > KeePassXC-Browser@1.8.3.1 build
    > node build.js
    
    Pulling translations from Transifex, please wait...

The build hangs indefinitely.

I assume this is because, as listed in https://github.com/keepassxreboot/keepassxc-browser/wiki/Loading-the-extension-manually

> Prerequisites
> […] Log into Transifex (e.g., with your GitHub account) […]

This causes the source code to be dependent on a Transifex account or GitHub account, in order to build. A person who has neither, or who cannot connect to those services for any reason, cannot continue the build according to the instructions.

Please ensure that a person who has the source code and the necessary build dependencies, can build the entire 'keepassxc-browser' from only the source code. (This is a requirement for people who wish to redistribute the work under GNU GPLv3.)

varjolintu:
Updates all translations for offline use. Also adds an optional `--skip-translations` parameter to build script.

Fixes #1790.

### DIFF
--- a/build.js
+++ b/build.js
@@ -40,7 +40,10 @@ async function updateTranslations() {
 }
 
 (async() => {
-    await updateTranslations();
+    if (process.argv.slice(2)[0] !== '--skip-translations') {
+        await updateTranslations();
+    }
+
     await fs.copyFile(`${DEST}/manifest.json`, `./${DEFAULT}`);
 
     for (const browser in BROWSERS) {
--- a/build.js
+++ b/build.js
@@ -40,7 +40,8 @@ async function updateTranslations() {
 }
 
 (async() => {
-    if (process.argv.slice(2)[0] !== '--skip-translations') {
+    const params = process.argv.slice(2);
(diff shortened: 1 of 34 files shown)


build.js
```diff
@@ -40,7 +40,10 @@ async function updateTranslations() {
 }
 
 (async() => {
-    await updateTranslations();
+    if (process.argv.slice(2)[0] !== '--skip-translations') {
```
droidmonkey, line 43: Should do the slicing above this statement (in case we add more flags in the future). Also should do a ".contains(...)" or similar type lookup.

droidmonkey (approved):

PR #1825, opened 2023-01-15
Fix extension behavior for Auto-Open

lodavid:
This makes the extension behave as expected in my set up, where unlocking the primary DB (which unlocks the child DBs automatically) keep the extension in an "unlocked" state.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -103,7 +103,7 @@ kpxc.detectDatabaseChange = async function(response) {
     kpxcIcons.switchIcons();
 
     if (document.visibilityState !== 'hidden') {
-        if (response.hash.new !== '' && response.hash.new !== response.hash.old) {
+        if (response.hash.new !== '') {
             _called.retrieveCredentials = false;
             const settings = await sendMessage('load_settings');
             kpxc.settings = settings;


keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -103,7 +103,7 @@ kpxc.detectDatabaseChange = async function(response) {
     kpxcIcons.switchIcons();
 
     if (document.visibilityState !== 'hidden') {
-        if (response.hash.new !== '' && response.hash.new !== response.hash.old) {
+        if (response.hash.new !== '') {
```
varjolintu, line 106: I found some strange behavior when changing this with two databases open and closing just one of them with "Search credentials from all databases" enabled. I think we need double check these messages also.

varjolintu: Plus KeePassXC sends `database-unlocked` notifications two times when a single database is locked and the active one is switched to that one that's still open. Need to investigate that too.

lodavid: **Without** this change, I noticed strange behavior under the scenarios you describe. For example, with 2 DBs open, closing one of them will cause the username icon to change to the grayed out locked state, even when the second open DB that is switched to has valid credentials.

With this change applied, at least on my set up, I get the correct behavior with 2 DBs open and then closing one of them, at least for the autofill icon behavior. If the caching check is needed, could try to make this fix more specific (e.g. only do the locking part of `detectDatabaseChange()` if `!(response.hash.new !== '' && response.hash.new !== response.hash.old)`)

I also found some bugs where closing the first DB won't remove the entries from the autofill list that pops up when clicking on the key icon on the username field. I believe the relevant part for that is `keepassxc-browser.js:596`, where the context menu only gets updated under that specific condition. For my entries, none of them have stringFields available, so it doesn't seem like it's valid to only update the context menu only when those are available. Can make a separate PR for that if desired.

varjolintu: I can also take a look at the second bug. Seems we both reproduced it.

varjolintu: Got the bug solved with updating the Autocomplete Menu. I'm making a PR of it.

lodavid: Thanks, with both of these PRs applied to my local repo both of those issues are solved.

varjolintu (approved):
Let's keep the `hash.new` and `hash.old` intact (not removing them). I have a few ideas how to improve this in the future.

lodavid:
Thanks again for helping get to the bottom of this. Please let me know if there's anything else I need to do here (seems like the merge must be done by some with the right permissions) or if I can be of any more help.

varjolintu:
I'll do the merge soon. Thanks!

PR #1835, opened 2023-01-24
Fix login popup change with locked database

varjolintu:
Fixes the old annoying bug that shows this kind of popup if you switch from open database to a closed one, and the open database had credentials:
Screenshot 2023-01-24 at 18 13 39

This change also fixes the position of "Reopen database" buttons after database lock has been made from the popup icon.

Steps to reproduce:
1. Have two databases in KeePassXC, open the first one.
2. Go to a page where the open database has credentials.
3. Switch from KeePassXC to the closed database.
4. The extension popup shows no login data, and no error is shown.


droidmonkey (approved):
Good catch!

Issue #818, opened 2020-03-22
Since version 1.6.0, custom text fields not filled anymore


olifre:
## Expected Behavior
While this worked fine for most sites before version 1.6.0, I can't make it work with any site as of now :-(. 

## Current Behavior
Custom fields ignored. 

## Steps to Reproduce (for bugs)
1. Visit any page requiring a custom text field. 
2. Have that in the database prefixed with "KPH:". 
3. Observe it's not filled. 

## Debug info

KeePassXC - 2.5.3
KeePassXC-Browser - 1.6.0 / 1.6.1 (both fail)
Operating system: Linux
Browser: Firefox

varjolintu:
And it doesn't help if you select String Fields again for the same page? I tested this with my test site and everything worked normally. Can you give the site URL?

olifre:
Reselecting did not help, I tried. Sadly, both pages I have at hand are internal only. 
Can you tell me a public test page? Maybe I can reproduce there.

varjolintu:
https://meine.deutsche-bank.de/trxm/db/ is a good page to test it because there are multiple fields. I usually select the last two for username and passwd, and the first two for String Fields.

olifre:
Got it, that's the cause:
https://github.com/keepassxreboot/keepassxc-browser/commit/d36d4f4c36bc8873d4ce7ea1e597fb4ca0eb4d66#diff-412b80e8f63d0bb8ff5dcc811ec72823
The field on my page is "readonly" since it is filled by Javascript and not by typing usually. However, KeepassXC could fill it just fine before.

olifre:
My page has something like this:
https://jsfiddle.net/cptcrunchy/ezs7n4o8/16/
in addition to username / password.

varjolintu:
The extension also showed icons for readOnly input fields that couldn't be filled. We'll maybe have to make a fix where this could be still allowed for String Fields. Still, I think `readonly` should be respected.

olifre:
It's sadly common practice on some pages to have such a javascript pin in addition to username / password, so I'd really love it if for string fields explicitly selected by the user, "readonly" could be ignored. For username / password, I agree of course.

droidmonkey:
Read-only is read only, we should not be filling those fields

olifre:
[...]

PR #1841, opened 2023-01-27
Do not fill readonly fields

### LINKED ISSUE #1840: Readonly input boxes are filled with credentials
## Expected Behavior

I'm on OWA, trying to change the password for my account. I see this form:

image

KeePassXC offers to have credentials auto-filled here, which seems convenient enough to have my current password entered there. When I click the KeePassXC icon, I would expect my current password to be filled.

## Current Behavior

When I navigate to the form for the first time, my user name is already filled in and the textbox is readonly.

When I click the icon, my current password is filled in, but it also replaced the `Domain\user name` part in the readonly text input above it.

I'm not sure if this is even a problem, but I log in on the page with `user@domain.tld` and this input box displays/expects `domain\user`, which might be conflicting.

## Possible Solution

Don't change values already present in readonly text inputs.

## Debug info

KeePassXC - 2.7.4
KeePassXC-Browser - 1.8.4
Operating system: Win64
Browser: Chrome/Chromium 109.0.0.0

varjolintu:
Add a simple check for readonly fields.

Fixes #1840.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -663,6 +663,10 @@ kpxc.setValue = function(field, value) {
 
 // Sets a new value to input field and triggers necessary events
 kpxc.setValueWithChange = function(field, value) {
+    if (field.readOnly) {
+        return;
+    }
+
     field.value = value;
     field.dispatchEvent(new Event('input', { bubbles: true }));
     field.dispatchEvent(new Event('change', { bubbles: true }));


droidmonkey (approved):

olifre:
Does this exclude fields actively selected by the user, i.e. the use case covered by https://github.com/keepassxreboot/keepassxc-browser/pull/878 (see https://github.com/keepassxreboot/keepassxc-browser/issues/818 )?

varjolintu:
@olifre Yes. This doesn't allow filling readonly fields even if those are selected via Custom Login Fields.

olifre:
@varjolintu Thanks for the quick reply!

So that means it actually breaks the use case described in https://github.com/keepassxreboot/keepassxc-browser/issues/818#issuecomment-602271323 : Filling readonly Custom Fields, which are usually filled by javascript via user interaction, e.g. pin entry fields with numbers to click on, is not possible anymore, correct?

varjolintu:
@olifre Seems so. Gotta fix this regression before the next release.

olifre:
@varjolintu Thanks, that will be indeed be helpful :+1: .

varjolintu:
@olifre I'd suggest that we allow filling String Fields using Custom Login Fields also to readonly fields, but not username/password.

olifre:
@varjolintu That seems very reasonable, I don't know any form which would mark username or password read-only but expect users to fill it indirectly via javascript. The only cases I do know are all additional Custom Login Fields (pin codes etc.), so the proposal to allow filling readonly String Fields iff they are Custom Login Fields would work well. :+1:

pdinc-oss:
Oh this breaks iDRAC v9 (Dell) logins

PR #1851, opened 2023-02-15
Replace Match Patterns regex with a simpler function

varjolintu:
Replaces the old Match Patterns regex code (originally copied from Mozilla's wiki page) with a simpler functions that are easier to debug. The matching is used for Predefined Sites and Site Preferences.

This should be faster with Predefined sites than the previous implementation because it's not necessary to do regex matching every time, but matching can return false sooner if scheme or host does not match.

The new implementation also allows more complex wildcarded URL's like:
- 'https://test.example.com/*/another/a/'
- 'https://example.com/$/*'
- 'https://example.com:8448/login/page'

The only downside for this change is that asterisk wildcards must be replaced with a unique string (`1kpxcwc1` in this case) because Chrome and Firefox browsers handle the `URL` object in a different way. For example `https://*.example.com` can be generated as a new URL object in Chrome, resulting an escaped `https://%2A.example.com`. Firefox however does not accept  `https://*.example.com` or even the escaped version of the URL for the new object, but throws an exception.

TODO:
- [x] Check `file://` scheme.
- [x] Check that Add Username-Only option from popup still works.
- [x] Simplify Predefined Sites URL list


keepassxc-browser/common/global.js
```diff
@@ -56,86 +56,54 @@ const ManualFill = {
...
-
-    const matchPatternRegExp = new RegExp(
-        `^${schemeSegment}://${hostSegment}/${pathSegment}$`
-    );
-
-    const match = matchPatternRegExp.exec(pattern);
-    if (!match) {
-        throw new TypeError(`"${pattern}" is not a valid MatchPattern`);
+    // If there's no wildcard with hostname, just compare directly
+    if (separator === '.' && !firstUrlPart.includes(URL_WILDCARD) && firstUrlPart !== secondUrlPart) {
```
droidmonkey, line 66: I am not a fan of the separator usage, it is really just a selector for different code paths, it would be better to make it a boolean value or split this function into two, one for hostnames one for paths.

droidmonkey (approved):

PR #1857, opened 2023-02-21
Allow force-fill to String Fields even if readonly

varjolintu:
Fixing a regression for #818 from change #1841. Filling `readonly` fields should be allowed if those are specified as String Fields using Custom Login Fields. That's the only exception. No username/password/totp will be allowed to do that.


droidmonkey (approved):

pdinc-oss:
Does not fix the regression from #1841 - re: Dell Enterprise applications

PR #1873, opened 2023-03-04
Use defaultView instead of scrollingElement when positioning icons and menus

### LINKED ISSUE #1872: Autocomplete positioning issue when document is in quirks mode
## Expected Behavior

The autocomplete dropdown should be properly positioned when a page is in quirks mode.

## Current Behavior

I noticed that on https://identity.pagerduty.com/, the autocomplete dropdown is not positioned properly. Turns out that the page is rendered in quirks mode and `document.scrollingElement` is `null`. (see below for javascript error)

Screenshot 2023-03-03 at 17 22 55

## Possible Solution

I tried this:

```javascript
const scrollingElement = document.scrollingElement || document.documentElement;

// replace references to document.scrollingElement with just "scrollingElement"
```

This works fine until you scroll the document. When the document is scrolled the dropdown is not following along. If this had fully solved the problem then I would have opened a PR.

1. Create an entry for https://identity.pagerduty.com/.
2. Go to https://identity.pagerduty.com/.
3. Click the username input field.
4. Open javascript console and see `scrollingElement` error (screenshot below).

## Debug info

KeePassXC - 2.7.4
KeePassXC-Browser - 1.8.5.1
Operating system: Mac
Browser: Chrome

Screenshot 2023-03-03 at 17 22 14

An errant `<meta>` tag is causing the page to render in quirks mode:

Screenshot 2023-03-03 at 17 27 17

```
document.compatMode
'BackCompat'
```

varjolintu:
With some webpages `document.scrollingElement` or `document.documentElement` are not providing correct `scrollTop` and `scrollLeft` values. Instead `document.defaultView` is used with `scrollX` and `scrollY` values.

Needs a little more testing to ensure this doesn't break any sites with scrolling.

Fixes #1872.

### DIFF
--- a/keepassxc-browser/content/ui.js
+++ b/keepassxc-browser/content/ui.js
@@ -143,12 +143,20 @@ kpxcUI.setIconPosition = function(icon, field, rtl = false, segmented = false) {
         top = iconOffset[1];
     }
 
-    const scrollTop = document.defaultView?.scrollY ?? document.scrollingElement?.scrollTop ?? 0;
-    const scrollLeft = document.defaultView?.scrollX ?? document.scrollingElement?.scrollLeft ?? 0;
+    const scrollTop = kpxcUI.getScrollTop();
+    const scrollLeft = kpxcUI.getScrollLeft();
     icon.style.top = Pixels(top + scrollTop + offset + 1);
     icon.style.left = rtl
                     ? Pixels((left + scrollLeft) + offset)
                     : Pixels(left + scrollLeft + field.offsetWidth - size - offset);
+};
+
+kpxcUI.getScrollTop = function() {
+    return document.defaultView?.scrollY ?? document.scrollingElement?.scrollTop ?? 0;
+};
+
+kpxcUI.getScrollLeft = function() {
+    return document.defaultView?.scrollX ?? document.scrollingElement?.scrollLeft ?? 0;
 };
 
 kpxcUI.getRelativeLeftPosition = function(rect) {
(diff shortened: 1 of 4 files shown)


droidmonkey:
Nothing like consistency in web development! This is why I quit making web pages back in 2010.

keepassxc-browser/content/ui.js
```diff
@@ -143,8 +143,8 @@ kpxcUI.setIconPosition = function(icon, field, rtl = false, segmented = false) {
         top = iconOffset[1];
     }
 
-    const scrollTop = document.scrollingElement ? document.scrollingElement.scrollTop : 0;
-    const scrollLeft = document.scrollingElement ? document.scrollingElement.scrollLeft : 0;
+    const scrollTop = document.defaultView ? document.defaultView.scrollY : document.scrollingElement.scrollTop;
+    const scrollLeft = document.defaultView ? document.defaultView.scrollX : document.scrollingElement.scrollLeft;
```
stefansundin, line 147: Could also do something like this:

```suggestion
    const scrollLeft = document.defaultView?.scrollX ?? document.scrollingElement?.scrollLeft ?? 0;
```

varjolintu: Gotta bump up the minimum Firefox version for these (67 -> 74 so no biggie). But those are a nice cleanup.

stefansundin: I think both `?.` and `??` are in use in small parts of the code already. I think I added the `?.` before I even considered compatibility.

varjolintu: If you already use browser greater than the minimum version there's no problems. But if you use e.g. Firefox 67 I bet those will not work and will throw an exception. A rare situation of course, because I doubt no one is using versions that old.

droidmonkey (approved):

PR #1877, opened 2023-03-06
Modify checks for page.tabs

### LINKED ISSUE #1860: initContentScript error: Error: Cannot read properties of undefined (reading 'errorMessage')
keepassxc-browser often gets unresponsive on web pages, in which situation the following error is shown in the console:

```
[Error keepassxc-browser.js:806] KeePassXC-Browser - initContentScript error: Error: Cannot read properties of undefined (reading 'errorMessage')
```

This might be related to #1355, but I don't face any possible causes described in that issue.

The problem often goes away for a specific page after a refresh; this makes me think it's a race condition. The error is apparently re-thrown [here](https://github.com/keepassxreboot/keepassxc-browser/blob/develop/keepassxc-browser/content/keepassxc-browser.js#L818), I'm not sure where it originates.

As this is happening quite frequently for me, but arguably not others (as I don't see a lot of related reports), this is perhaps caused by some other extension (like uBlock Origin). Yet, it's not the behaviour that should be shown in any circumstances.

KeePassXC - 2.7.4
KeePassXC-Browser - 1.8.2.2 (was actually troubling me over the past couple of years already)
Operating system: Linux
Browser: Chrome

varjolintu:
Modifies some checks for `page.tabs` array. Some changed to include optional chaining.

Fixes #1860.

### DIFF
--- a/keepassxc-browser/background/browserAction.js
+++ b/keepassxc-browser/background/browserAction.js
@@ -47,7 +47,7 @@ browserAction.showDefault = async function(tab) {
         tab = tabs[0];
     }
 
-    if (page.tabs[tab.id] && page.tabs[tab.id].loginList.length > 0) {
+    if (page.tabs[tab.id]?.loginList.length > 0) {
         popupData.iconType = 'questionmark';
         popupData.popup = 'popup_login';
     }
--- a/keepassxc-browser/background/event.js
+++ b/keepassxc-browser/background/event.js
@@ -23,7 +23,9 @@ kpxcEvent.showStatus = async function(tab, configured, internalPoll) {
         browserAction.showDefault(tab);
     }
 
-    const errorMessage = page.tabs[tab.id].errorMessage;
+    const errorMessage = page.tabs[tab.id]?.errorMessage ?? undefined;
+    const usernameFieldDetected = page.tabs[tab.id]?.usernameFieldDetected ?? false;
+
     return {
         identifier: keyId,
         configured: configured,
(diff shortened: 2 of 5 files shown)


droidmonkey (approved):

keepassxc-browser/background/browserAction.js
```diff
@@ -47,7 +47,7 @@ browserAction.showDefault = async function(tab) {
         tab = tabs[0];
     }
 
-    if (page.tabs[tab.id] && page.tabs[tab.id].loginList.length > 0) {
+    if (page.tabs[tab.id]?.loginList.length > 0) {
```
stefansundin, line 50: I _think_ these should also have a `?` before `.length`, i.e.:

```javascript
    if (page.tabs[tab.id]?.loginList?.length > 0) {
```

Otherwise this will happen:
```
({})?.loginList.length > 0
Uncaught TypeError: (intermediate value).loginList is undefined
```

varjolintu: Could you add this to #1983? Thanks.

stefansundin: Sure 👍

stefansundin: You know, I think I was wrong after all. Thanks to the [short circuiting feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#short-circuiting) in optional chaining, this is safe. My example was flawed, this is more accurate:

```javascript
({})[1234]?.loginList.length > 0
false
```

PR #1880, opened 2023-03-11
Code cleanup for 1.8.6

varjolintu:
Various code cleanups:
- Use `??=` and `&&=` assignments where it clears up the code.
- Use optional chaining `?.` instead of `if (value && value.something)` comparisons.
- Use `Object.hasOwn()` instead of `hasOwnProperty()` in some cases.
- Use negative indexing `.at(-1)` instead of `arr[arr.length — 1]`.
- Use simplified for loops when index is needed.
- Refactor default settings to use a loop based setter.

The Firefox minimum version must be increased to 91, which is the current max version Debian stable provides.

### DIFF
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -81,7 +81,7 @@ keepassClient.sendNativeMessage = function(request, enableTimeout = false, timeo
 
         const listener = ((port, action) => {
             const handler = (msg) => {
-                if (msg && msg.action === action) {
+                if (msg && msg?.action === action) {
                     // If the request has a separate requestID, check if it matches when there's no nonce (an error message)
                     const isNotificationOrError = !msg.nonce && request.requestID === msg.requestID;
 
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -644,7 +644,7 @@ keepass.migrateKeyRing = function() {
 };
 
 keepass.saveKey = function(hash, id, key) {
-    if (!(hash in keepass.keyRing)) {
+    if (!Object.hasOwn(keepass.keyRing, hash)) {
         keepass.keyRing[hash] = {
             id: id,
             key: key,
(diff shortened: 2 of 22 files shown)


keepassxc-browser/background/client.js
```diff
@@ -59,7 +59,7 @@ const messageBuffer = {
 
     matchAndRemove(msg) {
         for (let i = 0; i < this.buffer.length; ++i) {
-            if (msg.nonce && msg.nonce === keepassClient.incrementedNonce(this.buffer[i].nonce)) {
+            if (msg?.nonce === keepassClient.incrementedNonce(this.buffer[i].nonce)) {
```
droidmonkey, line 62: I think doing this would produce incorrect results. The `?.` operator returns `undefined` if nonce is nullish. So you could potentially compare undefined with undefined, which would evaluate to true.

Not saying that would happen in this instance, but if we blanket apply this operator everywhere it might be possible.

keepassxc-browser/background/keepass.js
```diff
@@ -691,7 +691,8 @@ keepass.deleteKey = function(hash) {
 keepass.getCryptoKey = function() {
     let dbkey = null;
     let dbid = null;
-    if (!(keepass.databaseHash in keepass.keyRing)) {
+
+    if (!Object.hasOwn(keepass.keyRing, keepass.databaseHash)) {
```
droidmonkey, line 695: The original `in` operator seems more readable, is there a specific reason for this change?

keepassxc-browser/background/client.js
varjolintu: True. We should keep the previous implementation just in case. This is the reason I didn't put this in every possible place.

keepassxc-browser/background/keepass.js
varjolintu: In this case we could also keep the previous one for readability. There's no actual reason why we should use the new one.

droidmonkey: Keep `in` then

droidmonkey (approved):

PR #1885, opened 2023-03-25
Improve message origin check for banner selector

palant:
The `topLevelDomainMatches()` function logic in `custom-fields-banner.js` is currently rather flawed. On `github.com` it will accept `https://github.co` as a valid message origin but not `https://abc.abc.github.com`.

While this allows malicious websites to send messages that circumvent the origin check, I think that the security impact here is low given the functionality triggered by these messages. Still, this needs fixing.

Note that my change does not address the concern that this code will consider `co.uk` to be a domain name. Fixing this requires using https://publicsuffix.org/.

### DIFF
--- a/keepassxc-browser/content/custom-fields-banner.js
+++ b/keepassxc-browser/content/custom-fields-banner.js
@@ -678,24 +678,11 @@ const topLevelDomainMatches = function(host) {
     }
 
     const originUrl = new URL(host);
-    const frameUrl = new URL(window.self.document.location.origin);
-    const urlParts = originUrl.host.split('.');
-    const dotCount = urlParts.length - 1;
+    const frameUrl = new URL(window.location.origin);
 
-    // Simple host like google.com, check directly
-    if (dotCount < 1) {
-        return false;
-    } else if (dotCount === 1) {
-        return frameUrl.host.includes(originUrl.host);
-    }
-
-    // Get the top-level-domain using counts of '.' but backwards, max 3.
-    // A basic host is like idmsa.apple.com, a more complex one like www.bbva.com.ar.
-    const index = Math.min(dotCount, 3);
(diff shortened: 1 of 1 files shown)


varjolintu:
I wish browsers could implement their own `topLevelDomain()` for URL's.

palant:
> I wish browsers could implement their own `topLevelDomain()` for URL's.

They do. And Mozilla even exposed it to extensions. And then their extension system was dropped in favor of Chrome’s. Which no longer exposes this functionality. 🤷‍♂️

> Maybe we could also use https://github.com/keepassxreboot/keepassxc-browser/blob/develop/keepassxc-browser/common/global.js#L116 but of course it doesn't solve the problem with every possible public suffix.

Better not. `com.works` is not a TLD. This regexp potentially makes things worse.

palant:
Either way, `psl.min.js` weights 134 kB. It can be added to the extension in order to solve this properly.

varjolintu:
As an addition we could add a randomly generated request ID to the messages (stored in background script) and keep the slightly faulty function intact. Not sure if it's needed to add a third party script file to the extension just because some very low security impact. After all this means the script file must be maintained and updated in every release.

palant:
If we are talking about the integrity of internal communication, `window.postMessage()` should not be used at all – random request IDs or not. That’s what `runtime.sendMessage()` is for. It’s less convenient because messages have to be proxied by the background page, but it is safe.

The decision whether third-party frames should ever be touched is independent of that. Usually that’s not advisable, though “belonging to same domain” might be an acceptable relaxation of the rule.

varjolintu:
You are absolutely right that `postMessage()` shouldn't be used even in this case. I made a PR for using `sendMessage()` instead, which makes the whole top level domain check obsolete:
https://github.com/keepassxreboot/keepassxc-browser/pull/1886

PR #1886, opened 2023-03-26
Transfer iframe messages without postMessage()

varjolintu:
Instead of unsafe `postMessage()` all messages between iframes should be transferred using `sendMessage()`.

For this a new event `frame_message` is added to `event.js` which only bounces the message back to the current tab's frames. After that `keepassxc-browser.js` handles the internal message (`frame_request_to_frames` or `frame_request_to_parent`) and calls the correct functions based on top window or iframe status. It's possible to use the same `frame_message` for other similar situations in the future.

### DIFF
--- a/keepassxc-browser/content/custom-fields-banner.js
+++ b/keepassxc-browser/content/custom-fields-banner.js
@@ -117,7 +117,7 @@ kpxcCustomLoginFieldsBanner.create = async function() {
         = kpxc.settings[DEFINED_CUSTOM_FIELDS] && kpxc.settings[DEFINED_CUSTOM_FIELDS][location]
             ? 'inline-block' : 'none';
     if (window.self !== window.top && kpxcCustomLoginFieldsBanner.buttons.clearData.style.display === 'inline-block') {
-        sendMessageToParent('enable_clear_data_button');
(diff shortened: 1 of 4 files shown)


droidmonkey (approved):

droidmonkey:
Yet another wonderful treat of webpage/browser development.

keepassxc-browser/content/custom-fields-banner.js
```diff
@@ -626,138 +618,120 @@ kpxcCustomLoginFieldsBanner.getNonSelectedElements = function() {
...
-const handleParentWindowMessage = function(e) {
-    if (!topLevelDomainMatches(e.origin)) {
+kpxcCustomLoginFieldsBanner.handleParentWindowMessage = function(args) {
+    if (!kpxcCustomLoginFieldsBanner.created) {
         return;
     }
 
-    if (e.data === 'username_button_clicked') {
+    const e = {};
+    e.isTrusted = args?.[1];
```
varjolintu, line 658: This is the only value that's needed, but some wrapping is needed so it's compatible with an actual Event. The whole event cannot be cloned and sent to background script.

PR #1909, opened 2023-05-02
Improvements to segmented TOTP field detection

varjolintu:
Improves the segmented TOTP field detection.
In some cases the extension has already identified the segmented fields as separate TOTP fields to a six different `combination` objects. This situation is not currently handled at all.

The solution identifies if six TOTP fields are in the combination, changes that to a single combination object with `totpInputs` and removes the six field from the original combination array.

Related issue in comment: https://github.com/keepassxreboot/keepassxc-browser/issues/1358#issuecomment-1529743424

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -56,10 +56,8 @@ kpxcFields.getAllCombinations = async function(inputs) {
         combinations.push(combination);
     }
 
-    // Check for multiple segmented TOTP fields
-    if (combinations.length === 0) {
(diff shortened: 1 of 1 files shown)


droidmonkey (approved):

droidmonkey:
Do you have a couple good test sites for this?

varjolintu:
> Do you have a couple good test sites for this?

No. Just the local code mentioned in the issue. I'll probably make an account to Posteo to test it out.

varjolintu:
I'll probably do some tests for the segmented TOTP fields. Currently there's none for those.

Issue #465, opened 2019-04-06
Optional monochrome (b/w) favicon to fit into browser-design 


rugk:
For Firefox, it would be nice to switch to a black/white (monochrome) version for the browser icon, so it fits into the other icons of your browser.

I'll mostly talk about Firefox, here, e.g. look how it looks when you have some adjusted add-ons and current Firefox icons:

![image](https://user-images.githubusercontent.com/11966684/55669932-970eda00-587d-11e9-83c8-19eb33191131.png)

(Nightly screenshot)

---

See https://design.firefox.com/photon/visuals/color.html#icons-and-other-elements for how it should be done.

---

I could take this on, if you want. :smiley:

whisdol:
I like the idea! Maybe this could also switch the autocomplete color back to monochrome (white?), for people who don't like the new no_jquery look.

sebast889:
I'm not a fan of this. Monochrome icons won't really fit into other icons better if you use the default/lighter themes. Majority of users don't use custom CSS or dark theme and unless there is a way to change icons automatically based on themes, the current colored icons will still look much better imo.

rugk:
It works for light and dark themes. And, personally, I would always want a **switch** in the settings, so this here should be optional, for users who don't want colored icons.

> autocomplete color

I don't know what you mean, but I rather think this is off-topic.

varjolintu:
The only problem with this suggestion is that in the current version monochrome icon means there's no connection to KeePassXC (with one exception I still haven't fixed though..), and the coloured one means the connection is currently active.

If this is implemented, we have to find another way to use different statuses in the icon, or just not use them at all, which doesn't seem like a good idea.

rugk:
Ah, status things is something yes. You could use colored badges for that somehow, or don't use it at all. (or only an error icon in the corner if there is an error; and otherwise just the monochrome version)

varjolintu:
I'll make some testing. Maybe it's possible to use a monochrome icon with more contrast when connection is active etc.

rugk:
No, not more contrast, then the color changes again and that should be fixed. As per [Firefox Photon guide](https://design.firefox.com/photon/visuals/color.html#icons-and-other-elements) the color should be `rgba(12, 12, 13, 0.8)` for the light icon version and `rgba(249, 249, 250, 0.8)` for the dark one.

[...]

PR #1938, opened 2023-07-13
Add optional monochrome toolbar icon

dynobo:
Would solve #465.

I'd be happy to receive your feedback and incorporate potential changes.

If anyone has an idea how to apply dark/light changes on the fly (currently a restart is required), that would be great. But such a feature could be shipped independently as a separate PR, I guess.

Screenshot of dark/light variants and setting:


varjolintu:
First of all, if you look at the original cross icon:
https://github.com/keepassxreboot/keepassxc-browser/blob/34f3e01d304db8d6b68c5215b322af4982672dc7/keepassxc-browser/icons/toolbar/colored/icon_cross.png

And then compare that to the ones in this PR, the icons are not the same. The cross has no background, and it's even done in a different size. Doing this properly would be to use the same icons and just change the saturation, for example.

https://github.com/keepassxreboot/keepassxc-browser/blob/34f3e01d304db8d6b68c5215b322af4982672dc7/keepassxc-browser/icons/toolbar/light/icon_cross.png

https://github.com/keepassxreboot/keepassxc-browser/blob/34f3e01d304db8d6b68c5215b322af4982672dc7/keepassxc-browser/icons/toolbar/dark/icon_cross.png

dynobo:
> Doing this properly would be to use the same icons and just change the saturation, for example.

I see four arguments _against_ just de-saturating the existing icons (ranked by impact):
1) It would loose the colors of the badges (lock, cross,...), which seem like an important visual cue for the user.
2) It would only work for the light scheme. A dark theme would need further editing/adjustments.
3) Keeping the same amount of detail (gradient, outlines) but removing the color information makes the details smudging together.
4) The icons would still look out of place from the flat Firefox icons.

That said: I'm not a designer and I am happy about any thoughts on the icon. E.g. the comment regarding the red cross is super valid! I'd love to collect more feedback on the visual, and can prepare another version later. If some designer could contribute to this PR: even better!

Besides the optics, do you think the functional implementation is fine and worth pursuing?

droidmonkey:
The contrast ratio of the monochrome icons is not in alignment with our accessibility goals. Dark background with red question mark is very hard to discern.

dynobo:
How would you feel about something like this?

more-contrast

It uses the stronger colors from the original. The "deactivated"
 variants from the original could be indicated e.g. by dimming the icon a bit.

droidmonkey:
I like that more

dynobo:
The corresponding (deactivated) bang icon would look like:
bang

The new badge could also have a more intense color, and potentially an outline. Also the boldness can be varied:
new-badge


I could also try white on green background for that badge, but it would then be more different from the original.

varjolintu:
The outlines are a must because there's not enough contast without those.

dynobo:
Based on your feedback, here's a version with outlines and red background boxes:

next-variant

What do you think?

droidmonkey:
Those look great, what are the last two icons for?

It might be worth putting a border on the read background overlays as well. It looks great on the lock overlay and the green star.

varjolintu:
We don't actually use the two last icons. I'll remove those.

dynobo:
Here's the variant with outline on red boxes, too:
next-variant

BTW: is there a reason why you are using small PNGs for the toolbar icon instead of SVGs which probably render nicer?

varjolintu:
> Here's the variant with outline on red boxes, too: next-variant
> 
> BTW: is there a reason why you are using small PNGs for the toolbar icon instead of SVGs which probably render nicer?

Chromium/Chrome still does not support SVG's for toolbar icons (The "[bug](https://bugs.chromium.org/p/chromium/issues/detail?id=29683)" has been open since 2009). We could make Firefox build to use SVG images though.

droidmonkey:
Wow those are great! I agree on the SVG front, if possible.

varjolintu:
> Wow those are great! I agree on the SVG front, if possible.

I'll make those changes to the same PR that removes the obsolete icons.

Also, maybe we could add the same outlines to the normal icons? Those look nice.

dynobo:
Light theme (dark icons) should be ready now, as well as the logic to use SVG on Firefox. 

Could you give it a quick look before I create the other themes? 
It's quite some effort to export all those variants, so I'd rather not do it too many times... :wink: 

Todo's for me: 
- [ ] Update dark theme
- [ ] Update colored theme

dynobo:
> And if you have the SVG's saved for the icons, can you share them? Thanks :)

For now I've added the [icons' source-SVG](https://github.com/keepassxreboot/keepassxc-browser/blob/5b56269c970d65e0f4b73e46faec625c734d2d99/keepassxc-browser/icons/toolbar/icons_source.svg) along with a [short how-to](https://github.com/keepassxreboot/keepassxc-browser/blob/5b56269c970d65e0f4b73e46faec625c734d2d99/keepassxc-browser/icons/toolbar/readme.md) to the toolbar-icons root folder. 

I'm not sure, if this your preferred way. Certainly convenient for developers. On the otherhand, I guess it will be shipped with the extension, which might not be desired.

varjolintu:
> > And if you have the SVG's saved for the icons, can you share them? Thanks :)
> 
> For now I've added the [icons' source-SVG](https://github.com/keepassxreboot/keepassxc-browser/blob/5b56269c970d65e0f4b73e46faec625c734d2d99/keepassxc-browser/icons/toolbar/icons_source.svg) along with a [short how-to](https://github.com/keepassxreboot/keepassxc-browser/blob/5b56269c970d65e0f4b73e46faec625c734d2d99/keepassxc-browser/icons/toolbar/readme.md) to the toolbar-icons root folder.
> 
> I'm not sure, if this your preferred way. Certainly convenient for developers. On the otherhand, I guess it will be shipped with the extension, which might not be desired.

Maybe we could add a `dev-resources` folder to the root of the repo. Other relevant content could be added there.

dynobo:
Okay, I now included the complete sets, included the updated original one. SVGs as well as PNGs.

For reference, this is original set:
original

Colored (slightly updated original set):
colored

Light:
light

Dark:
dark

Is anything else necessary?

varjolintu:
How would the dark set look like with black outlines?

droidmonkey:
Love it, but I agree a dark border (like #333333) on the dark set would probably look better

keepassxc-browser/background/browserAction.js
```diff
@@ -52,8 +52,13 @@ browserAction.generateIconName = function(iconType) {
     let name = 'icon_';
     name += (keepass.keePassXCUpdateAvailable()) ? 'new_' : '';
     name += (!iconType || iconType === 'normal') ? 'normal' : iconType;
-
-    return `/icons/toolbar/${name}.png`;
+    
+    let style = 'colored'
+    if (page.settings.useMonochromeToolbarIcon) {
+        style = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
```
varjolintu, line 58: This seem to only apply to the System theme now, right? If I manually switch to Light theme in the extension settings, the dark icons will be still used.

keepassxc-browser/_locales/en/messages.json
```diff
@@ -1199,4 +1207,4 @@
         "message": "Lock database",
         "description": "Lock database button title text."
     }
-}
+}
```
varjolintu, line 1210: What's happening here?

dynobo: White space change by my auto-formatter. I'll revert that.

keepassxc-browser/background/browserAction.js
dynobo: Good catch! I haven't thought of the theme setting...

dynobo:
Regarding the badge-outlines on the dark theme, I have some more variants. Left to right: 
- black 
- cut-out (more effort)
- white 80% opacity (as before, for reference)
- white 100%
light

(on gray background, black is a bit misleading)

keepassxc-browser/_locales/en/messages.json
dynobo: Fixed in [c1eb96b](https://github.com/keepassxreboot/keepassxc-browser/pull/1938/commits/c1eb96b4afd80aeb696b3ff68edbc9a0ed838543)

keepassxc-browser/background/browserAction.js
dynobo: Fixed in [86bdfe4](https://github.com/keepassxreboot/keepassxc-browser/pull/1938/commits/86bdfe441368fd0838cd3b5f499f0c7eb19fd7d0)

varjolintu:
The first one has the most contrast, which looks best, at least for me.

droidmonkey:
Agreed first one

dynobo:
So, then black outline it is, for all variants. This also eases the maintenance a bit. 

I also was able to simplify the SVG (by removing some unnecessary transforms), which also reduces their file-size.

Original:
original

Colored:
colored

Dark:
dark

Light:
light

(For the record, the cut-out version would have been my favorite as it looks a bit more minimalistic, but I agree, it also has a little bit less contrast.)

droidmonkey:
Those are really nice, thank you!

varjolintu:
> Those are really nice, thank you! @varjolintu let's add credit to @dynobo for his work in the about/settings dialog

Sure :) There's already the Contributors list. I'd like to modify that a little bit and add anyone who does big/relevant PR's, unless they want to stay "hidden".

dynobo:
> Can I add you to the contributors list?

Sure, thanks!

keepassxc-browser/_locales/en/messages.json
```diff
@@ -811,6 +815,10 @@
         "message": "Show notifications for errors and when user interaction is required.",
         "description": "Show notifications option help text."
     },
+    "optionsUseMonochromeToolbarIconHelpText": {
+        "message": "Use a monochrome icon with colored badges in toolbar that adapts to light/dark browser theme. Requires browser restart.",
```
varjolintu, line 819: One final fix to this: at least for me browser restart is no longer required. If the system theme is used, the icon will change after a tab is switched. With a specified dark/light theme the icons change instantly. So maybe this part of the text could be removed.

dynobo: Oh, true! Seems like when "System" theme is enabled, a restart is not necessary. However, on my system it still required an _interaction_ with the icon (e.g. hover or click) to change it from dark/ligh to colored or vice versa.

How about I change it to "(Might require browser restart)"?

dynobo:
Oh, and I noticed, that in the dark theme the icons are not _quite_ white enough, but a tiny bit darker than expected. 
Let me just fix that before merging, too.

keepassxc-browser/_locales/en/messages.json
dynobo: Just added "might" here: [d9c0082](https://github.com/keepassxreboot/keepassxc-browser/pull/1938/commits/d9c008209faea6b16978e411f8b4550faf1982ad)

phoerious:
Please don't invert the background. There are dedicated inverse monochrome icons in this repository, which you can base your own designs on.

dynobo:
> There are dedicated inverse monochrome icons in this repository

You mean the de-saturated grayscale versions, right? If yes, please see [my response](https://github.com/keepassxreboot/keepassxc-browser/pull/1938#issuecomment-1634121023) to [varjolintu's comment](https://github.com/keepassxreboot/keepassxc-browser/pull/1938#issuecomment-1634096520), who had the same objection.

phoerious:
No, there are also actual monochrome version of the icon. We use them for the tray icon. The desaturated one is only for the locked state if you use the colourful icon.

dynobo:
Oh, cool! That would indeed be better! Do you have a link for me? I somehow can't find it...

phoerious:
You should also adjust the colour scheme of the badges. The colours are too pure and clash with our CD.

dynobo:
> You should also adjust the colour scheme of the badges.

Do you have a reference for the color values? Then I don't need to guess. 
(I basically tried to copy the values from the badges of the _current_ toolbar icons)

Also, do you have some ideas regarding the badges' outlines? I'm still not convinced by the black outline...

dynobo:
Okay, so here are the 2 available official icons:
dark

Following this design would result in such icons:
dark

Those could be colored along the official color scheme, and would then look like:
dark

As far as I can tell, there is no "deactivated" state in the official theme, but of course we could again dim the icon:
dark

Those "following the brand" seem quite different from what we align on before... What do you think?

droidmonkey:
Yah those look great, I personally have no preference.

dynobo:
Okay, so the latest changes include the dark and light version, re-done with offical brand style:

dark

light

I _could_ also adjust the badge colors of the "colored" scheme to the official colors. That, however would bring them further away of what they look like _right now_. Shall I do it anyway?

varjolintu:
Hmm. Now the dark outlines are gone from the light versions. I liked those ones more. What do you think @phoerious? I think it improves the contrast quite a bit.

phoerious:
> I could also adjust the badge colors of the "colored" scheme to the official colors. That, however would bring them further away of what they look like right now. Shall I do it anyway?

Yes, please do so.

As for the contrast issue, you could try a dark outline on the light version or simply reduce the lightness of the badges.

dynobo:
> As for the contrast issue, you could try a dark outline on the light version or simply reduce the lightness of the badges.

I'll try darken the badges of the light theme a bit first, as the outlines are not used anywhere in the official themes I found.

I recently saw that the new Thunderbird version uses dark-color for as outline for badges to increase the contrast:
Screenshot from 2023-07-19 19-41-12

Could try that style, too.

dynobo:
From left to right:
1. Previous version
2. Darked badges
3. More darkend badges
4. Colored outline
5. Darker colored outline
6. Black outline

light

phoerious:
Second from the right.

droidmonkey:
I like the farthest right, but can live with the second from right. Maybe an inbetween?

varjolintu:
First from the right looks good to me.

dynobo:
Can we go with one of the two right ones (5. or 6.)? 
I'm a bit worried to change the locks main color, as it then is different again from the brand colors...

dynobo:
Here's the colored version with the adjusted badge colors:

colored

dynobo:
Here's the status after last commit:

light

dark

colored

dynobo:
Any ideas, how I could ensure the toolbar icon is reloaded after toggling "Use monochrom toolbar icon" in the settings? To avoid the current caveat of "Might require browser restart"?

varjolintu:
> Any ideas, how I could ensure the toolbar icon is reloaded after toggling "Use monochrom toolbar icon" in the settings? To avoid the current caveat of "Might require browser restart"?

I think the correct way is to add an exception to this if-else list: https://github.com/keepassxreboot/keepassxc-browser/blob/develop/keepassxc-browser/options/options.js#L87
And then send a command (for example `update_popup`). A new event must be added to `background/event.js` that calls `page` script.

Basically, add:
```
else if (name === 'useMonochromeToolbarIcon') {
    browser.runtime.sendMessage({ action: 'update_popup' });
}
```

And to `background/page.js`:
```
page.updatePopup = function(tab) {
    browserAction.showDefault(tab);;
}
```

And to `background/event.js` to the end of event list:
```
'update_popup': page.updatePopup
```

dynobo:
Thanks, @varjolintu . :pray: 
Done.

varjolintu:
I think everything looks really good now. Something still missing, @phoerious?

phoerious:
You might enable the pixel preview in your graphics editor with a 16x16px grid and check the badges for optimum pixel alignment, but I suppose other than that it's fine. More adjustments can be done later.

varjolintu:
FYI: You have probably noticed that the `!` icon is not used anywhere. I'll make a separate PR that shows it when the selected database is not connected and user action for a new connection is required.

dynobo:
Thanks for your suggestion, @phoerious. Before I optimize for that, one more question:
I was wondering about the size of **36x36px** that is currently used. Is this _really_ the correct size for the rasterized toolbar icons?

(I'm much more familiar with icon size like 16, 24, 32, 48, ...)

varjolintu:
> Thanks for your suggestion, @phoerious. Before I optimize for that, one more question: I was wondering about the size of **36 x 36px** that is currently used. Is this _really_ the correct size for the rasterized toolbar icons?
> 
> (I'm much more familiar with icon size like 16, 24, 32, 48, ...)

Currently Chrome/Chromium has only 16x16 and 32x32 [icon sizes](https://developer.chrome.com/docs/extensions/reference/browserAction/#icon) for the toolbar icon (SVG's are only supported in Firefox). If the icon is larger, it will be scaled. So the 38x38 is actually a bit too large. The page linked says: "You can provide any size icon to be used in Chrome, and Chrome will select the closest one and scale it to the appropriate size to fill the 16-dip space.", so I think we could provide larger icons (e.g. 64x64 or even 128x128) by default and let the browser handle the scaling.

dynobo:
Okay, so which size would you prefer for now? 64x64?

varjolintu:
> Okay, so which size would you prefer for now? 64x64?

I think 64x64 is fine for now. Hopefully Chromium will start to support SVG'n within this decade :)

dynobo:
Here's the status of last commit:
(64x64 in the commit, scaled down here to 36x36, for better comparison with previous versions):

colored

light

dark

dynobo:
Fix two more findings.

Current state:

colored

colored

light

varjolintu (approved):

droidmonkey (approved):

PR #1940, opened 2023-07-16
Handle errors properly when waiting for response

varjolintu:
If there's an error message returned when waiting for response, the listener should be destroyed. Applies at least to Password Generator widget when it is closed and an error message is received.

Some more testing needed to ensure this doesn't break anything.

### DIFF
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -83,7 +83,7 @@ keepassClient.sendNativeMessage = function(request, enableTimeout = false, timeo
             const handler = (msg) => {
                 if (msg && msg?.action === action) {
                     // If the request has a separate requestID, check if it matches when there's no nonce (an error message)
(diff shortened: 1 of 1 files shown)


droidmonkey (approved):

droidmonkey:
Seems this would only produce undefined behavior if it is an error message and it contains a nonce. That is already undefined behavior though based on the original code.

varjolintu:
> Seems this would only produce undefined behavior if it is an error message and it contains a nonce. That is already undefined behavior though based on the original code. Do we even care if it has a nonce or not if it has an error message/code?

If there's a nonce we assume the message is a direct reply to some request. If we are having an error reply in this kind of situation, the message is never resolved and it can cause annoying "Nonce compare failed" situations.

In Protocol V2 draft this kind of problem does not exist because every request has a unique request ID.

droidmonkey:
ok save any fixes / improvements to my statement to protocol V2. Agree the request ID is the key here.

PR #1961, opened 2023-08-15
Refactor extension messaging

varjolintu:
This change refactors the way messages and stored to the message buffer, and how the response message is being waited and handled.

Previously the extension created a new listener for every command sent via `.addListener()`, and after a proper response `.removeListener()` was used after a response. In some cases the response is not received, or there is only a generic error message available that cannot be identified as a response to any message. When these messages (and listeners) pile up, the extension will be shown `Nonce compare failed` error messages and the connection to KeePassXC might get broken. In thise case KeePassXC' native messaging handler can show multiple messages waiting inside one socket message, and the messages are not handled properly.

This is been fixed by creating a new custom `Message` class that stores the `Promise` needed to be wait for the response, without creating any new listeners to the native messaging port. Instead, when sending a new message, new `Message` class is stored to a buffer, and the native messaging's `onMessage()` handler goes through the buffer and finds the corresponding `Promise` to be resolved. After that the `Message` will be removed from the buffer, and the response is returned to the command function that awaits it.

The change fixes password generator popups and possibly other situations where only locking and unlocking the database (or reconnecting to KeePassXC) has been the only workaround.

Minimum version of Firefox has been increased from 93 to 96 because of [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API).

### DIFF
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -56,7 +56,7 @@ const messageBuffer = {
     },
 
     removeMessageFromIndex(index) {
-        if (this.buffer.length >= index + 1) {
+        if (index >= 0 && index < this.buffer.length) {
             this.buffer.splice(index, 1);
         }
     }
@@ -95,17 +95,16 @@ class Message {
 //--------------------------------------------------------------------------
 
 keepassClient.sendNativeMessage = async function(request, enableTimeout = false, timeoutValue) {
+    if (!keepassClient.nativePort) {
+        return;
+    }
+
     const message = new Message(request, enableTimeout, timeoutValue);
     await navigator.locks.request('messageBuffer', async (lock) => {
         messageBuffer.addMessage({ request: request, message: message });
     });
 
-    // Send the request
-    if (keepassClient.nativePort) {
-        keepassClient.nativePort.postMessage(request);
-    }
-
-    // Wait for response
+    keepassClient.nativePort.postMessage(request);
     return await message.promise;
 };
 
@@ -115,7 +114,7 @@ keepassClient.handleNativeMessage = async function(response) {
     // Parse through the message buffer to find the corresponding Promise.
     await navigator.locks.request('messageBuffer', async (lock) => {
         for (let i = 0; i < messageBuffer.buffer.length; ++i) {
-            if (! messageBuffer.buffer[i]) {
+            if (!messageBuffer.buffer[i]) {
                 continue;
             }
 
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -51,80 +51,85 @@ const kpErrors = {
 const messageBuffer = {
     buffer: [],
 
-    addMessage(msg) {
-        if (!this.buffer.includes(msg)) {
-            this.buffer.push(msg);
-        }
+    addMessage(message) {
+        this.buffer.push(message);
     },
 
-    matchAndRemove(msg) {
-        for (let i = 0; i < this.buffer.length; ++i) {
-            if (msg.nonce && msg.nonce === keepassClient.incrementedNonce(this.buffer[i].nonce)) {
-                this.buffer.splice(i, 1);
-                return true;
-            }
+    removeMessageFromIndex(index) {
+        if (this.buffer.length >= index + 1) {
+            this.buffer.splice(index, 1);
         }
-
-        return false;
     }
 };
+
+// Basic class for a message to be sent. The Promise inside the class will be resolved when
+// the response to the message is received.
+class Message {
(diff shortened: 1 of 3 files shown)


varjolintu:
We could also use https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API to secure the handling of `messageBuffer`.

droidmonkey:
What is the LTS version

varjolintu:
> 

ESR's are 91 and 102, so we should be good. I think even Debian stable provides 102 now.

droidmonkey:
Easy choice to me

keepassxc-browser/background/client.js
```diff
@@ -51,78 +51,87 @@ const kpErrors = {
...
     },
 
-    matchAndRemove(msg) {
-        for (let i = 0; i < this.buffer.length; ++i) {
-            if (msg.nonce && msg.nonce === keepassClient.incrementedNonce(this.buffer[i].nonce)) {
-                this.buffer.splice(i, 1);
-                return true;
-            }
+    removeMessageFromIndex(index) {
+        if (this.buffer.length >= index + 1) {
```
droidmonkey, line 59: Usually this would be written as `0 >= index < this.buffer.length`

keepassxc-browser/background/client.js
```diff
@@ -51,78 +51,87 @@ const kpErrors = {
...
+keepassClient.sendNativeMessage = async function(request, enableTimeout = false, timeoutValue) {
+    const message = new Message(request, enableTimeout, timeoutValue);
+    await navigator.locks.request('messageBuffer', async (lock) => {
+        messageBuffer.addMessage({ request: request, message: message });
+    });
 
-        // Store the request to the buffer
-        messageBuffer.addMessage(request);
+    // Send the request
+    if (keepassClient.nativePort) {
```
droidmonkey, line 104: If this is false shouldn't an error be returned immediately?

keepassxc-browser/background/keepass.js
```diff
@@ -168,18 +169,24 @@ keepass.retrieveCredentials = async function(tab, args = []) {
 
 keepass.generatePassword = async function(tab) {
     if (!keepass.isConnected) {
-        return [];
+        return '';
+    }
+
+    if (keepass.passwordRequestSent) {
+        return undefined;
```
droidmonkey, line 176: Technically we check for this condition on the keepassxc side and just focus the existing password generator. I kind of like that behavior.

varjolintu: I don't like this either, but then we'll have to accept that with Firefox some messages are left in the buffer until browser is closed. I don't know why, but sometimes no messages are returned from KeePassXC.
Also, there's a possibility that this boolean stays false if some error occurs in the extension, so I'm removing this. Let's add some kind of checking for this later if actually needed.

varjolintu: Plus I need to modify the return value to an Object so we can properly identify when an error message is needed to show.

varjolintu:
I have been using this branch for a while, and it works much better than the current implementation. There's one strange situation remaining where the messages still goes out of sync, but that will be resolvable in Protocol V2.

keepassxc-browser/background/client.js
```diff
@@ -51,78 +51,86 @@ const kpErrors = {
...
-                    error: kpErrors.getError(kpErrors.TIMEOUT_OR_NOT_CONNECTED),
-                    errorCode: kpErrors.TIMEOUT_OR_NOT_CONNECTED
-                };
-                keepass.isKeePassXCAvailable = false;
-                ev.removeListener(listener.handler);
-                resolve(errorMessage);
-            }, messageTimeout);
-        }
+keepassClient.sendNativeMessage = async function(request, enableTimeout = false, timeoutValue) {
+    if (!keepassClient.nativePort) {
```
droidmonkey, line 98: Could maybe write an error to a log here

keepassxc-browser/background/client.js
```diff
@@ -51,78 +51,86 @@ const kpErrors = {
...
+    keepassClient.nativePort.postMessage(request);
+    return await message.promise;
+};
+
+keepassClient.handleNativeMessage = async function(response) {
+    const isError = Boolean(!response.nonce && response.error && response.errorCode);
+
+    // Parse through the message buffer to find the corresponding Promise.
+    await navigator.locks.request('messageBuffer', async (lock) => {
+        for (let i = 0; i < messageBuffer.buffer.length; ++i) {
```
droidmonkey, line 116: Could use a logged error if there is no matching message found (ie, the loop exits after exhausting the buffer)

droidmonkey (approved):

PR #1966, opened 2023-08-21
Remove internal password generator

varjolintu:
Removes the internal password generator and only leaves the support to launch KeePassXC's generator.
The feature requires at least KeePassXC 2.7.0, and some older Linux distros still provide only the 2.6.2 by default, but they just should update manually to the latest KeePassXC.


droidmonkey:
Yup plenty of options to get the latest version of keepassxc onto their legacy distribution.

droidmonkey (approved):

PR #1974, opened 2023-08-30
Fill whole combination when Autocomplete Menu is disabled

### LINKED ISSUE #1973: Filling with Autocomplete Menu disabled does not fill the whole combination
## Expected Behavior
I expect to fill usernames and passwords.

## Current Behavior
Filling of logins is not working well since I upgraded recently.

1. It works from the toolbar button
2. If I use the keyboard shortcut or the right-click menu for "Fill username and password" then it fills the username into wherever my cursor is. If there are multiple logins it does not show a menu it just picks one.
3. If I do "Fill password only" then it puts the password into wherever my cursor is, even if that's the username field. Again it doesn't show a menu.

As above. Here are some of my settings:

- OFF Activate username field icons.
- OFF Activate autocomplete for username fields
- ON Auto-submit login forms

## Debug info
KeePassXC - 2.7.6
KeePassXC-Browser - 1.8.7
Browser: Mozilla Firefox 117.0

varjolintu:
If Autocomplete Menu is disabled from the extension settings, only active element is filled with username or password, not the whole combination.

Note: Filling only the active element should be done only if a combination is not detected at all (for example single username field that is not automatically detected).

Fixes #1973.

### DIFF
--- a/keepassxc-browser/content/fill.js
+++ b/keepassxc-browser/content/fill.js
@@ -30,7 +30,7 @@ kpxcFill.fillInFromActiveElement = async function(passOnly = false) {
         return;
     }
 
-    if (kpxc.combinations.length > 0 && kpxc.settings.autoCompleteUsernames) {
+    if (kpxc.combinations.length > 0) {
         if (await kpxcFill.fillFromCombination(passOnly)) {
             // Combination found and filled
             return;
@@ -64,7 +64,7 @@ kpxcFill.fillFromCombination = async function(passOnly) {
     // Set focus to the input field
     field.focus();
 
-    if (kpxc.credentials.length > 1) {
+    if (kpxc.credentials.length > 1 && kpxc.settings.autoCompleteUsernames) {
         // More than one credential -> show autocomplete list
         kpxcUserAutocomplete.showList(field);
     } else {


keepassxc-browser/content/fill.js
```diff
@@ -64,7 +64,7 @@ kpxcFill.fillFromCombination = async function(passOnly) {
     // Set focus to the input field
     field.focus();
 
-    if (kpxc.credentials.length > 1) {
+    if (kpxc.credentials.length > 1 && kpxc.settings.autoCompleteUsernames) {
         // More than one credential -> show autocomplete list
         kpxcUserAutocomplete.showList(field);
     } else {
```
droidmonkey, line 70: I feel like this changes the else condition. I am not familiar with the autoCompleteUsernames option

droidmonkey:
I may be getting old but that code is so hard to read. The switching between combinations and credentials...

keepassxc-browser/content/fill.js
varjolintu: The setting is named a bit poorly, but when it's `true` the Autocomplete Menu is shown when needed. Before this change the filling from combination was prevented if the option was disabled, but actually that needs to be checked later when we are actually trying to show the list.

If the option is disabled and no list is shown, the first credential will be filled.

droidmonkey: Would it be a big change to change the name of the setting to align with its use?

varjolintu: If we change the name, we need to make a function that detects the old setting and transfers the value to the new one. Otherwise old configs are gonna reset the option. So, it's possible.

droidmonkey: Save that for protocol v2

droidmonkey (approved):

PR #1983, opened 2023-09-09
Cosmetic fixes

stefansundin:
Hello. I have a lot of minor fixes, mostly cosmetic. I took great care to split them up into small commits and I put some of my reasoning in the commit messages. Feel free to squash when merging though!

I think most of it should be fairly uncontroversial, but feel free to let me know if you want anything changed.

I want to work on some more changes too, but I'll wait for this PR to be merged. :)

---

Making sure all buttons have whitespace between the icon and the text:

Screenshot1
Screenshot2
<br>
Screenshot3
Screenshot4

Some of the translations have spaces in the front of the text string to account for this, but it is very inconsistent and I don't think it is something that translators should have to worry about. You might want to update the strings after this change, perhaps.

Run this to find the strings that have a space in the beginning:

```
grep -r ': " ' keepassxc-browser/_locales
```

Fixed dropdown carets not showing:

Screenshot5

Screenshot6

Corrected bootstrap styling:

Screenshot

Screenshot


keepassxc-browser/popups/popup_functions.js
```diff
@@ -48,7 +48,7 @@ async function initColorTheme() {
 }
 
 async function getLoginData() {
-    const tab = getCurrentTab();
+    const tab = await getCurrentTab();
     if (!tab) {
         return [];
     }
```
stefansundin, line 54: I don't know if this mistake had any adverse impact. It would just mean that the `if (!tab) {` check never did anything, as `tab` is always a `Promise` here.

If the source code was converted to TypeScript then these mistakes could be detected with the proper eslint packages. Has that ever been considered?

droidmonkey:
Very nice!

keepassxc-browser/popups/popup_functions.js
varjolintu: > f the source code was converted to TypeScript then these mistakes could be detected with the proper eslint packages. Has that ever been considered?

At one point, yes. There's just two small problems with this:
1) I have never used TypeScript.
2) I like to keep the code in vanilla JS so it's easy to debug on-the-fly (by anyone) and the code size stays small.

varjolintu:
Thank you! :)

varjolintu (approved):

PR #1985, opened 2023-09-16
Use active element on fillFromCombination()

varjolintu:
Find the active element from the combinations list instead of just returning the first value that has username or password fields. There can be multiple combinations detected from a page.

Can be tested using https://banking.postbank.de/#/login

### DIFF
--- a/keepassxc-browser/content/fill.js
+++ b/keepassxc-browser/content/fill.js
@@ -30,26 +30,26 @@ kpxcFill.fillInFromActiveElement = async function(passOnly = false) {
         return;
     }
 
+    const elem = document.activeElement;
     if (kpxc.combinations.length > 0) {
-        if (await kpxcFill.fillFromCombination(passOnly)) {
+        if (await kpxcFill.fillFromCombination(elem, passOnly)) {
             // Combination found and filled
             return;
         }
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/fill.js
```diff
@@ -30,26 +30,26 @@ kpxcFill.fillInFromActiveElement = async function(passOnly = false) {
...
 // Fill from combination, if found
-kpxcFill.fillFromCombination = async function(passOnly) {
+kpxcFill.fillFromCombination = async function(elem, passOnly) {
     const combination = passOnly
-        ? kpxc.combinations.find(c => c.password)
-        : kpxc.combinations.find(c => c.username);
+        ? kpxc.combinations.find(c => c.password === elem)
+        : kpxc.combinations.find(c => c.username === elem);
     if (!combination) {
         logDebug('Error: No combination found.');
```
droidmonkey, line 54: Just a comment... "combination" is very ambiguous here for the log reader. Can we change this to "No username/password field combination found."?

droidmonkey (approved):

varjolintu:
Found one more bug concerning this. Making a second fix.

PR #1987, opened 2023-09-16
Upgrade bootstrap for improved dark mode support

stefansundin:
There's a new color mode feature in Bootstrap v5.3. The form elements on the options page definitely looks better in dark mode now. https://getbootstrap.com/docs/5.3/customize/color-modes/

I got the new bootstrap files from https://getbootstrap.com/docs/5.3/getting-started/download/:
- https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/dist/css/bootstrap.min.css
- https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/dist/js/bootstrap.min.js

Then I removed the source map lines at the bottom of each file as that appears to have been done before.


droidmonkey (approved):

varjolintu (changes requested):
When system theme is selected, the text colors are incorrect and table coloring is half-dark/light:
Screenshot 2023-09-17 at 7 37 07
Screenshot 2023-09-17 at 7 38 30

stefansundin:
Thanks for finding those bugs, I think I've addressed them now.

The `data-bs-theme` attribute is now always set to `light` or `dark`, and never `system`. Some of this code was checking for an `undefined` value but I wasn't able to get it to actually be `undefined` (even when manipulating the `browser.storage.local` value directly), so I don't think we need to account for it anymore.

I made the options page react to when the system theme changes too. We could probably make this happen in the popup/content elements too.

I also made some other changes to the options page (renaming the method, etc).

varjolintu:
> Some of this code was checking for an `undefined` value but I wasn't able to get it to actually be `undefined` (even when manipulating the `browser.storage.local` value directly), so I don't think we need to account for it anymore.

I think the `undefined` was addressed before because the whole setting didn't exist in the beginning, and there was some issue with it even when set via `page.js`. Cannot remember the exact case anymore. But true, it's not needed anymore.

keepassxc-browser/content/ui.js
```diff
@@ -324,16 +324,11 @@ const DOMRectToArray = function(domRect) {
...
-
-    if (colorTheme === undefined) {
-        elem.removeAttribute('data-color-theme');
-    } else if (colorTheme === 'system') {
-        const theme = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
-        elem.setAttribute('data-color-theme', theme);
-    } else {
-        elem.setAttribute('data-color-theme', colorTheme);
+    let theme = kpxc.settings['colorTheme'];
+    if (theme === 'system') {
```
varjolintu, line 328: This same if sentence repeats in many files. It could be changed to a function inside `ui.js`, for example:
```
setTheme = function(theme, elem) {
    if (theme === 'system') {
        theme = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
    }
    elem.setAttribute('data-bs-theme', theme);
}
```

Then calling it like `setTheme(theme, document.documentElement);` etc.

stefansundin: Hmm.. It seems like `ui.js` is only usable for content script functions? I tried it but I didn't find that it helped so much since I couldn't use it elsewhere.

I guess it is probably possible to include `ui.js` everywhere, but I haven't looked into it that much. I can try it later.

stefansundin:
Found and fixed one bug related to "first run" (before localStorage is populated).

keepassxc-browser/content/ui.js
varjolintu: How about creating a separate `theme.js`?

varjolintu:
Updated the following colors:
- Table row backgrounds, table font colors (fonts were too bright, colors were different when compared to current version)
- Popup error message color (from pink to red)
- Popup horizontal line color to match ones in the Settings page

varjolintu (approved):

droidmonkey:
I'm good with this

PR #2054, opened 2023-12-22
Small fixes to the options page

stefansundin:
Hello, I'm back with some small fixes that I have accumulated over a few months.

It was possible to get the "empty message row" to show up when having things in the table, for example in the list of connected databases. If you add your first connection from the options page then you'll get this:

Screenshot 2023-12-22 at 11 06 43

So I rewrote the logic that hides the row in pure CSS, which makes the code much cleaner and fixes this bug.

---

I removed `table-bordered` from the tables since I think it gives the pages a cleaner look. Here's the before/after:

Screenshot 2023-12-22 at 12 06 52

Screenshot 2023-12-22 at 12 06 59

Hopefully you like this look better too. 😄

---

Also fixed some other things, please see the commit messages for more details. Let me know if you want me to change anything.

### DIFF
--- a/keepassxc-browser/options/options.html
+++ b/keepassxc-browser/options/options.html
@@ -596,7 +596,7 @@
                             <i class="fa fa-remove" aria-hidden="true"></i>
                             <span data-i18n="optionsButtonCancel"></span>
                           </button>
-                          <button class="btn btn-sm yes btn-success">
+                          <button class="btn btn-sm yes btn-danger">
                             <i class="fa fa-check" aria-hidden="true"></i>
                             <span data-i18n="optionsButtonRemoveNow"></span>
                           </button>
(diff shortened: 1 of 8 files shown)


varjolintu:
Thank you! This is an old bug I never had time to fix :)

keepassxc-browser/options/options.html
```diff
@@ -543,7 +543,7 @@ <h2 class="pb-3 mt-0" data-i18n="optionsConnectedDatabasesTab"></h2>
               <div class="card my-4 shadow">
                 <div class="card-body">
                   <div class="table-responsive">
-                    <table class="table table-sm table-striped table-bordered">
+                    <table class="table table-sm table-striped">
```
varjolintu, line 546: Why is `table-bordered` removed from every table? After the change there are no borders in left or right.

stefansundin: I think it gives the pages a cleaner look, but if you feel strongly about it then I can revert this change.

varjolintu: I understand your point. Maybe bigger design changes should be done in a separate PR anyway.

varjolintu (approved):

droidmonkey (approved):

droidmonkey:
Personally don't like tables without side borders, they tend to just melt into the background which isn't the intent of a table. I think that table (or all tables) could use a better, darker, border color in dark mode instead.

PR #2079, opened 2024-01-14
Add support for TLD check and allowing Cross-Origin iframes option

### LINKED ISSUE #2076: Credential request ignored from another domain
## Expected Behavior
The password should be automatically filled in on this webpage: https://account.aliyun.com/login/login.htm

## Current Behavior
After I updated KeePassXC-Browser to version 1.8.11, it no longer automatically fills in on this website, whereas it worked before the update

## Possible Solution
The domain name of this website is https://account.aliyun.com/, but the login domain is https://passport.aliyun.com/. Previously, I added passport.aliyun.com to the additional URLs in KeePassXC, and it was functioning
After the update, the same website reports the following error in the debugging information: 'KeePassXC-Browser - Error: Credential request ignored from another domain: passport.aliyun.com
Is there a way to fix this issue or manually set the trusted domains for the current domain to not be ignored?

1. Add https://passport.aliyun.com/ to the additional URLs in KeePassXC
2. Access https://account.aliyun.com/login/login.htm
3. It can be noticed that the KeePassXC icon is grayed out and the password is not found
4. With the debug log opened, it can be observed in the console that there is an error reported as 'KeePassXC-Browser - Error: Credential request ignored from another domain: passport.aliyun.com.

## Debug info
KeePassXC - 2.7.6
KeePassXC-Browser - 1.8.11
Operating system: Mac
Browser: Chrome

varjolintu:
Adds support for checking Top-Level-Domains from hostnames, and returning a base domain. The implementation is almost similar to KeePassXC's https://github.com/keepassxreboot/keepassxc/pull/9935 that is using a cookie API for TLD's (there's no direct API for it).

If a Cross-Origin iframe is detected, the popup will show a new option:
Screenshot 2024-01-14 at 10 10 45
Pressing the button will add the tab/window URL to Site Preferences and enables the new "Allow Cross-Origin iframes" option.

Content script's `isIframeAllowed()` now performs the following:
- Checks if Cross-Origin iframes are allowed by Site Preferences.
- Checks the base domain of window/iframe using TLD list.
- Allows the iframe automatically if it's under the same base domain.
- Disallows iframes that do not match with the tab's/window's URL.

Unit tests are disabled for now, because it's unclear how the background script can be tested with Cookies API enabled. I ran the same tests directly in the background script when making the feature, and all of those passed.

Fixes #2076.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ KeePassXC-Browser extension requests the following permissions:
 | ----- | ----- |
 | `activeTab`               | To get URL of the current tab |
 | `contextMenus`            | To show context menu items |
+| `cookies`                 | To access browser's internal Public Suffix List |
 | `clipboardWrite`          | Allows password to be copied from password generator to clipboard |
 | `nativeMessaging`         | Allows communication with KeePassXC application |
 | `notifications`           | To show browser notifications |
@@ -42,7 +43,7 @@ KeePassXC-Browser extension requests the following permissions:
 
 ## Protocol
 
-The details about the messaging protocol used with the browser extension and KeePassXC can be found [here](keepassxc-protocol.md).
+Check [keepassxc-protocol](keepassxc-protocol.md) for the details about the messaging protocol used between the browser extension and KeePassXC.
 
 ## Translations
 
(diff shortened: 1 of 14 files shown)


varjolintu:
There are some problems with the automatic subdomain detection with Firefox. Trying to fix it.

Felix-phi:
The same problem. How can I get the fixed version?

varjolintu:
@Felix-phi What do you mean?

droidmonkey (approved):
Works perfectly on the test/issue sites

PR #2107, opened 2024-02-01
Do not allow filling password (only) to plain text fields

### LINKED ISSUE #2102: Fill Password Only keyboard shortcut fills in plaintext fields
Sometimes I accidentally press the shortcut for filling in the password instead of TOTP, and the password ends up in clear in the field. I guess the shortcuts could be adjusted, but the defaults are quite prone to this mistake, at least with a QWERTY layout.

## Expected Behavior

It would be nice if KeepassXC-Browser didn't autofill the password in a plaintext field.

## Current Behavior

It does.

## Possible Solution

Don't autofill the password in plaintext inputs unless it's been setup as a custom field.  
Or maybe ask for confirmation, such as requiring the user to press the shortcut a second time (seems like more work though).

## Debug info
None of these really matter here, it's easily reproducible with any existing version of the extension.

KeePassXC - 2.7.6
KeePassXC-Browser - 1.8.11
Browser: Mozilla Firefox 120.0

---

P.S. thanks for all the work, I use this extension daily and it's generally great!

varjolintu:
When filling password from a keyboard shortcut or from a context menu, plain text fields should not be filled. When Custom Login Fields are used and the input element is specified as a password field, then the fill is allowed.

Fixes #2102.

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -347,6 +347,10 @@
         "message": "Error: Unable to find a password field.",
         "description": "Message shown when no password fields are found."
     },
+    "fieldsPasswordFillNotAccepted": {
+        "message": "Filling password to a plain text field is prevented.",
+        "description": "Message shown when password fill to a plain text field is prevented."
+    },
     "rememberNothingChanged": {
         "message": "Error: Could not detect changed credentials.",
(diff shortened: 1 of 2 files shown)


droidmonkey (approved):

ThinkChaos:
Thank you both!  
Looking forward to the next release :)

PR #2120, opened 2024-02-16
Fix calculation of icon offset

tinyoverflow:
Hello,

this PR fixes the current calculation of the icon offset relative to input fields.
Especially with large input fields it can happen, that the icon is not placed appropiately. Please refer to my before/after screenshots for a comparison.

**Before:**
1_before

**After:**
1_after

**Before:**
2_before

**After:**
2_after

### DIFF
--- a/keepassxc-browser/content/ui.js
+++ b/keepassxc-browser/content/ui.js
@@ -126,7 +126,8 @@ kpxcUI.updateIconPosition = function(iconClass) {
 };
 
 kpxcUI.calculateIconOffset = function(field, size) {
-    const offset = Math.floor((field.offsetHeight - size) / 3);
+    const offset = Math.floor((field.offsetHeight / 2) - (size / 2) - 1);
+    console.log(field.offsetHeight, size, offset);
     return (offset < 0) ? 0 : offset;
 };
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/ui.js
```diff
@@ -126,7 +126,8 @@ kpxcUI.updateIconPosition = function(iconClass) {
 };
 
 kpxcUI.calculateIconOffset = function(field, size) {
-    const offset = Math.floor((field.offsetHeight - size) / 3);
+    const offset = Math.floor((field.offsetHeight / 2) - (size / 2) - 1);
+    console.log(field.offsetHeight, size, offset);
```
droidmonkey, line 130: No need to write this to the log, I assume you left that in here from debugging.

droidmonkey (changes requested):

keepassxc-browser/content/ui.js
tinyoverflow: Totally missed that. Fixed :)

droidmonkey (approved):

varjolintu (approved):

PR #2121, opened 2024-02-19
Passkey improvements

varjolintu:
Makes some improvements to the Passkeys implementation:
- Respects the `sameOriginWithAncestors`. If the origin does not match, the request are ignored.
- Only basic checks are done here. Everything else is moved to KeePassXC side to ensure it's safer to use from 3rd party clients.
- Timeouts are already set here instead of KeePassXC.
- Adds a lifetime timer (with a greater timeout than the actual request). When credential creation is requested and some unexpected error happens or credential is excluded, the timer will run out. This prevents possible requests from a fraudulent site that tries to figure out which credentials are registered.
- Throw specific exceptions for errors.

Corresponding KeePassXC side PR: https://github.com/keepassxreboot/keepassxc/pull/10318


droidmonkey:
Need to also add an https/certificate validity check if that is possible. See https://github.com/keepassxreboot/keepassxc-browser/issues/2122

droidmonkey (approved):

varjolintu:
I cannot get the Microsoft Passkeys creation to work with Firefox (macOS). It always popups the OS' own Passkeys prompt instead. I thought it might be related to this bug https://bugzilla.mozilla.org/show_bug.cgi?id=1878598 but after setting that option to false, I get a different prompt from Firefox.

I'm still trying to disable few other options too and see if it helps.

Ortham (changes requested):
Hi, I'm still going through the corresponding KeePassXC changes (I've just gotten through the create credentials flow, still have get to go), but I spotted some issues and some things I'm not sure about in this PR.

keepassxc-browser/content/passkeys.js
```diff
@@ -31,13 +94,20 @@ const postMessageToExtension = function(request) {
                 return null;
             }
 
-            const response = await postMessageToExtension({ action: 'passkeys_create', publicKey: options.publicKey });
+            const sameOriginWithAncestors = isSameOriginWithAncestors();
```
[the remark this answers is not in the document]
varjolintu: Each iframe has its own content scripts, yes.

keepassxc-browser/content/passkeys-utils.js
```diff
@@ -101,21 +99,20 @@ kpxcPasskeysUtils.buildCredentialCreationOptions = function(pkOptions) {
...
 
 // Create a new object with base64 strings for KeePassXC
-kpxcPasskeysUtils.buildCredentialRequestOptions = function(pkOptions) {
+kpxcPasskeysUtils.buildCredentialRequestOptions = function(pkOptions, sameOriginWithAncestors) {
     try {
-        checkErrors(pkOptions);
+        checkErrors(pkOptions, sameOriginWithAncestors);
 
         if (!pkOptions.rpId) {
             pkOptions.rpId = window.location.hostname;
```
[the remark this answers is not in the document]
varjolintu: Some pages (like https://webauthn.me/debugger) are leaving this `undefined`, but I can always add that manually back for testing purposes.

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -806,35 +810,63 @@ kpxc.enablePasskeys = function() {
...
+            clearTimeout(lifetimeTimer);
+        }
+    };
 
-            const ret = await sendMessage('passkeys_get', [ publicKey, window.location.origin ]);
-            if (ret) {
-                if (ret.response && ret.response.errorCode) {
-                    const errorMessage = await sendMessage('get_error_message', ret.response.errorCode);
+    const letTimerRunOut = function(errorCode) {
+        return errorCode === PASSKEYS_WAIT_FOR_LIFETIMER || errorCode === PASSKEYS_CREDENTIAL_IS_EXCLUDED;
```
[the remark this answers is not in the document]
varjolintu: https://www.w3.org/TR/2019/REC-webauthn-1-20190304/#sec-make-credential-privacy <- this is what I was thinking when not returning an excluded credential early. If it's possible to sniff quickly if there's a credential existing for a site, wouldn't it be better to let the extension timeout? The user is still shown a notification about an excluded credential, but the Relying Party doesn't know it.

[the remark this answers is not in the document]
Ortham: Yeah, I think I misunderstood when the `PASSKEYS_CREDENTIAL_IS_EXCLUDED` code was used when I wrote that comment, I thought it was used when the user had consented to reveal the credential's existence - since you're never asking for consent in the excluded case, then waiting for the timeout is the correct approach.

keepassxc-browser/content/passkeys.js
```diff
@@ -31,13 +94,20 @@ const postMessageToExtension = function(request) {
...
+                action: 'passkeys_create',
+                publicKey: options.publicKey,
+                sameOriginWithAncestors: sameOriginWithAncestors,
+            });
             if (!response.publicKey) {
+                handleError(response?.errorCode, response?.errorMessage);
                 return response.fallback ? originalCredentials.create(options) : null;
             }
 
             response.publicKey.getClientExtensionResults = () => {};
```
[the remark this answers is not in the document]
varjolintu: I think this can be done later if it's really necessary.

keepassxc-browser/content/passkeys-utils.js
```diff
@@ -16,65 +20,61 @@ const arrayBufferToBase64 = function(buf) {
...
+const checkErrors = function(pkOptions, sameOriginWithAncestors) {
+    if (!pkOptions) {
+        throw new Error('No publicKey configuration options were provided');
+    }
+
+    if (pkOptions.signal && pkOptions.signal.aborted) {
+        throw new DOMException('Abort signalled', DOMException.AbortError);
+    }
+
+    if (!sameOriginWithAncestors || (sameOriginWithAncestors && sameOriginWithAncestors === false)) {
```
Ortham, line 32: `sameOriginWithAncestors && sameOriginWithAncestors === false` is always `false` so I think `if (!sameOriginWithAncestors)` is enough.

Ortham (commented):
I just spotted one redundant if condition in your changes.

keepassxc-browser/content/passkeys.js
```diff
@@ -85,6 +85,38 @@ const handleError = function(errorCode, errorMessage) {
...
+        clientExtensionResults: {
+            value: () => {}
+        },
+        getClientExtensionResults: {
+            value: () => {}
+        },
+        id: {
+            value: publicKeyOptions.id
+        },
+        isConditionalMediationAvailable: {
```
Ortham, line 103: I don't think this is right as it's supposed to be a static property, i.e. it should be on the prototype rather than the object instance, as it's used like `PublicKeyCredential.isConditionalMediationAvailable()`, not `publicKey.isConditionalMediationAvailable()` (where `publicKey` is an instance of `PublicKeyCredential`). I think the `Object.defineProperty()` you've already got for this property further down is enough.

keepassxc-browser/content/passkeys.js
```diff
@@ -146,9 +175,6 @@ const handleError = function(errorCode, errorMessage) {
         Object.defineProperty(window.PublicKeyCredential, 'isConditionalMediationAvailable', {
             value: isConditionalMediationAvailable,
         });
-        Object.defineProperties(window.PublicKeyCredential, 'isUserVerifyingPlatformAuthenticatorAvailable', {
```
Ortham: Was removing this intentional?

Ortham (changes requested):

keepassxc-browser/content/passkeys.js
varjolintu: I was having constant errors in the console that overriding this is not possible.

keepassxc-browser/content/passkeys.js
varjolintu: Ah, got it. I was about to remove the `Object.defineProperty()` one though, but maybe it's better to keep this there.

keepassxc-browser/content/passkeys.js
Ortham: What browser was that with? I don't recall seeing that on Firefox.

varjolintu: There was a typo. Fixed.

droidmonkey (approved):
Adding a final approval, works great

PR #2125, opened 2024-02-24
Show number of credentials in the extension icon

varjolintu:
Mostly a suggestion, but the red markers in extension toolbar icon should be restricted to purely error situations. When credentials are found for the site a badge text could indicate how many credentials are found. Some users have reported that the questionmark with the red background seems to indicate an error, and it's not obvious what is currently happening.

Questionmark icons are removed from the sources.

Before:
Screenshot 2024-02-24 at 13 20 15
After:
Screenshot 2024-02-24 at 13 20 22


droidmonkey:
Looks good, I'll await manifest 3 before review

varjolintu:
Made the changes needed by Manifest V3.

droidmonkey (approved):

PR #2152, opened 2024-03-19
Fix monochrome icon in Chrome when `colorTheme == 'system'` (MV3 fix)

stefansundin:
Since the MV3 upgrade, the background script in Chrome does not have a `window`. Attempting to use the monochrome icon when the theme is `system` results in an error.

Screenshot 2024-03-18 at 20 04 00

This uses the offscreen API to work around the issue. Another option is to set the "effective" theme when the options are saved. This would be a poor experience if the user changes the theme (or if their computer is set to automatically change it based on the time of day).

This doesn't close the offscreen document, which I _think_ is ok. I think it will automatically close on its own. I think this is called a lot so it is probably good to keep it around. 🤷

### DIFF
--- a/keepassxc-browser/offscreen/offscreen.js
+++ b/keepassxc-browser/offscreen/offscreen.js
@@ -1,0 +1,9 @@
+chrome.runtime.onMessage.addListener(({ target, action }, sender, sendResponse) => {
+    if (target !== 'offscreen') {
+        return;
+    }
+    if (action === 'retrieve_color_scheme') {
+        const style = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+        sendResponse(style);
+    }
+});
--- a/.eslintrc
+++ b/.eslintrc
@@ -159,6 +159,7 @@
         "Pixels": true,
         "PREDEFINED_SITELIST": true,
         "RED_BUTTON": true,
+        "retrieveColorScheme": true,
         "sendMessage": true,
         "showNotification": true,
         "siteMatch": true,
--- a/dist/manifest_chromium.json
+++ b/dist/manifest_chromium.json
@@ -143,6 +143,7 @@
         "cookies",
         "nativeMessaging",
         "notifications",
+        "offscreen",
         "storage",
         "tabs",
         "webNavigation",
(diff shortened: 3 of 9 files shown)


keepassxc-browser/offscreen/offscreen.js
```diff
@@ -0,0 +1,9 @@
+chrome.runtime.onMessage.addListener(({ target, action }, sender, sendResponse) => {
```
stefansundin, line 1: I wasn't able to get this working with the browser-polyfill. Maybe I did it wrong or it didn't load in time, or something. I didn't spend a lot of time on it.

varjolintu: https://github.com/mozilla/webextension-polyfill/issues/329 This issue thread didn't help either. I'll take a look when I the time.

stefansundin: I don't know if the polyfill is that useful anymore since Chrome supports promises for most (if not all) APIs, and Firefox has always supported `chrome` instead of `browser`. 🤷 

Unless there's another reason for the polyfill that I have missed?

varjolintu: The polyfill was used mostly because of the differences between promises and callbacks. I could investigate if we can get rid of the whole polyfill library for good :)

varjolintu: I tested this a little bit and it seems to work fine. What exactly doesn't work here? If you expect async response for the `addListener()` you can add `return true;` after the `sendResponse()`.

`manifest_json` and `manifest_firefox.json` needs to include `background/offscreen.js`, otherwise `retrieveColorScheme()` will fail.

stefansundin: Sorry for the confusion, it definitely works. I just wanted to point out that it is an exception to what is done elsewhere in the codebase.

If I remember correctly this is perfectly good to merge.

And yes, it might be worth removing the polyfill down the line, but that is not a requirement for this change.

stefansundin: I just read your comment more closely and fixed the manifest files.

varjolintu (approved):

varjolintu:
It would be preferred if it could be possible to just send a request to content script and return the answer, but after looking at the issue myself this seems to be the only proper solution.

PR #2154, opened 2024-03-23
Fix filling the next field from password generator

varjolintu:
Caused by https://github.com/keepassxreboot/keepassxc-browser/pull/1966. The `kpxcPasswordDialog` object no longer exists.

### DIFF
--- a/keepassxc-browser/content/pwgen.js
+++ b/keepassxc-browser/content/pwgen.js
@@ -122,8 +122,12 @@ kpxcPasswordGenerator.fill = function(elem, password) {
 
     if (elem.getAttribute('maxlength')) {
         if (password.length > elem.getAttribute('maxlength')) {
-            const message = tr('passwordGeneratorErrorTooLong') + '\r\n'
-                            + tr('passwordGeneratorErrorTooLongCut') + '\r\n' + tr('passwordGeneratorErrorTooLongRemember');
+            const message =
+                tr('passwordGeneratorErrorTooLong') +
+                '\r\n' +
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/pwgen.js
```diff
@@ -137,14 +141,13 @@ kpxcPasswordGenerator.fill = function(elem, password) {
...
         const index = kpxc.inputs.indexOf(elem);
-        const nextField = kpxc.inputs[index + 1];
-
-        kpxcPasswordDialog.nextField =
-            nextField && nextField.getLowerCaseAttribute('type') === 'password' ? nextField : undefined;
-        if (kpxcPasswordDialog.nextField) {
-            kpxcPasswordDialog.nextField.value = password;
-            kpxcPasswordDialog.nextField.dispatchEvent(new Event('input', { bubbles: true }));
-            kpxcPasswordDialog.nextField.dispatchEvent(new Event('change', { bubbles: true }));
+        const next = kpxc.inputs[index + 1];
```
droidmonkey, line 144: Need to check the length of the inputs first, you just check for >0, or does js just return null and not complain about out of bounds?

varjolintu: Right. Let's add some extra checks here just in case.

varjolintu: Actually `.at()` and [] operator both returns `undefined` if there's no value.

droidmonkey (approved):

PR #2161, opened 2024-03-30
Parse algorithm type as Number(). Fix required Firefox version.

varjolintu:
Fixes the Firefox manifest having an incorrect minimum version.
Also adds some compatibility for some older WebAuthn/passkeys servers that are returning COSEAlgorithmIdentifiers as strings instead of integers.


keepassxc-browser/content/passkeys-utils.js
piyushgarg, line 1: I have verified on Version 122.0.6261.128 (Official Build, ungoogled-chromium) Arch Linux (64-bit).

droidmonkey (approved):

PR #2166, opened 2024-04-02
Handle iframe URLs correctly

varjolintu:
When comparing URL's with the new iframe support, things are not working as planned.
Fixes:
- Do not use `toLowerCase()`. The paths can contain uppercase letters, and we no longer rely just on domains.
- Use `trimURL()` to remove search part of the URL so a proper comparisons can be made.

Fixes the issue mentioned in https://github.com/keepassxreboot/keepassxc-browser/issues/1358#issuecomment-2031388026.

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -340,7 +340,7 @@ page.setAllowIframes = async function(tab, args = []) {
     const [ allowIframes, site ] = args;
 
     // Only set when main windows' URL is used
-    if (tab?.url === site) {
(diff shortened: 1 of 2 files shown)


apollo13:
When loading your branch as temporary addon I get:
```
SecurityError: Permission denied to get property "href" on cross-origin object keepassxc-browser.js:36
```
when trying to add the URL via the addon popup.

apollo13:
(Aside from that error it seems to work fine though)

varjolintu:
I'll check that error too. Thanks.

droidmonkey (approved):

Issue #2184, opened 2024-04-30
Suggestions for Improvement in Detecting Hidden Form Elements Methods


Felix026x:
Dear KeePassXC Developer Team,

We have recently delved into the KeePassXC AutoType (or Autofill functionality) to detect the visibility of web form elements. Through meticulous debugging and analysis, we have identified potential areas for improvement in the current detection method to provide a more robust **security** guarantee. We believe that enhancing this detection mechanism can further bolster the security and accuracy of KeePassXC when dealing with hidden form elements.

We have inspected the `isVisible` function in the source code `https://github.com/keepassxreboot/keepassxc-browser/blob/develop/keepassxc-browser/content/fields.js`, and observed the existing implementation. It conducts its detection process through the following steps: it starts by checking the size and position of the element, then verifies the CSS visibility of the element, and finally the opacity of its parent elements.

We now summarize the results and report the following scenarios: the password field is hidden, and KeePassXC should have detected its invisibility and should not autofill the password into it. After inspecting and debugging the source code, we find several potential issues.

1. Lack of Detection for Element Overlay: The current method does not account for cases where elements might be covered by other page elements, making them invisible to users. For instance, a password field in a form could be obscured by another element or moved outside the parent element with a modification of the CSS `overflow: hidden` property to achieve hiding. (**1-Overlay**)
2. Limitations in Parent Element Opacity Detection: The practice of using the `f.style` method to detect `opacity` presents a limitation, as it only captures inline CSS property values directly defined within the element (e.g.,`<input id="password" style="opacity: 0;">`).  This means that `opacity` set through nested CSS files or `<style>` tags in the HTML `<header>` tags for a parent element cannot be effectively detected. (**2-Parent-Opacity**)
3. Insufficiency in Element's Own Visibility Detection: The current detection mechanism fails to cover CSS properties such as `clip` and `clip-path`, which may lead to certain hiding techniques not being promptly identified. (**3-Clip-Path-Self**)
4. Incomplete Detection Regarding Parent Elements: The present detection lacks a necessary mechanism to check for CSS properties that may affect the visibility of child elements without alerting their size (e.g., the parent element is set CSS property with `content-visibility` or `clip-path`). This might result in elements that are capable of being filled but are invisible being missed. (**4-Content-Visibility-Parent**)

We have provided four examples on a demo website to demonstrate the reproduction process.

- 1-Overlay: https://felix026x.github.io/overlay.html 
[...]

PR #2186, opened 2024-04-30
[FIX # 2184] Add clip, clip-path check; Add parent visibility check

Felix026x:
To solve Issue #2184, we modify the original method for visibility check, and consider the hidden techniques using `clip` and `clip-path`. We also fix the `style` issue, where the original code utilizes the style of the inline element while overlooking the nested CSS styles. However, for performance overhead, we do not implement the detection for hidden fields caused by overlays.

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -391,17 +391,55 @@ kpxcFields.isVisible = function(elem) {
     if (elemStyle.visibility && (elemStyle.visibility === 'hidden' || elemStyle.visibility === 'collapse')
         || (opacity < MIN_OPACITY || opacity > MAX_OPACITY)
         || parseInt(elemStyle.width, 10) <= MIN_INPUT_FIELD_WIDTH_PX
-        || parseInt(elemStyle.height, 10) <= MIN_INPUT_FIELD_WIDTH_PX) {
+        || parseInt(elemStyle.height, 10) <= MIN_INPUT_FIELD_WIDTH_PX
+        // [FIX #2184] added by felix
+        || kpxcFields.isElementClipped(elem)
+    ) {
         return false;
     }
 
-    // Check for parent opacity
-    if (kpxcFields.traverseParents(elem, f => f.style.opacity === '0')) {
+    // [FIX #2184] added by felix
+    // Check for parent visibility
+    if (
+        kpxcFields.traverseParents(
+            elem,
+            (f) => {
+                const fStyle = getComputedStyle(f, null);
+                const isVisible = fStyle.visibility !== 'hidden' && fStyle.visibility !== 'collapse';
+                const isOpaqueEnough = Number(fStyle.opacity) >= MIN_OPACITY && Number(fStyle.opacity) <= MAX_OPACITY;
+                const hasValidSize = parseInt(fStyle.width, 10) > MIN_INPUT_FIELD_WIDTH_PX && parseInt(fStyle.height, 10) > MIN_INPUT_FIELD_WIDTH_PX;
+                const isNotClipped = !kpxcFields.isElementClipped(f);
+                return !(isVisible && isOpaqueEnough && hasValidSize && isNotClipped);
+            }
+        )
+    ) {
         return false;
     }
 
     return true;
 };
+
+// [FIX #2184] added by felix
+// TODO: This set is limited, but can cover many cases
+kpxcFields.isElementClipped = function(elem) {
+    var clipPathStyles = new Set([
+        "inset(50%)",
+        "inset(100%)",
+        "inset(100% 100% 0% 0%)",
+        "circle(0)",
+        "circle(0px)",
+        "circle(0px at 50% 50%)",
+        "polygon(0 0, 0 0, 0 0, 0 0)",
+        "polygon(0px 0px, 0px 0px, 0px 0px, 0px 0px)"
+    ]);
+    var clipStyles = new Set([
+        "rect(0px, 0px, 0px, 0px)"
+    ]);
+    const eStyle = getComputedStyle(elem, 'null');
+    const clipPath = eStyle.clipPath;
+    const clip = eStyle.clip;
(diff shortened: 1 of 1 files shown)


droidmonkey (changes requested):

keepassxc-browser/content/fields.js
```diff
@@ -391,18 +391,56 @@ kpxcFields.isVisible = function(elem) {
     if (elemStyle.visibility && (elemStyle.visibility === 'hidden' || elemStyle.visibility === 'collapse')
         || (opacity < MIN_OPACITY || opacity > MAX_OPACITY)
         || parseInt(elemStyle.width, 10) <= MIN_INPUT_FIELD_WIDTH_PX
-        || parseInt(elemStyle.height, 10) <= MIN_INPUT_FIELD_WIDTH_PX) {
+        || parseInt(elemStyle.height, 10) <= MIN_INPUT_FIELD_WIDTH_PX
+        // [FIX #2184] added by felix
```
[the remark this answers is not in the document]
Felix026x: Thanks for your response, I will delete comments like this.

keepassxc-browser/content/fields.js
```diff
@@ -391,18 +391,56 @@ kpxcFields.isVisible = function(elem) {
...
         return false;
     }
 
     return true;
 };
 
+// [FIX #2184] added by felix
+// TODO: This set is limited, but can cover many cases
+kpxcFields.isElementClipped = function(elem) {
+    var clipPathStyles = new Set([
```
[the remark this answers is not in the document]
Felix026x: Thanks for your insightful response. I have understand the complexity and limitations of our method for detecting the `clip` and `clip-path` property. We use a simple method for detecting the properties. If this is improper, I will consider to delete this function. 

Best regards,
Felix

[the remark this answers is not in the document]
droidmonkey: You should think of a regex or similar pattern if you must do text comparisons. Ideally you would resolve this based on actual properties and not text matches.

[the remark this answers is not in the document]
Felix026x: Thanks for your kindly response, I will think deeply to solve this issue. 

Best regards,
Felix

keepassxc-browser/content/fields.js
```diff
@@ -391,18 +391,64 @@ kpxcFields.isVisible = function(elem) {
...
+kpxcFields.isElementOpaqueEnough = function(elem, fStyle) {
+    if (fStyle === null) {
+        const fStyle = getComputedStyle(f, null);
+    }
+    return Number(fStyle.opacity) >= MIN_OPACITY && Number(fStyle.opacity) <= MAX_OPACITY;
+}
+
+kpxcFields.isElementHasValidSize = function(elem, fStyle) {
+    if (fStyle === null) {
+        const fStyle = getComputedStyle(f, null);
```
droidmonkey, line 431: `f` is not a variable of this function. Also, you shouldn't try to compute the style in each of these functions, the point of passing the fstyle parameter is to avoid this. If fstyle is null, just bale out.

droidmonkey (changes requested):

keepassxc-browser/content/fields.js
Felix026x: Sorry for this mistake, I will fix it.

varjolintu:
Are there any common known sites that are affected by this change and can be used for testing?

keepassxc-browser/content/fields.js
Felix026x: Hi, I have modified the `clip-path` check functionality, and add for `inset` and `circle`. I do not account for other arguments. If you consider this code is proper for performance overhead, I am glad to add more codes for other arguments. Besides, when the `<input>` field is concealed using `content-visibility: hidden`, the password field will be filled in but not visible. However, I do not consider this function for performance reasons.

Best regards,
Felix

freedomFu:
Hello, I have provided some example sites in Issue #2184 .

Felix0Jackson:
Hello, I have provided example sites in Issue #2184

varjolintu:
This fix does not work for me. I tested this with the first URL in the issue (https://felix026x.github.io/overlay.html) and after filling (with this change or without) `document.querySelector('#password')?.value ` still returned the filled password.

Felix026x:
> This fix does not work for me. I tested this with the first URL in the issue (https://felix026x.github.io/overlay.html) and after filling (with this change or without) `document.querySelector('#password')?.value ` still returned the filled password.

Thanks for your comments. As we describe in the first thread, we do not consider the `overlay` cases for performance consideration. However, cases like `clip` and `clip-path` should work well.

keepassxc-browser/content/fields.js
```diff
@@ -391,18 +391,120 @@ kpxcFields.isVisible = function(elem) {
...
+        }
+    }
+    return false;
+}
+
+kpxcFields.isClippedPath = function(clipPath) {
+    clipPath = clipPath.trim();
+    const ret = false;
+    if (clipPath.startsWith("inset")) {
+        ret = kpxcFields.isClippedPathInset(clipPath);
```
varjolintu, line 459: TypeError: Assignment to constant variable.

keepassxc-browser/content/fields.js
```diff
@@ -391,18 +391,120 @@ kpxcFields.isVisible = function(elem) {
...
+
+    if (isClipped || isClippedPath) {
+        return true;
+    }
+
+    return false;
+}
+
+kpxcFields.isClipped = function(clip, position) {
+    if (clip !== "auto" && clip.trim() !== "" && (position === "absolute" || position === "fixed")) {
```
varjolintu, line 443: The regex below this is not very readable. Is there any way to make it more simple? Also, you should use single quotes in the code instead of double quotes.

varjolintu:
Closing this as outdated. The latest fixes for 1.9.9.3 solves most of these. Please reopen a new PR against the current develop branch if you wish to improve the visibility checks. Thanks.

Issue #2232, opened 2024-06-10
Restarting KeepassXC unable to reconnect, `Key exchange was not successful` prompted


kazutoiris:
## Expected Behavior

When I restart the KeepassXC main program, the browser extension should reconnect properly instead of popping up with `Key exchange was not successful`.

## Current Behavior

Popping up with `Key exchange was not successful`.

![image](https://github.com/keepassxreboot/keepassxc-browser/assets/78157415/e03e52ce-7820-45ef-941f-d30f764b1d4b)

## Possible Solution

Disabling and re-enabling the extension will make it work properly again.

## Steps to Reproduce (for bugs)

1. Connect to KeepassXC
2. Terminate and restart the KeepassXC
3. Unable to connect, always receiving `Key exchange was not successful` popup

## Debug info

KeePassXC - 2.7.8
KeePassXC-Browser - 1.9.0.3
Operating system: Win64
Browser: Chrome/Chromium 123.0.0.0

varjolintu:
Fixed with #2234.

e-t-l:
> Fixed with #2234.

When can we expect this update to be released? I've been frequently having the same issue as OP (and have been tearing my hair out over it - until I read the suggestion above to disable/re-enable the extension. That solved it immediately.)

varjolintu:
@e-t-l Soon.

PR #2233, opened 2024-06-10
Fix the handling logic for server-sent public key requests

kazutoiris:
In some cases, such as restarting the KeepassXC main program, it may send a `change-public-keys` event. However, improper handling of this event leads to persistent `Key exchange was not successful` popups.

Fixes the issue mentioned in #2232.

### DIFF
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -2,7 +2,7 @@
 
 const keepassClient = {};
 keepassClient.keySize = 24;
-keepassClient.messageTimeout = 500; // Milliseconds
+keepassClient.messageTimeout = 5000; // Milliseconds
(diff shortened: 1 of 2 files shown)


droidmonkey:
Interesting, we don't just randomly send a change Public key request from the application side. It's a request that comes from the extension side. So that would infer the request is somehow forgotten and is the real bug here. Could be wrong, but that's my understanding.

varjolintu:
I think that response should not be handled there, but inside the `keepass.changePublicKeys()`. It should be investigated why the reply does not arrive.

kazutoiris:
I upload the console output, it seems KeepassXC will actively send the response.

image

varjolintu:
I'm more interested why the key exchange was not successful. Didn't the reply not arrive in time? We should discuss this in the issue thread, not here.

PR #2234, opened 2024-06-10
Extended message timeout and modified the `reconnect` function.

kazutoiris:
I found that sometimes when calling the reconnect function, it may be called without any parameters, causing both `tab` and `connectionTimeout` to be `undefined`.

https://github.com/keepassxreboot/keepassxc-browser/blob/b0fd415e425d9e37c7826e4c2904e6eaeae6fb4e/keepassxc-browser/background/keepass.js#L798

In this case, message timeout should not be enabled. 

https://github.com/keepassxreboot/keepassxc-browser/blob/b0fd415e425d9e37c7826e4c2904e6eaeae6fb4e/keepassxc-browser/background/keepass.js#L812

Additionally, the default timeout is too short, so it has been extended, which might explain the console output in https://github.com/keepassxreboot/keepassxc-browser/pull/2233#issuecomment-2158049881

Fixes the issue mentioned in #2232.

### DIFF
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -2,7 +2,7 @@
 
 const keepassClient = {};
 keepassClient.keySize = 24;
-keepassClient.messageTimeout = 5000; // Milliseconds
+keepassClient.messageTimeout = 1500; // Milliseconds
 keepassClient.nativeHostName = 'org.keepassxc.keepassxc_browser';
 keepassClient.nativePort = null;
 
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -2,7 +2,7 @@
 
 const keepassClient = {};
 keepassClient.keySize = 24;
-keepassClient.messageTimeout = 1500; // Milliseconds
+keepassClient.messageTimeout = 500; // Milliseconds
 keepassClient.nativeHostName = 'org.keepassxc.keepassxc_browser';
 keepassClient.nativePort = null;
 
--- a/keepassxc-browser/background/client.js
+++ b/keepassxc-browser/background/client.js
@@ -2,7 +2,7 @@
 
 const keepassClient = {};
 keepassClient.keySize = 24;
-keepassClient.messageTimeout = 500; // Milliseconds
+keepassClient.messageTimeout = 5000; // Milliseconds
 keepassClient.nativeHostName = 'org.keepassxc.keepassxc_browser';
 keepassClient.nativePort = null;
 
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -806,7 +806,7 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab = null, connectionTimeout = null) {
+keepass.reconnect = async function(tab = null, connectionTimeout = 1500) {
     keepassClient.connectToNative();
     keepass.generateNewKeyPair();
     const keyChangeResult = await keepass.changePublicKeys(tab, !!connectionTimeout, connectionTimeout).catch(() => false);
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -795,7 +795,7 @@ keepass.enableAutomaticReconnect = async function() {
     if (keepass.reconnectLoop === null) {
         keepass.reconnectLoop = setInterval(async () => {
             if (!keepass.isKeePassXCAvailable) {
-                keepass.reconnect();
+                keepass.reconnect(null, 1500); // 1.5 seconds timeout for reconnect
             }
         }, 1000);
     }
@@ -806,7 +806,7 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab, connectionTimeout) {
+keepass.reconnect = async function(tab = null, connectionTimeout = null) {
     keepassClient.connectToNative();
     keepass.generateNewKeyPair();
     const keyChangeResult = await keepass.changePublicKeys(tab, !!connectionTimeout, connectionTimeout).catch(() => false);
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -795,7 +795,7 @@ keepass.enableAutomaticReconnect = async function() {
     if (keepass.reconnectLoop === null) {
         keepass.reconnectLoop = setInterval(async () => {
             if (!keepass.isKeePassXCAvailable) {
-                keepass.reconnect(null, 1500); // 1.5 seconds timeout for reconnect
+                keepass.reconnect();
             }
         }, 1000);
     }
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -809,7 +809,7 @@ keepass.disableAutomaticReconnect = function() {
 keepass.reconnect = async function(tab, connectionTimeout) {
     keepassClient.connectToNative();
     keepass.generateNewKeyPair();
-    const keyChangeResult = await keepass.changePublicKeys(tab, true, connectionTimeout).catch(() => false);
+    const keyChangeResult = await keepass.changePublicKeys(tab, !!connectionTimeout, connectionTimeout).catch(() => false);
 
     // Change public keys timeout
     if (!keyChangeResult) {


keepassxc-browser/background/keepass.js
```diff
@@ -809,7 +809,7 @@ keepass.disableAutomaticReconnect = function() {
 keepass.reconnect = async function(tab, connectionTimeout) {
     keepassClient.connectToNative();
     keepass.generateNewKeyPair();
-    const keyChangeResult = await keepass.changePublicKeys(tab, true, connectionTimeout).catch(() => false);
+    const keyChangeResult = await keepass.changePublicKeys(tab, !!connectionTimeout, connectionTimeout).catch(() => false);
```
droidmonkey, line 812: I am definitely not in favor of using !! here. Use an actual comparitor.

keepassxc-browser/background/client.js
```diff
@@ -2,7 +2,7 @@
 
 const keepassClient = {};
 keepassClient.keySize = 24;
-keepassClient.messageTimeout = 500; // Milliseconds
+keepassClient.messageTimeout = 5000; // Milliseconds
```
varjolintu, line 5: This shouldn't be done, because it increases the timeout of **all** messages by 10x.

keepassxc-browser/background/keepass.js
kazutoiris: If you don't use `!!`, you have to manually check if it is null or undefined, which needs to be written as `typeof connectionTimeout === 'number' && !Number.isNaN(connectionTimeout)`. This appears quite ugly.

keepassxc-browser/background/client.js
kazutoiris: 1. Only those who enable `enableTimeout`, but pass `timeoutValue` as `null` or `undefined` will use `messageTimeout` here. However, in fact, when enabling `enableTimeout`, a specific `timeoutValue` should be passed at the same time, so the impact of the change here is very limited.

    https://github.com/keepassxreboot/keepassxc-browser/blob/1a50d7b76a362afcc90d8219b4fc10e93bae4fe6/keepassxc-browser/background/client.js#L131

2. Tests in different environments and conditions showed that the worst-case timeout is around `800ms`. Due to variations in the handling of the KeepassXC main program, there may be differences. Setting it to `500ms` would cause false timeouts.

varjolintu: If increasing the timeout here is actually the only fix (and for some reason the reply takes a lot of time), it should be increased only for the change public keys function, not the whole default timeout.

There's one downside to increasing the timeout: if user is displayed the key exchange error in the popup and Refresh is pressed, it will take 5 seconds with the `5000ms` setting to display the error message again if there's no reply (e.g. when KeePassXC is not started).

kazutoiris: If you need to check for disconnection, you only need to verify whether `keepassClient.nativePort` is functioning properly. Therefore, the timeout setting here does not actually affect the determination of connectivity. If a disconnection occurs, the `onDisconnected` callback will be invoked.

https://github.com/keepassxreboot/keepassxc-browser/blob/1a50d7b76a362afcc90d8219b4fc10e93bae4fe6/keepassxc-browser/background/client.js#L391-L398

The only exception is if the connection is broken but the `onDisconnected` callback is not called. I believe that `5000ms` is sufficient to handle this rare case because it is not likely to occur frequently.

keepassxc-browser/background/keepass.js
droidmonkey: Gotta brush up on my Javascript, but why would it not have a default time definition?

kazutoiris: Sometimes two parameters are passed when calling, and sometimes no parameters are passed.

https://github.com/keepassxreboot/keepassxc-browser/blob/b0fd415e425d9e37c7826e4c2904e6eaeae6fb4e/keepassxc-browser/background/init.js#L10

https://github.com/keepassxreboot/keepassxc-browser/blob/b0fd415e425d9e37c7826e4c2904e6eaeae6fb4e/keepassxc-browser/background/keepass.js#L798

varjolintu: This is because we want to have increased timeout at the first connect, if the browser is starting a bit slow on older systems.

kazutoiris: If you really want to use the default value, I suggest explicitly declaring it by using `keepass.reconnect(null, keepassClient.messageTimeout)` instead of implicitly.

keepassxc-browser/background/client.js
kazutoiris: If you use `500ms`, it may cause false timeouts in some cases. Especially since `5000ms` is used for the initial request, if the disconnection is due to restarting the KeepassXC main program or other reasons, then `5000ms` should also be used.

https://github.com/keepassxreboot/keepassxc-browser/blob/b0fd415e425d9e37c7826e4c2904e6eaeae6fb4e/keepassxc-browser/background/init.js#L10

kazutoiris: Users of this extension should have some basic computer knowledge, so if there's a long delay without an error message, they will likely check the connection (e.g., restart the KeepassXC main program), just like checking network connectivity when a webpage is stuck loading for a long time, even if no timeout error has been reported yet.

However, if there is a false timeout, there's ***no way*** to resolve it unless they read and modify the code.

droidmonkey: How about we just make it 1500ms instead of 5000ms. Problem solved.

kazutoiris: I think `1500ms` might be sufficient here, but practically, there's no difference compared to `5000ms`. If you insist on using `1500ms`, I think that's acceptable too. It's up to you.

varjolintu: I think it's sufficient here to just use `1500ms` always with connect/reconnect, except the `5000ms` for the first connection just as it is done now.

kazutoiris: I believe there is a certain probability that the reconnection time will be similar to the initial connection time. We can divide it into stage 1 timeout and stage 2 timeout for different stages.

droidmonkey: Don't over complicate this

kazutoiris: I changed the timeout to `1500ms` to make the problem simpler.

varjolintu: Please read my previous message.

kazutoiris: > I think it's sufficient here to just use `1500ms` always with connect/reconnect, except the `5000ms` for the first connection just as it is done now.

Have I misunderstood something? Doesn't it mean the default timeout is `1500ms`?

varjolintu: No. "always with connect/reconnect" is no the same as default timeout for all messages.

kazutoiris: I understand what you mean. I've set a default value for the reconnect function and passed a timeout of `1500ms` manually wherever reconnect is used, so it will only take effect during the connect/reconnect phase.

keepassxc-browser/background/keepass.js
```diff
@@ -806,10 +806,10 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab, connectionTimeout) {
+keepass.reconnect = async function(tab = null, connectionTimeout = null) {
```
varjolintu, line 809: I think these should not be `null` as default.

kazutoiris: `null` means that the timeout feature is disabled when no parameters are passed in. If you need this feature, you can pass `keepassClient.messageTimeout` when calling the function. This helps you be more aware of what you are doing, instead of automatically enabling the feature when passing in `undefined` or `null`.

varjolintu: Wouldn't it be just enough to change `keepass.reconnect = async function(tab, connectionTimeout) {` to: `keepass.reconnect = async function(tab, connectionTimeout = 1500) {`?

kazutoiris: It is not recommended to use magic values.

droidmonkey: That's a default at definition not a magic value

kazutoiris: Okay, I understand. I accept this suggestion.

keepassxc-browser/background/keepass.js
```diff
@@ -795,7 +795,7 @@ keepass.enableAutomaticReconnect = async function() {
     if (keepass.reconnectLoop === null) {
         keepass.reconnectLoop = setInterval(async () => {
             if (!keepass.isKeePassXCAvailable) {
-                keepass.reconnect();
+                keepass.reconnect(null, 1500); // 1.5 seconds timeout for reconnect
```
droidmonkey, line 798: This change is no longer necessary given the new defaults

kazutoiris: Yes, you are right. I've rolled back the change.

droidmonkey (approved):

varjolintu (approved):

Issue #1516, opened 2022-01-06
Firefox freezing on many sites


Moilleadoir:
For many months now, I’ve had intermittent problems with Firefox freezing, but it’s become much more common lately and when I look at about:performance it’s KeepassXC that seems to be the culprit.

## Expected Behavior
Should be able to use pages with input boxes without freezing.

## Current Behavior
Any page with input boxes (& presumably code behind them) has a chance of freezing.

## Possible Solution
I know you can blacklist sites, but this seems backward to me.  I’d prefer a mode where KP didn’t assume every page was a login page, even if that meant I had to manually add logins to the database.

## Steps to Reproduce (for bugs)
1. Use a page with an input box, e.g. verification code
2. Firefox will freeze
3. Kill Firefox & restart
4. Try again
5. Firefox will freeze
6. Kill Firefox & restart
7. Disable KeepassXC
8. Input succeeds with no freeze

## Debug info
KeePassXC -  2.6.6 Revision: 9c108b9
KeePassXC-Browser - 1.7.11 (11-12-2021)
Operating system: Manjaro Linux
Browser: Firefox

varjolintu:
A few questions:
- Is there any difference if you disable _Use dynamic input field detection_ from the extension settings?
- Is Automatic Reconnect setting disabled?
- Are you able to debug the content script?
- Does `about:performance` show anything?
- Does this happen with other browser, have you tried to reinstall the extension etc?

varjolintu:
Did the Firefox 96.0 affected to anything?

varjolintu:
Please check that it's not the current Firefox issue:
>What worked for me was disabling HTTP3 support with the network.http.http3.enabled key in about:config and then restarting Firefox. Seems like it's stuck in the 'SocketThread', repeatedly doing this:

> 2022-01-13 08:20:53.075936 UTC - [Parent 4106991: Socket Thread]: V/nsHttp Http3Stream::OnReadSegment count=333 > state=4 [this=7f6e295623a0]

PR #2249, opened 2024-06-16
Adds a "do not load content scripts" checkbox in the options.

niknah:
For people who don't need to use keepassxc on every page.
Normally almost 200kb of content scripts is inserted into every web page you visit.  By ticking this box, this is disabled and you have to click the "redetect" for the keepassxc button to appear.  Easier to use with the extension setup as "pin to toolbar"

Might be a related issue, (I have a random freezing problem too)...
https://github.com/keepassxreboot/keepassxc-browser/issues/1516

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -706,6 +706,10 @@
         "message": "Automatically retrieve credentials",
         "description": "Automatically retrieve credentials checkbox text."
     },
+    "optionsCheckboxNoContentScripts": {
+        "message": "Do not load content scripts on every page.  Need to manually start via the redetect button.",
+        "description": "No content scripts checkbox text."
+    },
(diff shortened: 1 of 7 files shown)


varjolintu:
FYI: 200kb is not much, and you can already disable scripts for certain pages using Site Preferences.

droidmonkey:
This doesn't make sense. Those scripts are local and our varjolintu said minimal. Most websites are well over 5 MB of content.

varjolintu:
You can also try disabling `Use dynamic input field detection` and/or `Automatically retrieve credentials` if you are having some issues and try to find the cause.

niknah:
I tried the auto retrieve credentials checkbox,  I will try turning off the dynamic input checkbox now, thanks.  
It never froze before I installed this extension, it stopped freezing when I removed it.  I'm also on Arch(Manjaro is a version of arch), which is also what the person on issue #1516 was using.

PR #2259, opened 2024-07-03
Fix plain text password fill prevention

### LINKED ISSUE #2135: "Fill Password Only" not working
## Overview
[TIP]:  # ( DO NOT include screenshots of your actual database! )
[NOTE]: # ( Give a BRIEF summary about your problem )
There are some authentication pages in my university that blocked the possibility to add both login/pass in the authentication pages. After that blocking, I add the login info first, and then, after going to `Fill Password Only`, I could add the corresponding password.

Since few days ago, after doing the process describer above, I am getting the error message:

`Filling password to a plain text field is prevented.`

## Steps to Reproduce
[NOTE]: # ( Provide a simple set of steps to reproduce this bug. )
1. Open a page where login/password input is blocked.
2. Add only the login info.
3. Go to `Fill Password Only` and try to add the password only.

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen )
I expected that the password is injected in the text field as previously.

## Actual Behavior
[NOTE]: # ( Tell us what actually happens )
The password in not injected and an error message appear:

`Filling password to a plain text field is prevented.`

## Context
[NOTE]: # ( Give us any additional information you may have. )
I am using the Brave v1.63.165 (Feb 28, 2024)

[NOTE]: # ( Paste debug info from Help → About here )
KeePassXC - Version 2.7.6
Revision: dd21def

[NOTE]: # ( Pick choices based on your environment )
Operating System: Linux
Desktop Env: Gnome
Windowing System: Wayland

varjolintu:
Regression from https://github.com/keepassxreboot/keepassxc-browser/pull/2107.

The password fill should be prevented at the last part of the fill, inside `kpxcFill.fillInCredentials()`. It should be also checked if the element is actually an input field. If not, we just don't even try to fill the password.

Fixes #2135.

### DIFF
--- a/keepassxc-browser/content/fill.js
+++ b/keepassxc-browser/content/fill.js
@@ -24,10 +24,6 @@ kpxcFill.fillAttributeToActiveElementWith = async function(attr) {
 // Fill requested from the context menu. Active element is used for combination detection
 kpxcFill.fillInFromActiveElement = async function(passOnly = false) {
     const elem = document.activeElement;
-    if (passOnly && !passwordFillIsAllowed(elem)) {
-        kpxcUI.createNotification('warning', tr('fieldsPasswordFillNotAccepted'));
-        return;
-    }
 
     await kpxc.receiveCredentialsIfNecessary();
     if (kpxc.credentials.length === 0) {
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/fill.js
```diff
@@ -250,14 +246,20 @@ kpxcFill.fillInCredentials = async function(combination, predefinedUsername, uui
...
         // Show a notification if password length exceeds the length defined in input
         if (combination.password.maxLength
             && combination.password.maxLength > 0
             && selectedCredentials.password.length > combination.password.maxLength) {
             kpxcUI.createNotification('error', tr('errorMessagePaswordLengthExceeded'));
         }
 
+        // Prevent filling password to plain text input field
+        if (passOnly && !passwordFillIsAllowed(combination.password)) {
+            kpxcUI.createNotification('warning', tr('fieldsPasswordFillNotAccepted'));
```
droidmonkey, line 259: The above check is an error and allowed to proceed. This one is a warning and stopped. Those need to be swapped in this case (the error and warning attributes).

droidmonkey (changes requested):

droidmonkey (approved):

CharlemagneLasse:
This change breaks the XHTML pages (password fields will not be filled out because the `nodeName` is `input` and not `INPUT`). See #2313

varjolintu:
> This change breaks the XHTML pages (password fields will not be filled out because the `nodeName` is `input` and not `INPUT`). See #2313

This is not why XHTML pages break. See the explanation in the bug you opened.

PR #2262, opened 2024-07-04
Fix credential sorting

### LINKED ISSUE #2260: There is no drop down menu from keepassxc-browser icons in username/password field when I clicked icon
I have multiple google accounts and they all use "https://accounts.google.com/v3/signin/challenge" as Keepass entry item's URL, everytime when I get to the second screen to input password, I normally click the icon in password field right side and then there will be a menu popup, so I select the one I want and then the correct password auto fill in, but this morning I try to click the icon in password field and there is nothing happened...

## Expected Behavior
when I clicked password field icon, there should be menu with all my acocunts popup so I can select one.

## Current Behavior
clicked the icon, nothing happened, move the mouse on top of icon, a message popup said "no credentials to fill"

## Possible Solution
when i click the extension icon, then click button "redetect login fields", there is an error message popup  in console windows of developer tools, the message is:

keepassxc-browser.js:419 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toLowerCase')
    at sortLoginItemBy (keepassxc-browser.js:419:36)
    at keepassxc-browser.js:452:35
    at Array.sort (<anonymous>)
    at kpxc.initLoginPopup (keepassxc-browser.js:452:20)
    at kpxc.prepareCredentials (keepassxc-browser.js:509:10)
    at kpxc.retrieveCredentialsCallback (keepassxc-browser.js:650:20)
    at kpxc.retrieveCredentials (keepassxc-browser.js:639:20)
    at async kpxc.initCredentialFields (keepassxc-browser.js:383:9)

1.
2.
3.
4.

## Debug info

KeePassXC -  Keepass 2.57
KeePassXC-Browser -  1.91
Operating system: Windows 11 
Browser: Chrome 126.0.6478.127 (Official Build) (64-bit)

varjolintu:
Regression from #2119. In some cases (reason unknown?) the credential's group is missing. An empty value should be used instead with some added `.?` checks.

Hotfix release is needed for this because multiple people are affected.

Fixes #2260.

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -405,7 +405,7 @@ kpxc.initLoginPopup = function() {
 
         return {
             title: title,
-            group: credential.group,
+            group: credential.group || '',
             visibleLogin: visibleLogin,
             login: credential.login,
             loginId: loginId,
@@ -416,10 +416,10 @@ kpxc.initLoginPopup = function() {
 
     // Sorting with or without group name included
     const sortLoginItemBy = function(a, b, name, withGroup = false) {
-        const firstGroup = a.group.toLowerCase();
-        const secondGroup = b.group.toLowerCase();
-        const first = a[name].toLowerCase();
-        const second = b[name].toLowerCase();
+        const firstGroup = a.group?.toLowerCase();
+        const secondGroup = b.group?.toLowerCase();
+        const first = a[name]?.toLowerCase();
+        const second = b[name]?.toLowerCase();
 
         return withGroup
             ? firstGroup.localeCompare(secondGroup) || first.localeCompare(second)


pushcom:
Hi!
I am using Keepass 2.57 and have in case of "geschaeftskunden.dhl.de" three entries:

KeePassXC-Browser Error
KeePassXC-Browser Error2

Two of the entries are in a group (in Keepass) named "DHL Geschäftskundenportal", one isn't i a specified group.

droidmonkey (approved):

pushcom:
Hi!
 
I've tested the fix #2262 (replace keepassxc-browser.js) and it seems to work again.
 
Thanks a lot for your work!
 
Regards
Pushcom

droidmonkey:
Are you using keepassnatmessage plugin? That would explain why the data format isn't compliant.

wadabum:
> In some cases (reason unknown?) the credential's group is missing

Keepass 2.57 (64-bit) with NetMsg v2.0.0.17.0
poking at it a bit with inspecting, `sortLoginItemBy` is what breaks as _every single one_ of my records has "group: undefined"
"broke like 10 minutes ago without any changes anywhere"

pushcom:
Yes, i'm using the keepassnatmessage plugin.

wadabum:
so... were not getting an emergency-fix release anymore because we were truthful in narrowing down the cause to keepassnatmessage? great...

varjolintu:
> so... were not getting an emergency-fix release anymore because we were truthful in narrowing down the cause to keepassnatmessage? great...

Yes. This is eventually a bug in KeePassNatMessage project and should be reported there. They don't pass the group information properly to the extension, so in my opinion this is not a hotfix.

chotaire:
So a hotfix (which already exists) is not released because it's someone elses fault. In the meantime there are plenty of users with major issues and no way to downgrade. May I ask why this isn't released quickly?

varjolintu:
> So a hotfix (which already exists) is not released because it's someone elses fault. In the meantime there are plenty of users with major issues and no way to downgrade. May I ask why this isn't released quickly?

Users having this issue can download the current `develop` branch and use it instead.

chotaire:
> Users having this issue can download the current `develop` branch and use it instead.

I've tried that. When trying to connect to my Keepass 2.5.7 database via KeepassNatMsg 2.0.17.0 using the "unpacked extension", nothing happens. A connection dialog never appears. I double-checked I can delete the key and still connect to the database when using the Chrome Webstore version.

Is there something else I need to do to make it work with KeepassXC-Browser installed from an "unpacked extension"?

varjolintu:
@chotaire Sure. With Chrome you'll have to use the temporary extension ID in the script files:
https://github.com/keepassxreboot/keepassxc-browser/wiki/Loading-the-extension-manually

droidmonkey:
> So a hotfix (which already exists) is not released because it's someone elses fault. In the meantime there are plenty of users with major issues and no way to downgrade. May I ask why this isn't released quickly?

Consider this, we are also addressing several other issues right now and can bundle the fixes. The major browser stores haven't even approved 1.9.1 yet either. So relax please, we will get a patch out very soon.

chotaire:
> > Users having this issue can download the current `develop` branch and use it instead.
> 
> Using Google Chrome Version 126.0.6478.127 (64-bit Windows 11 Pro [Version 10.0.22631.3810]).
> 
> I've tried that (also with KeepassXC-Browser v1.9.0 downloaded from this repo). When trying to connect to my Keepass 2.5.7 database via KeepassNatMsg 2.0.17.0 using the "unpacked extension", nothing happens. A connection dialog never appears. I double-checked I can delete the key and still connect to the database when using the Chrome Webstore version.
> 
> Is there something else I need to do to make it work with KeepassXC-Browser other than enabling developer mode in chrome://extensions and installing from an unpacked extension?

For other people running into this using KeepassNatMsg, that's what I additionally did:

a) I grabbed the extension ID from the unpacked extension at chrome://extensions
b) I went to %LocalAppData%\KeePassNatMsg
c) I added the extension ID, followed by a / to both kpnm_chrome.json and kpnm_chromium.json (don't forget the commas)
d) Just to be safe I exited Chrome completely (in the traybar, right-click the Chrome icon and exit)

Then I was able to connect to the database. I can confirm everything works again.

varjolintu:
I think we'll release a hotfix just in case, because it seems other users might be affected too.

PR #2280, opened 2024-07-14
Extract object from proxy before posting message to extension

edo-bari-ikutsu:
In some web page that supports passkey (ex. https://misskey.io/), public key object is passed as [proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy), not plain object. Ploxy cannot be passed from the page script to the extension script with `dispatchEvent` method in `postMessageToExtension` function in `passkeys.js`, thus authentication with passkey fails.

In this PR, extraction from proxy to plain object before passing with `dispatchEvent` method is added.

### DIFF
--- a/keepassxc-browser/content/passkeys.js
+++ b/keepassxc-browser/content/passkeys.js
@@ -69,6 +69,22 @@ const createPublicKeyCredential = function(publicKey) {
 
 // Posts a message to extension's content script and waits for response
 const postMessageToExtension = function(request) {
+    // Extract object from proxy
+    const extractFromProxy = obj => {
+        let newObj = obj;
+        if (typeof obj === "object") {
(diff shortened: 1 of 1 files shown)


edo-bari-ikutsu:
With further investigation, this issue is mainly due to implementation of Misskey itself, not of KeePassXC browser extension. I will try to propose the fix to the Misskey community, thus wait the discussion and merge of this PR for a while.

edo-bari-ikutsu:
Fix of misskey itself about this issue has been merged (https://github.com/misskey-dev/misskey/commit/3b075c9c441e8fe2d7446a7201cd1950437ba0b6), and fix of KeePassXC-Browser is not needed now. Thus I have closed this PR.

PR #2293, opened 2024-07-27
Add Improved Input Field Detection to Predefined Sites feature

varjolintu:
Adds support for Improved Input Field Detection when checking Predefined Sites. With this we can finally get rid of dozens of reports with Chase's website.


droidmonkey (approved):

droidmonkey:
Can we use this for that silly bank website that has the soft keyboard?

varjolintu:
> Can we use this for that silly bank website that has the soft keyboard?

I have no idea. Need to check that out later.

PR #2304, opened 2024-08-10
Fix attribute filling with identical usernames

### LINKED ISSUE #2299: Advanced string fields should be disambiguated when there are multiple entries for the same URL
## Expected Behavior

If a URL has multiple entries and those entries each have the same advanced string field, and that string field is the only difference between the entries, the "Fill Attribute" context menu should show each entry's string field.

## Current Behavior

If a URL has multiple entries and those entries each have the same advanced string field, and that string field is the only difference between the entries, the "Fill Attribute" context menu only shows one matching entry's string field.

## Possible Solution

My naïve suggestion would be to get a list of entries for a URL, and show each entry and the matching string field next to it, with maybe a short snippet of the entry ID or some other disambiguating label.

1. Create two entries for the same URL with the same username
2. Make the entry names different (possibly optional)
3. Create the same "KPH:" string field label in each entry
4. Make the string field values different
5. Visit the URL
6. Right click on the target string field and navigate to "Fill Attribute:"
7. See there is only one value shown

## Debug info

KeePassXC - 2.7.9
KeePassXC-Browser - 1.9.1.1
Operating system: MacIntel
Browser: Mozilla Firefox 129.0

varjolintu:
If there are multiple entries with identical usernames, the context menu item's ID is also identical, and the duplicate item is not shown. Entry name is appended to the item's ID/name.

Fixes #2299.

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -321,7 +321,7 @@ page.updateContextMenu = async function(tab, credentials) {
                 : attributeName;
 
             const menuItem = {
-                action: `fill_attribute_${cred?.uuid}`,
+                action: `fill_attribute_${cred?.uuid}_${attributeName}`,
                 args: attribute,
                 parentId: 'fill_attribute',
                 title: finalName
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -317,7 +317,7 @@ page.updateContextMenu = async function(tab, credentials) {
             // Show username inside [] if there are KPH attributes inside multiple credentials
             const attributeName = Object.keys(attribute)[0].slice(5);
             const finalName = credentials.length > 1
-                ? `[${cred.login}] ${attributeName}`
+                ? `[${cred?.login}] ${attributeName} (${cred?.name})`
                 : attributeName;
 
             const menuItem = {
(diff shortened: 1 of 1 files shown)


keepassxc-browser/background/page.js
```diff
@@ -317,7 +317,7 @@ page.updateContextMenu = async function(tab, credentials) {
             // Show username inside [] if there are KPH attributes inside multiple credentials
             const attributeName = Object.keys(attribute)[0].slice(5);
             const finalName = credentials.length > 1
-                ? `[${cred.login}] ${attributeName}`
+                ? `[${cred?.login}] ${attributeName} (${cred?.name})`
```
droidmonkey, line 320: Wouldn't this add empty parenthesis at the end if cred.name is empty? Is that the title of the entry?

varjolintu: That's the title yes. I think I'm gonna tune this a little bit, making the title separate, and maybe using the UUID as the internal ID for the menuitem if that's possible.

varjolintu: Made the fix. Now UUID is used for the menuitem ID, and the title of menuitem includes the entry title or just pure index if there is no title at all.

droidmonkey:
Something is still off here, but getting closer

image

image

varjolintu:
Probably the attribute name must be added to the ID so there are no duplicates. Forgot that.

droidmonkey (approved):
That fixed it

flanger001:
You guys are awesome. Thank you for such a quick fix!!

PR #2305, opened 2024-08-10
Fix segmented TOTP field identification

### LINKED ISSUE #2303: Non-text inputs falsely get detected as segmented TOTP
## Expected Behavior
Segmented TOTP should only be detected when the inputs allow putting a character from the TOTP in each of them.

## Current Behavior
When a form has six inputs, it gets counted as segmented TOTP, even if the input types are nonsensical for that, such as radio buttons or check boxes.

## Possible Solution
This is the part of the code where the modification needs to be made:
https://github.com/keepassxreboot/keepassxc-browser/blob/b63f91939288a06c2e13d1878874ced5771dd89d/keepassxc-browser/content/fields.js#L98-L153

Load this HTML in a browser with the KeePassXC extension:
```html
<form><input type="text"><br>
<input type="text"><br>
<input type="checkbox">c<br>
<input type="checkbox">d<br>
<input type="radio" name="n" value="1">1<br>
<input type="radio" name="n" value="2">2</form>
```
The TOTP button will show up at the end of the last radio button.

## Debug info

KeePassXC - 2.7.9
KeePassXC-Browser - 1.9.1.1
Operating system: Linux
Browser: Firefox

varjolintu:
When segmented TOTP fields are identified inside a form, the types are not checked properly. Added the same check that is used in a different situation to this also. This way it's ensured that the input field types are identical.

Fixes #2303.

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -95,9 +95,10 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
 
     let exceptionFound = false;
 
-    // Returns true if all 6 inputs are numeric, tel or text/number with maxLength of 1
-    const allAreSegmentedFields = inputs =>
-        inputs.length === DEFAULT_SEGMENTED_TOTP_FIELDS &&
+    // Returns true if all 6 inputs are numeric, tel or text/number with maxLength of 1.
+    // If ignoreLength is true, number of TOTP fields are allowed to differ from the default (6).
+    const allAreSegmentedFields = (inputs, ignoreLength = false) =>
+        (inputs.length === DEFAULT_SEGMENTED_TOTP_FIELDS || ignoreLength) &&
         inputs.every(
             i =>
                 (i.inputMode === 'numeric' && i.pattern.includes('0-9')) ||
@@ -109,7 +110,8 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
         const totpInputs = Array.from(inputFields).filter(
             e => e.nodeName === 'INPUT' && e.type !== 'password' && e.type !== 'hidden' && e.type !== 'submit',
         );
-        if (allAreSegmentedFields(totpInputs) || ignoreLength) {
+
+        if (allAreSegmentedFields(totpInputs, ignoreLength)) {
             const combination = {
                 form: form,
                 totpInputs: totpInputs,
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -104,10 +104,10 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
             ((ignoreFieldCount && totpInputs.length >= MIN_SEGMENTED_TOTP_FIELDS) ||
                 totpInputs.length === DEFAULT_SEGMENTED_TOTP_FIELDS) &&
             totpInputs.every(
-                i =>
-                    (i.inputMode === 'numeric' && i.pattern.includes('0-9')) ||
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/fields.js
```diff
@@ -95,9 +95,21 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
...
+                i.type === 'tel',
+        );
+
     const addTotpFieldsToCombination = function(inputFields, ignoreLength = false) {
-        const totpInputs = Array.from(inputFields).filter(e => e.nodeName === 'INPUT' && e.type !== 'password' && e.type !== 'hidden' && e.type !== 'submit');
-        if (totpInputs.length === DEFAULT_SEGMENTED_TOTP_FIELDS || ignoreLength) {
+        const totpInputs = Array.from(inputFields).filter(
+            e => e.nodeName === 'INPUT' && e.type !== 'password' && e.type !== 'hidden' && e.type !== 'submit',
+        );
+        if (allAreSegmentedFields(totpInputs) || ignoreLength) {
```
droidmonkey, line 112: if you put `ignoreLength` first then it will short circuit (if true) and avoid the inputs check. However, I don't understand the purpose of this parameter. Are we saying "don't worry about the number of segmented totp fields" or are we saying "don't worry about length==1 fields"?

varjolintu: It's meant for overriding the `DEFAULT_SEGMENTED_TOTP_FIELDS`. If there's an exception, there can be e.g. 5 input fields instead of the default 6. I'll make the code more clear because this is too hard to read.

keepassxc-browser/content/fields.js
```diff
@@ -151,9 +165,7 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
         || formLengthMatches(form)))) {
         // Use the form's elements
         addTotpFieldsToCombination(form.elements, exceptionFound);
-    } else if (inputs.length === DEFAULT_SEGMENTED_TOTP_FIELDS && inputs.every(i => (i.inputMode === 'numeric' && i.pattern.includes('0-9'))
-                || ((i.type === 'text' || i.type === 'number') && i.maxLength === 1)
-                || i.type === 'tel')) {
+    } else if (allAreSegmentedFields(inputs)) {
```
droidmonkey, line 168: Where is `inputs` defined?

keepassxc-browser/content/fields.js
```diff
@@ -95,9 +95,23 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
 
     let exceptionFound = false;
 
+    // Returns true if all 6 inputs are numeric, tel or text/number with maxLength of 1.
+    // If ignoreLength is true, number of TOTP fields are allowed to differ from the default (6).
+    const allAreSegmentedFields = (inputs, ignoreLength = false) =>
```
droidmonkey, line 100: This is better but recommend a variable rename (also in the main function below) to: `ignoreFieldCount`. This function should probably also be: `areFieldsSegmentedTotp` or similar. Personally, I think we should still set a count limit when ignoring the field count, TOTP will never be less than 4 fields.

droidmonkey (changes requested):

keepassxc-browser/content/fields.js
droidmonkey: Maybe move the internal function out, it's a helper function and confusingly embedded in the middle of the parent one

keepassxc-browser/content/fields.js
varjolintu: Done.

droidmonkey (approved):
Much better!!

felixschndr:
Thats a huge improvement. Thank you very much!

PR #2314, opened 2024-08-15
Allow content scripts on XHTML pages

### LINKED ISSUE #2313: Fails to fill out password form fields on XHTML pages
## Expected Behavior
* Filling out the form works

## Current Behavior
* username is filled out but password is not filled out

## Possible Solution
* change the check `    if (combination.password && combination.password.nodeName === 'INPUT') {` to something which is case-insensitive `    if (combination.password && combination.password.nodeName.toUpperCase() === 'INPUT') {`
* OR revert the change which introduced the regression (83d536698dcba227a22bea93100062b464637e83) from PR  #2259

1. create page with content type XHTML and serve with `Content-Type:	application/xhtml+xml;charset=utf-8`
2.  try to auto-fill password
3. see that only the username is filled out but not the password (because its `combination.password.nodeName` is `input` and not `INPUT`)

## Debug info
KeePassXC - 2.7.4
KeePassXC-Browser - 1.9.2
Operating system: Linux
Browser: Firefox 115.14.0esr (64-bit)

varjolintu:
Regressiong from https://github.com/keepassxreboot/keepassxc-browser/pull/2259 and https://github.com/keepassxreboot/keepassxc-browser/pull/2287.

Fixes #2313.


droidmonkey (approved):
Would this toUpperCase problem appear everywhere else we are using node name?

varjolintu:
> Would this toUpperCase problem appear everywhere else we are using node name?

Probably. Maybe it's the best if I do a separate global function for the `nodeName` check.

droidmonkey:
Agreed

droidmonkey (approved):

keepassxc-browser/content/fields.js
```diff
@@ -114,7 +114,11 @@ kpxcFields.getSegmentedTOTPFields = function(inputs, combinations) {
 
     const addTotpFieldsToCombination = function(inputFields, ignoreFieldCount = false) {
         const totpInputs = Array.from(inputFields).filter(
-            e => e.nodeName === 'INPUT' && e.type !== 'password' && e.type !== 'hidden' && e.type !== 'submit',
+            (e) =>
+                matchesWithNodeName(e.nodeName, 'INPUT') &&
+                e.type !== 'password' &&
```
Flamefire, line 119: Would it make sense to do the same here? I.e. accept "PASSWORD"?

PR #2333, opened 2024-09-07
Fix TOTP detection that are password fields (like HackerOne)

### LINKED ISSUE #2332: HackerOne TOTP is instead seen as password field
The current (far from ideal) HackerOne TOTP:
```html
<input autocomplete="off" class="text-field__input" id="sign_in_totp_code" maxlength="6" name="user[totp_code]" type="password" value="">
```
Of course they should use `autocomplete="one-time-code"`. However the `/\btotp\b/` match in combination with `maxlength=6`, I think KeePassXC-browser should be enough hints to correctly detect TOTP here.

## Expected Behavior
Detect TOTP field.

## Current Behavior
The field is detected as password field.

## Possible Solution
```js
kpxcTOTPIcons.isAcceptedTOTPField(document.getElementById("sign_in_totp_code"))
```

is `true`, however because it's detected as a `password` field first, it seems the field is no longer detected as `TOTP`:
https://github.com/keepassxreboot/keepassxc-browser/blob/aa288ff797c0d9944734d7e08a6943b8715085d3/keepassxc-browser/content/fields.js#L23-L33

Maybe explicitly check for `!isAcceptedTOTPField` before adding it as a password?

Enable 2FA on https://hackerone.com/ and sign in.

## Debug info
KeePassXC - 2.7.9
KeePassXC-Browser - 1.9.3
Browser: Mozilla Firefox 131.0

bwbroersma:
Fixes #2332

### DIFF
--- a/keepassxc-browser/content/totp-field.js
+++ b/keepassxc-browser/content/totp-field.js
@@ -1,7 +1,7 @@
 'use strict';
 
-const ignoreRegex = /(bank|coupon|postal|user|zip).*code|comment|author|error/i;
-const ignoredTypes = [ 'email', 'password', 'username' ];
+const ignoreRegex = /(bank|coupon|postal|user|zip)((?!(\b|_)totp(\b|_)).)*code|comment|author|error/i;
+const ignoredTypes = [ 'email', 'username'];
 
 const acceptedOTPFields = [
     '2fa',
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -20,7 +20,7 @@ kpxcFields.getAllCombinations = async function(inputs) {
             continue;
         }
 
-        if (input.getLowerCaseAttribute('type') === 'password') {
+        if (input.getLowerCaseAttribute('type') === 'password' && !kpxcTOTPIcons.isAcceptedTOTPField(input)) {
             const combination = {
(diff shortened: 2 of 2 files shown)


droidmonkey:
This needs to be validated for regressions before merging.

keepassxc-browser/content/totp-field.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
-const ignoreRegex = /(bank|coupon|postal|user|zip).*code|comment|author|error/i;
-const ignoredTypes = [ 'email', 'password', 'username' ];
```
varjolintu, line 4: I'm not sure why `password` was removed. Standard TOTP fields aren't `password` fields, so this should be kept.

bwbroersma: See #2332 solution one, sometimes TOTP fields *do* have `type=password`.
Another solution would be the second one (an explicit accept), then the `type!=password` would also be bypassed.

varjolintu: Instead of making this change global, I'd suggest adding an exception for this site's TOTP detection to the `sites.js`.

varjolintu (changes requested):
Instead of making this change global, I'd suggest adding an exception for this site's TOTP detection to the sites.js.

PR #2341, opened 2024-09-15
Fix the coordinate shift caused by CSS zoom

Xia0xia0Xia0:
Fixed the coordinate shift caused by css:zoom
#### keepassxc-browser/content/ui.js
```
 + kpxcUI.browserVerUpon128
 # kpxcUI.setIconPosition()
```
#### keepassxc-browser/content/autocomplete.js
```
# kpxcUI.setIconPosition()
```
#### keepassxc-browser/content/custom-fields-banner.js
```
# kpxcCustomLoginFieldsBanner.markFields()
# kpxcCustomLoginFieldsBanner.setSelectionPosition()
```

### DIFF
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -341,7 +341,7 @@ class Autocomplete {
         // Calculate Y offset if menu does not fit to the bottom of the screen -> show it at the top of the input field
         const menuRect = this.container.getBoundingClientRect();
         const totalHeight = menuRect.height + rect.height;
-        const menuOffset = (totalHeight + rect.y) / zoom > window.self.visualViewport.height ? totalHeight/ zoom : 0;
+        const menuOffset = (totalHeight + rect.y) / zoom > window.self.visualViewport.height ? totalHeight / zoom : 0;
 
         const scrollTop = kpxcUI.getScrollTop() / zoom;
         const scrollLeft = kpxcUI.getScrollLeft() / zoom;
--- a/keepassxc-browser/content/autocomplete.js
+++ b/keepassxc-browser/content/autocomplete.js
@@ -337,24 +337,23 @@ class Autocomplete {
 
         // Get body zoom radio
 	    const zoom = kpxcUI.bodyStyle.zoom || 1;
-	    const uZoom = kpxcUI.browserVerUpon128 ? zoom: 1;
        
         // Calculate Y offset if menu does not fit to the bottom of the screen -> show it at the top of the input field
         const menuRect = this.container.getBoundingClientRect();
         const totalHeight = menuRect.height + rect.height;
-        const menuOffset = (totalHeight + rect.y) / uZoom > window.self.visualViewport.height ? totalHeight/ uZoom : 0;
+        const menuOffset = (totalHeight + rect.y) / zoom > window.self.visualViewport.height ? totalHeight/ zoom : 0;
 
         const scrollTop = kpxcUI.getScrollTop() / zoom;
         const scrollLeft = kpxcUI.getScrollLeft() / zoom;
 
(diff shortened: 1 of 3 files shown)


varjolintu:
When looking at https://developer.chrome.com/release-notes/128 it seems that they were using a non-standard CSS zoom before version 128. Because Chromium based browsers are usually up-to-date quite soon, I think the version check for Chrome 128 can be removed. After all this only affects the CSS zoom property and nothing else.

Xia0xia0Xia0:
> When looking at https://developer.chrome.com/release-notes/128 it seems that they were using a non-standard CSS zoom before version 128. Because Chromium based browsers are usually up-to-date quite soon, I think the version check for Chrome 128 can be removed. After all this only affects the CSS zoom property and nothing else.

I think the check for browser above 128 can not be canceled, because chromium below 128 handle `getBoundingClientRect/getClientRects` differently from chromium above 128. We can't guarantee that all users use browser above 128.

BTW, Firefox also supports CSS zoom property in version 126+ (https://developer.mozilla.org/en-US/docs/Web/CSS/zoom), so the necessary version check can not be canceled.

varjolintu:
> I think the check for browser above 128 can not be canceled, because chromium below 128 handle `getBoundingClientRect/getClientRects` differently from chromium above 128. We can't guarantee that all users use browser above 128.
> 
> BTW, Firefox also supports CSS zoom property in version 126+ (https://developer.mozilla.org/en-US/docs/Web/CSS/zoom), and the latest ESR version is 128.0, so the necessary version check can not be canceled.

Usually the ESR version is enough for Firefox. And btw, you are not even checking the version 126 in the function. What I meant is that this bug is touching a very small area of users, which means the version check might not be necessary at all.

Xia0xia0Xia0:
> Usually the ESR version is enough for Firefox. And btw, you are not even checking the version 126 in the function. What I meant is that this bug is touching a very small area of users, which means the version check might not be necessary at all.

Most of Firefox's third-party forked browsers are based on the ESR version, so I chose ESR version 128 instead of CSS zoom property's minimum compatible version 126.

Some third-party chromium browsers have custom versions lower than 128, such as Supermium, centBrowser, etc. I'd like to know why we won't make the extension compatible with browsers lower than 128? The number of such user groups is not small.

varjolintu:
As I said, the bug is quite minor and I highly doubt it's worthwhile to check against the browser version. If the latest Firefox ESR already supports the CSS zoom, that is good enough. Third-party Chromium browsers have the responsibility to keep their browsers up-to-date anyway.

Xia0xia0Xia0:
I have removed the browser version check. 

And I will release self-signed extensions for third-party portable browsers below chromium version 128 to use. I wonder if that's okay?

keepassxc-browser/content/autocomplete.js
```diff
@@ -335,21 +335,25 @@ class Autocomplete {
...
         this.list.style.maxWidth = `max(${Pixels(this.input.offsetWidth)}, 600px)`;
 
+        // Get body zoom radio
+	    const zoom = kpxcUI.bodyStyle.zoom || 1;
+       
         // Calculate Y offset if menu does not fit to the bottom of the screen -> show it at the top of the input field
         const menuRect = this.container.getBoundingClientRect();
         const totalHeight = menuRect.height + rect.height;
-        const menuOffset = totalHeight + rect.y > window.self.visualViewport.height ? totalHeight : 0;
+        const menuOffset = (totalHeight + rect.y) / zoom > window.self.visualViewport.height ? totalHeight/ zoom : 0;
```
varjolintu, line 344: Code format. Probably there should be a space before `/`?

Xia0xia0Xia0: fixed.

varjolintu (approved):

PR #2346, opened 2024-09-21
Add configuration option for modifying the passkeys default group

varjolintu:
Extension side implementation of https://github.com/keepassxreboot/keepassxc/pull/11260. Adds a new input field for configuring the default group where passkeys are stored.
Screenshot 2024-09-21 at 12 03 42

Will be only enabled for KeePassXC 2.7.10 and newer.


varjolintu:
Reminder to myself that the version query should be refactored: you could just give a list for all version to compare with and results for all of those instead of making multiple `sendMessage()`s.

droidmonkey (approved):

PR #2350, opened 2024-09-28
Use non-deprecated globals with ESLint

varjolintu:
Using `true` or `false` for setting globals is deprecated: https://eslint.org/docs/latest/use/configure/language-options#specifying-globals


.eslintrc
```diff
@@ -80,106 +80,106 @@
...
-        "SORT_BY_GROUP_AND_USERNAME": true,
-        "SORT_BY_MATCHING_CREDENTIALS_SETTING": true,
-        "SORT_BY_RELEVANT_ENTRY": true,
-        "SORT_BY_TITLE": true,
-        "SORT_BY_USERNAME": true,
-        "statusResponse": true,
-        "Tests": true,
-        "tr": true,
-        "trimURL": true
+        "_called": "writable",
```
droidmonkey, line 83: Is this intentional?

droidmonkey (approved):

.eslintrc
varjolintu: It was, but now after double checking, that one is an Object that is never rewritten, even if its values are. Fixed.

PR #2359, opened 2024-10-06
Add support for draggable banners

### LINKED ISSUE #2327: Allow moving Credential Banner somewhere else (to the bottom?)
## Expected Behavior

https://www.matkahuolto.fi/ site uses notifications which flow from the top for "wrong password" and "email sent" (password recovery).

I expect to be able to move the Credential Banner, so can check the top of the page, without dismissing it.

## Current Behavior

Credential Banner is hard-coded to the top

1.
2.
3.
4.

## Debug info

KeePassXC - {VERSION}
KeePassXC-Browser - {VERSION}

varjolintu:
Adds a grab cursor when hovering on Credential or Custom Login Field banner, and allows user to move it to the bottom, or back to the top. If the credential or group selection dialog is open while dragging, its position is changed as well.

Known issues:
The dragging animation is not visible if dragging is canceled, or dragged to the same position where the banner already is. I couldn't find a reliable way to do that. The best solution was to disable the animation for good.

Banner on top (default):
Screenshot 2024-10-06 at 22 19 11

Banner on bottom (after dragging):
Screenshot 2024-10-06 at 22 19 16

Fixes #2327.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -92,6 +92,7 @@
         "AuthenticatorAssertionResponse": "readonly",
         "AuthenticatorAttestationResponse": "readonly",
         "Autocomplete": "readonly",
+        "BannerPosition": "readonly",
         "BLUE_BUTTON": "readonly",
         "bootstrap": "readonly",
         "browser": "readonly",
--- a/keepassxc-browser/background/event.js
+++ b/keepassxc-browser/background/event.js
@@ -237,6 +237,8 @@ kpxcEvent.sendBackToTabs = async function(tab, args = []) {
 kpxcEvent.messageHandlers = {
     'add_credentials': keepass.addCredentials,
     'associate': keepass.associate,
+    'banner_get_position': page.getBannerPosition,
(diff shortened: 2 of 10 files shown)


droidmonkey:
This works well but the banner showed up on the top of the page immediately after a refresh.

varjolintu:
> This works well but the banner showed up on the top of the page immediately after a refresh (ie, the banner being dragged to the bottom is not remembered)

The original idea was that this dragging is just temporary and done when it's needed. Of course we could just remember the state or add a separate option for remembering it. What do you think?

droidmonkey:
Keep the drag, its a neat shortcut. We should remember the "last used state" though.

varjolintu:
Added a setting so the position will be remembered even after browser restart.

droidmonkey (commented):
Works great, I actually prefer the banner on the bottom now!

droidmonkey (approved):

PR #2366, opened 2024-10-20
Add support for managed settings using 3rdparty policy

### LINKED ISSUE #2365: Add support for setting the settings using 3rdparty policy
Add support for [https://mozilla.github.io/policy-templates/#3rdparty](https://mozilla.github.io/policy-templates/#3rdparty).

Already referenced by [this issue](https://github.com/keepassxreboot/keepassxc-browser/issues/2319).

## Expected Behavior

Should be able to modify extension setting in Firefox using policy:

```
policies.3rdparty.Extensions."keepassxc-browser@keepassxc.org".settings = { ... }
```

## Current Behavior

Currently, this is not possible or does not work.

## Possible Solution

Some additional documentation can be found here [https://extensionworkshop.com/documentation/enterprise/enterprise-development/#how-to-add-policy](https://extensionworkshop.com/documentation/enterprise/enterprise-development/#how-to-add-policy).

Example:

```json
{
  "policies": {
    "3rdparty": {
      "Extensions": {
        "YOUR_EXTENSION_ID": {
          "STRING": "value",
          "BOOLEAN": true,
          "INTEGER": 10
        }
      }
    }
  }
}
```

How uBlock Origin has it implemented:

```json
{
  "policies": {
    "3rdparty": {
      "Extensions": {
        "uBlock0@raymondhill.net": {
          "adminSettings": {
            "selectedFilterLists": [
              "ublock-privacy",
              "ublock-badware",
              "ublock-filters",
              "user-filters"
            ]
          }
        }
      }
    }
  }
}
```

## Debug info

KeePassXC - 2.7.9
KeePassXC-Browser - 1.9.2
Browser: Firefox

varjolintu:
Adds support for distributing global settings to the extension using [3rdparty policy](https://mozilla.github.io/policy-templates/#3rdparty). Includes a new `managed_storage.json` file that describes the settings that can be set. The schema file must be formatted correctly or the extension does not start because of failed validation.

Easiest way to test this was to use `web-ext` with Firefox that had the policy JSON file added. Just made sure I deleted all storage data from the temporary extension before launching it again. That caused the `settings` to be an empty Object, and the getter for managed settings will be called.

Example `policy.json` file:
```
{
  "policies": {
    "3rdparty": {
      "Extensions": {
        "keepassxc-browser@keepassxc.org": {
          "settings": {
            "defined-custom-fields": {
              "https://example.com": {
                "fields": [
                  [
                    "/html/body/table/tbody/tr[10]/td[4]/table/tbody/tr/td/table/tbody/tr[2]/td[4]/form/table/tbody/tr[2]/td[2]/input",
                    "INPUT password form_password "
                  ]
                ],
                "password": [
                  "/html/body/table/tbody/tr[10]/td[4]/table/tbody/tr/td/table/tbody/tr[2]/td[4]/form/table/tbody/tr/td[2]/input",
                  "INPUT text form_username "
                ],
                "username": [
                  "/html/body/table/tbody/tr[10]/td[4]/table/tbody/tr/td/table/tbody/tr[2]/td[4]/form/table/tbody/tr[2]/td[2]/input",
                  "INPUT password form_password "
                ]
              }
            },
            "passkeys": true,
            "sitePreferences": [
              {
                "allowIframes": false,
                "ignore": "ignoreNothing",
                "improvedFieldDetection": false,
                "url": "https://example.com",
                "usernameOnly": true
              }
            ]
          }
        }
      }
    }
  }
}

```

Related links:
https://extensionworkshop.com/documentation/enterprise/enterprise-development/#how-to-add-policy
https://developer.chrome.com/docs/extensions/reference/manifest/storage

Fixes #2365.

### DIFF
--- a/dist/manifest_chromium.json
+++ b/dist/manifest_chromium.json
@@ -162,5 +162,8 @@
         "https://*/*",
         "http://*/*"
     ],
+    "storage": {
+      "managed_schema": "managed_storage.json"
+    },
     "default_locale": "en"
 }
--- a/dist/manifest_firefox.json
+++ b/dist/manifest_firefox.json
@@ -168,5 +168,8 @@
             "strict_min_version": "96.0"
         }
     },
+    "storage": {
+      "managed_schema": "managed_storage.json"
+    },
     "default_locale": "en"
 }
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -62,6 +62,20 @@ page.popupData = {
 page.initSettings = async function() {
     try {
         const item = await browser.storage.local.get({ 'settings': {} });
+
+        // Load managed settings if found when extension is started for the first time
+        if (Object.keys(item.settings).length === 0) {
+            try {
+                const managedSettings = await browser.storage.managed.get('settings');
+                if (managedSettings?.settings) {
+                    console.log('Managed settings found.');
+                    item.settings = managedSettings.settings;
+                }
+            } catch (err) {
+                logError('page.initSettings error: ' + err);
+            }
+        }
+
         page.settings = item.settings;
         page.settings.autoReconnect = false;
 
--- a/keepassxc-browser/managed_storage.json
+++ b/keepassxc-browser/managed_storage.json
@@ -1,0 +1,208 @@
+{
+  "type": "object",
+  "properties": {
+    "settings": {
+      "type": "object",
+      "properties": {
+        "afterFillSorting": {
+          "title": "How to sort credentials after fill. Default: sortByMatchingCredentials",
+          "type": "string"
+        },
+        "afterFillSortingTotp": {
+          "title": "How to sort credentials after fill for TOTP. Default: sortByRelevantEntry",
+          "type": "string"
+        },
+        "autoCompleteUsernames": {
+          "title": "Use Autocomplete Menu to view credentials on web pages. Default: true",
+          "type": "boolean"
+        },
+        "autoFillAndSend": {
+          "title": "Automatically fill HTTP Basic Auth entries. Default: false",
+          "type": "boolean"
+        },
+        "autoFillSingleEntry": {
+          "title": "Automatically fill single credential entries. Default: false",
+          "type": "boolean"
+        },
+        "autoFillSingleTotp": {
+          "title": "Automatically fill single TOTP. Default: false",
+          "type": "boolean"
+        },
+        "autoRetrieveCredentials": {
+          "title": "Automatically retrieve credentials. Default: true",
+          "type": "boolean"
+        },
+        "autoSubmit": {
+          "title": "Automatically submit credentials. Default: false",
+          "type": "boolean"
+        },
+        "bannerPosition": {
+          "title": "Where to position the Credential Banner. Default (top): 1",
+          "type": "integer"
+        },
+        "checkUpdateKeePassXC": {
+          "title": "When to check updates for KeePassXC. Default (never): 0",
+          "type": "integer"
+        },
+        "clearCredentialsTimeout": {
+          "title": "Clear credential info from tabs after timeout. Default (seconds): 10",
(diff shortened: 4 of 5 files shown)


mihakrumpestar:
I tried to test it.

Build using `npm run build -- --skip-translations` which builds the extension, but it won't get installed due to error:

Installer error

varjolintu:
I need to check the build script with the new file addition. For testing you do not need to build this. Just load the temporary extension directly from the sources.

mihakrumpestar:
I did read the instructions for development now. Loading as temporary add-on also does not seem to work completely properly:

Temporary add-on

The settings are then missing Passkey options, which are also absent when I export config from temporary add-on.
Note that I am using a completely new and clean profile, so I am loading this without any prior or old setting.

varjolintu:
That warning in the temporary extension page is expected, becase the default `manifest.json` is using some properties that are Chromium-only. It still works.

varjolintu:
Do you mean the passkeys option is missing even if you have enabled it?

mihakrumpestar:
> Do you mean the passkeys option is missing even if you have enabled it?

No, the Passkey section in UI is missing altogether.

image

Left is the current one I am using, right is the dev temporary one (both in different profiles).

varjolintu:
I cannot reproduce that kind of error. I tried this with both temporary profile and a newly created profile with `web-ext`.

mihakrumpestar:
I am sorry. I was trying some things and forgot I changed to master branch.

Furthermore, I tested again (on the correct branch this time) and the settings for Passkeys are indeed there, and the 3rdparty extension settings also get applied successfully. Tested on NixOS.

TLDR: It is working.

varjolintu:
Did some fixes to the `defined-custom-fields`. Those should also work now.

droidmonkey (approved):

droidmonkey:
Only comment I have: settings that are string values, we should enumerate the possible options either directly or in documentation or in wiki.

varjolintu:
Added a new wiki page for this: https://github.com/keepassxreboot/keepassxc-browser/wiki/Managed-schema

JuneStepp:
@varjolintu Do you mind explaining why you decided to have this only apply when the extension is first installed? It seems like an odd use case to only manage the initial configuration with no updates applying afterwards.

varjolintu:
@JuneStepp Is that something users really want? That would override any user-specific settings for sites etc.

JuneStepp:
> Is that something users really want? That would override any user-specific settings for sites etc. Maybe we could add a notification that asks the user if there are new settings available, and update them manually.

I don't think a notification prompt makes sense, since the point of managed settings is that something is *managed* externally. It's up to the external party to determine how things get applied.

I can only vouch for my own needs, but here's an attempt at breaking down some of the potential options for how to load settings at launch:
1. Fully overwrite with managed settings, if available.
	- What this PR would do without `if (Object.keys(item.settings).length === 0)`.
	- Great for my use case of declaratively configuring and deploying browsers across devices for personal use. 
	-  For organizations, can interfere with the user if the organization doesn't fully manage the DB(s). 
2. Load the managed settings when there are no local settings.
	- The current behavior.
	- No updating of settings so only useful as a customized default for new installations.
	- I don't really see the need for this.
3. Merge managed and user settings with user ones taking precedence.
	- The current behavior, but the defaults are kept up to date.
	- Great if the main thing an organization needs are to add custom login fields and site preferences.
4. Merge managed and user settings with managed ones taking precedence.
	- Lets the organization actually manage the settings they want to without interfering with the user beyond that.
5. Merge managed and user settings with managed ones taking precedence and being made read-only.
	- Enforced version of above.

*4* and/or *5* is what people probably expect when using policies; since this PR is already merged and released, I think keeping *2* but adding an option for *1* makes the most immediate sense. It would be a trivial change and fix my use case. Then, you could wait for someone who needs something else to say so. 
I'd be happy to do a PR for that if you want it.

PR #2367, opened 2024-10-20
Ignore elements with animations when using dynamic input field detection

### LINKED ISSUE #2363: Animations on firefox are jittery due to keepassxc-browser
## Expected Behavior
The keepassxc browser extension does not cause animations on firefox to slow down

## Current Behavior
The keepassxc browser extension causes animations on firefox to slow down

## Possible Solution
This is probably a script somewhere that's messing with the rendering or querying the html too often

1. Download and set up the keepassxc-browser extension on firefox
2. Navigate to a site with animations, for example, [Aceternity UI](https://ui.aceternity.com/components/3d-card-effect)
3. Hover over the card there and move your mouse around. You would observe the animation to be jerky

## How do I know it's due to keepassxc-browser

The animation returns back to being smooth if I only disable the keepassxc-browser extension on firefox

## Debug info

KeePassXC - 2.7.9
KeePassXC-Browser - 1.9.3
Operating system: Mac
Browser: Firefox

varjolintu:
If a page has elements that have animation CSS defined, the `MutationObserver` will trigger with each small change. Without this change `getComputedStyle()` will be called to those changes, and it can cause performance issues.

This could potentially break some sites, but we can always add exceptions for those.

Fixes #2363.

### DIFF
--- a/keepassxc-browser/content/observer-helper.js
+++ b/keepassxc-browser/content/observer-helper.js
@@ -101,7 +101,10 @@ kpxcObserverHelper.initObserver = async function() {
 // Stores mutation style to an cache array
 // If there's a single style mutation, it's safe to calculate it
 kpxcObserverHelper.cacheStyle = function(mut, styleMutations, mutationCount) {
-    if (mut.attributeName !== 'style') {
+    // Do not cache elements with animations
+    if (mut?.attributeName !== 'style'
+        || mut?.target?.style?.transform !== ''
+        || mut?.target?.style?.transformStyle !== '') {
         return;
     }
 


droidmonkey (approved):
Could this have adverse affects on login windows that animate? Fade-in/out for example?

varjolintu:
> Could this have adverse affects on login windows that animate? Fade-in/out for example?

Possibly. That's why I'm little worried. I'll test some sites where I have seen animations before I merge this.

droidmonkey:
Maybe ignore anything within a FORM element (ie, don't worry about animations if its related to a FORM)

varjolintu:
> Maybe ignore anything within a FORM element (ie, don't worry about animations if its related to a FORM). This also brings up the discussion of deferring / bundling the checks. You could accumulate a list of activations through mutation observer, overwriting those that repeat, and do the processing on the list once a second or something like that. So if an element is mutating every millisecond you only do the intense processing once per second anyway.

Yes, I've thought something like that. In this case there was just one mutation with minimal changes to the CSS styles, but the class and everything else remains the same. This kind of site is pretty good for testing.

PR #2372, opened 2024-10-27
Add support for checkboxes with Custom Login Fields

### LINKED ISSUE #2050: Auto-tick "remember me" boxes on login pages
Many pages let users tick a "remember me" box after entering login details so the session can remain active after leaving the website.

These check-boxes are usually not checked by default so one must re-log into these websites if the the extension was used to fill-in and submit the login details automatically.

This could be implemented by creating a "KPH: Remember" boolean attribute, saved as a Custom login field so user can manually select the check box. And the extension could have rules to auto-detect them on login pages and tick them by default on auto-fill.

I know KeepassXC has an advanced setting to "Never ask before accessing credentials" but it doesn't tell the website to remember the login so new logins are always required.

NB: a [similar issue](https://github.com/keepassxreboot/keepassxc-browser/issues/1474) was closed in 2021 but it was confused with the "Never ask before accessing credentials" feature.

varjolintu:
Add support for selecting checkbox elements with Custom Login Fields.

Two different input query patterns are used, where only the String Fields selection allows checkboxes. Added `:not([disabled])` to prevent selecting disabled checkboxes.
The checkbox selection overlay is fixed size, and centered to the element.
`KPH` value `true` (any case is accepted) will be used for the element's `checked` attribute. Other values are ignored.

For testing:
- Go to GitLab's login page: https://gitlab.com/users/sign_in
- Select Custom Login Fields for username and password, plus a String Field for the checkbox.
- Add `KPH: checkbox` with value `true` to the entry attributes in KeePassXC.
- Go to the login page and fill the credentials. The checkbox should be checked after fill.

Fixes #2050.

### DIFF
--- a/keepassxc-browser/content/custom-fields-banner.js
+++ b/keepassxc-browser/content/custom-fields-banner.js
@@ -6,6 +6,8 @@ const STEP_SELECT_PASSWORD = 2;
 const STEP_SELECT_TOTP = 3;
 const STEP_SELECT_STRING_FIELDS = 4;
 
+const CHECKBOX_OVERLAY_SIZE = 20;
+
 const DEFINED_CUSTOM_FIELDS = 'defined-custom-fields';
 const FIXED_FIELD_CLASS = 'kpxcDefine-fixed-field';
 const DARK_FIXED_FIELD_CLASS = 'kpxcDefine-fixed-field-dark';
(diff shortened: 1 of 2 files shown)


droidmonkey (approved):
This is awesome, can we adjust the button from "String Field" to something like "Text or Checkbox". String is a programming term.

varjolintu:
> This is awesome, can we adjust the button from "String Field" to something like "Text or Checkbox". String is a programming term.

Maybe something like "Extra fields" or "Extra input or checkbox" could do. "Text" could mean for the user that you can actually select some text elements from the page?

droidmonkey:
Extra fields is perfect, add a tool tip that describes what kind of fields are supported. Would it also be possible in a future update to automatically create the KPH entries in KPXC as well?

varjolintu:
Could be. Although it will require a credential selection where to save the settings. The feature can be used without credentials as well, so it can be tricky.

varjolintu:
I'll make a separate change for the String Fields -> Extra Fields. That has to be changed for many files and translations. I'd like to change the inner variable names as well + documentation.

PR #2375, opened 2024-10-29
Refactor compare version functions

varjolintu:
Refactors the compare version functions:
- Move the actual `compareVersion()` function to `global.js`.
- Rename `keepass.compareVersion()`  to `keepass.compareVersions()` , and allow it to take in an Array for multiple comparisons.
- Add unit tests.

### DIFF
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -946,7 +946,7 @@ keepass.updateDatabaseHashToContent = async function() {
 };
 
 // Expects an array of versions to compare
-keepass.compareVersions = function(versions, current, canBeEqual = true) {
+keepass.compareMultipleVersions = function(versions, current, canBeEqual = true) {
     if (!Array.isArray(versions)) {
         return {};
     }
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -951,7 +951,7 @@ keepass.compareVersions = function(versions, current, canBeEqual = true) {
         return {};
     }
 
-    let result = {};
+    const result = {};
     for (const version of versions) {
         result[version] = compareVersion(version, current, canBeEqual);
     }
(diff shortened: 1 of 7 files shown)


keepassxc-browser/background/keepass.js
```diff
@@ -945,25 +945,18 @@ keepass.updateDatabaseHashToContent = async function() {
...
-        return false;
-    }
-
-    // Handle beta/snapshot builds as stable version
-    const snapshot = '-snapshot';
-    const beta = '-beta';
-    if (current.endsWith(snapshot)) {
-        current = current.slice(0, -snapshot.length);
+// Expects an array of versions to compare
+keepass.compareVersions = function(versions, current, canBeEqual = true) {
```
droidmonkey, line 949: compareVersion (the global function) and keepass.compareVersions (the class function) are way too similar to each other which makes it hard to read and debug which function you actually intend to use. Recommend different naming choices.

tests/global.spec.ts
```diff
@@ -1,11 +1,27 @@
 import { test, expect } from '@playwright/test';
 import {
+    compareVersion,
     matchesWithNodeName,
     siteMatch,
     slashNeededForUrl,
     trimURL
 } from '../keepassxc-browser/common/global.js';
 
+test('Test compareVersion()', async ({ page }) => {
```
droidmonkey, line 10: Need tests for keepass.compareVersions

droidmonkey (changes requested):

tests/global.spec.ts
varjolintu: At this point functions in `keepass.js` cannot be tested due to different dependencies the file expects. One solution could be to move this function entirely to `global.js` and `keepass.js` just returns its value directly.

droidmonkey (approved):

tests/global.spec.ts
droidmonkey: Nah it's good as it is

PR #2383, opened 2024-11-10
Add support for custom password change forms

varjolintu:
Refactors `kpxcForm` for handling pages where possible change password forms are not actually `form` elements, but three different input fields, and the submit button might not even be related to the same container.

If the new `kpxcForm.initCustomForm()` detects three different password fields and there's a form submit button exception found in `kpxcSites`, the fields are stored temporarily and the extension listens for the submit button click.

Refactors `kpxcForm.onSubmit()` to have multiple smaller functions that the new  `onCustomFormSubmit()` can also use:
- `activateCredentialBanner()` handles the form field results and activates the Credential Banner. Originall this code was at the end of `onSubmit()`.
- `getUsernameValue()` returns the username value from selected input field or from retrieved credentials. Originally this code was at the middle of `onSubmit()`.

Fixes an issue mentioned in: https://github.com/keepassxreboot/keepassxc-browser/issues/1358#issuecomment-2466180599

### DIFF
--- a/keepassxc-browser/common/sites.js
+++ b/keepassxc-browser/common/sites.js
@@ -149,11 +149,11 @@ kpxcSites.expectedTOTPMaxLength = function() {
 
 /**
  * Handles a few exceptions for certain sites where form submit button is not regognized properly.
- * @param {object} form     Form element
+ * @param {object} form     Form element (optional)
  * @returns {object}        Button element
  */
 kpxcSites.formSubmitButtonExceptionFound = function(form) {
-    if (form.action.startsWith(googleUrl)) {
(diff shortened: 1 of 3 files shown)


droidmonkey (approved):
Looks good, why is "3 the magic number" on password fields?

varjolintu:
> Looks good, why is "3 the magic number" on password fields?

Three default fields: current password, new password, repeat new password. Actually, I should change that to `>= 2` just in case, because the `getNewPassword()` already works that way.

droidmonkey:
Gtg

PR #2392, opened 2024-11-26
Update to 1.9.5

varjolintu:
Updates CHANGELOG, manifest files and offline translations for 1.9.5.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,13 @@
+1.9.5 (2024-11-27)
+=========================
+- Add support for checkboxes with Custom Login Fields [#2372]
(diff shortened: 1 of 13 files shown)


droidmonkey (approved):

droidmonkey:
Is there a way to store the version string in one place and inject it where necessary during the build/load process?

varjolintu:
> Is there a way to store the version string in one place and inject it where necessary during the build/load process?

The build script already supports a version parameter that can set the string to all manifest files. The problem is that the version needs to be updated to the repository anyway. For example, using the manifest V3 file for Chrome should always keep the version number up-to-date as well. So, I'll just need to make a better automation for this.

PR #2409, opened 2024-12-15
Auto-fill will be applied if related credentials are available

wolfsilver:
If 'Auto-fill single credential entry' is selected, when multiple credentials are returned, it will also auto-fill the related credentials when available.

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -919,7 +919,7 @@
         "description": "Auto-Fill Single Entry option help text."
     },
     "optionsAutoFillRelevantCredentialHelpText": {
-        "message": "Let KeePassXC-Browser automatically fill in credentials if it receives relevant credential.",
+        "message": "Automatically fill in the relevant credential if a second fill is needed.",
         "description": "Auto-Fill Relevant Credential option help text."
     },
     "optionsAutoFillSingleEntryWarning": {
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -766,10 +766,6 @@
         "message": "Automatically fill in single-credential entries",
         "description": "Automatically fill-in single credential entry checkbox text."
     },
-    "optionsCheckboxAutoFillRelevantCredential": {
-        "message": "Automatically fill in relevant credential entries",
-        "description": "Automatically fill-in relevant credential entry checkbox text."
-    },
     "optionsCheckboxAutoFillSingleTotp": {
         "message": "Automatically fill in single TOTP entries",
         "description": "Automatically fill-in single TOTP entries checkbox text."
@@ -918,10 +914,6 @@
         "message": "Let KeePassXC-Browser automatically fill in credentials if it receives only a single entry.",
         "description": "Auto-Fill Single Entry option help text."
     },
-    "optionsAutoFillRelevantCredentialHelpText": {
-        "message": "Let KeePassXC-Browser automatically fill in credentials if it receives relevant credential.",
-        "description": "Auto-Fill Relevant Credential option help text."
-    },
     "optionsAutoFillSingleEntryWarning": {
         "message": "Warning! Using auto-fill is not safe. Use at your own risk. KeePassXC-Browser automatically tries to detect login fields. However, they may be detected incorrectly, possibly filling sensitive data to unsafe input fields.",
         "description": "Auto-Fill Single Entry warning text."
@@ -947,8 +939,8 @@
         "description": "Use monochrome toolbar icon option help text."
     },
     "optionsUseCompactModeHelpText": {
-        "message": "Use a compact version of the dropdown list to select credentials.",
-        "description": "Use compact (legacy) mode help text."
+      "message": "Use a compact version of the dropdown list to select credentials.",
+      "description": "Use compact (legacy) mode help text."
     },
     "optionsSaveDomainOnlyHelpText": {
         "message": "When saving new credentials, save only the domain instead of full URL.",
@@ -1097,7 +1089,7 @@
     "optionsColumnAllowIframes": {
         "message": "Allow Cross-Origin iframes",
         "description": "Allow iframes column text."
-    },
+  },
     "optionsColumnDelete": {
         "message": "Delete",
         "description": "Site Preferences list column title."
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -766,6 +766,10 @@
         "message": "Automatically fill in single-credential entries",
         "description": "Automatically fill-in single credential entry checkbox text."
     },
+    "optionsCheckboxAutoFillRelevantCredential": {
+        "message": "Automatically fill in relevant credential entries",
+        "description": "Automatically fill-in relevant credential entry checkbox text."
+    },
     "optionsCheckboxAutoFillSingleTotp": {
         "message": "Automatically fill in single TOTP entries",
         "description": "Automatically fill-in single TOTP entries checkbox text."
@@ -914,6 +918,10 @@
         "message": "Let KeePassXC-Browser automatically fill in credentials if it receives only a single entry.",
         "description": "Auto-Fill Single Entry option help text."
     },
+    "optionsAutoFillRelevantCredentialHelpText": {
+        "message": "Let KeePassXC-Browser automatically fill in credentials if it receives relevant credential.",
+        "description": "Auto-Fill Relevant Credential option help text."
+    },
     "optionsAutoFillSingleEntryWarning": {
         "message": "Warning! Using auto-fill is not safe. Use at your own risk. KeePassXC-Browser automatically tries to detect login fields. However, they may be detected incorrectly, possibly filling sensitive data to unsafe input fields.",
         "description": "Auto-Fill Single Entry warning text."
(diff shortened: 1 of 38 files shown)


varjolintu:
Could you describe in more details what this fix actually does? What is the error that happens, and what is that is expected to happen?
The option is clearly described for filling a single credenial. Your change fills a related credential even if there's multiple available? That's not really recommended because user cannot choose any of the others.

wolfsilver:
> Your change fills a related credential even if there's multiple available? That's not really recommended because user cannot choose any of the others.

Yes. It will auto fill with related credential, but won't submit. User still can select others.

varjolintu:
This looks at the previously submitted login id, so is this valid only for passwords? That should be already covered.

wolfsilver:
It's valid for username and passwords. 
> Use the following method to get the relevant credentials
https://github.com/keepassxreboot/keepassxc-browser/blob/aba4ffc5e1b9b13b6eb0e06a9afce2b5d71d8688/keepassxc-browser/content/form.js#L135-L142

varjolintu:
What's still unclear:
- Why is this under the same option that describes a different feature?
- Where is the `pageUuid` set originally?
- When does this fill actually takes place?

wolfsilver:
> Why is this under the same option that describes a different feature?

Configuration description needs to be updated, or set a new configuration ?
KeePassXC-Browser will automatically fill the credentials when only one entry is received, or when relevant credentials are fetched.

> Where is the pageUuid set originally?

pageUuid is updated when relevant credentials are selected, previously available functionality

> When does this fill actually takes place?

same with filling a single credential current now.

varjolintu:
If you just go to any page, this feature will not work even if there are multiple credentials, because none have been selected before, right? So there must be some kind of situation where this works, and for some it does not work.

wolfsilver:
Yes, because the pageUuid is cleared when the browser is closed. Most of the time, the first credential is not the relevant one, so it will not automatically fill the first option.

varjolintu:
So my question is: if in that case none of the credentials are selected, in what scenario this added feature works as it should and fills the first "relevant" credential?

wolfsilver:
image

This pr depends on this function

varjolintu:
Then there should be a checkbox under that option that allows filling the relevant credential automatically. It shouldn't affect to the "fill single credential" option.

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -513,6 +513,19 @@ kpxc.prepareCredentials = async function() {
 
     kpxc.initLoginPopup();
     kpxc.initAutocomplete();
+
+    if (kpxc.settings.autoFillSingleEntry) {
+        const pageUuid = await sendMessage('page_get_login_id');
+        if (pageUuid) {
+            kpxc.credentials.some(c => {
```
droidmonkey, line 520: This is very wrong, you are using some,  a collector function,  to iterate. You need to use a for loop instead so that your return call actually ends the loop.

droidmonkey (changes requested):

keepassxc-browser/content/keepassxc-browser.js
wolfsilver: > This is very wrong, you are using some, a collector function, to iterate. You need to use a for loop instead so that your return call actually ends the loop.

According to mdn, does it immediately end loop if some return true ?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some#return_value

wolfsilver:
> Then there should be a checkbox under that option that allows filling the relevant credential automatically. It shouldn't affect to the "fill single credential" option.


Add new option to control the filling of relevant credential.

Use AI to translate into other languages. Needs to be recalibrated.

image

varjolintu:
> > Then there should be a checkbox under that option that allows filling the relevant credential automatically. It shouldn't affect to the "fill single credential" option.
> 
> Add new option to control the filling of relevant credential.
> 
> Use AI to translate into other languages. Needs to be recalibrated.
> 

As I said in my previous message, that option is related to the combobox selection in another section. Its place is not here.

Use AI to translate..? What?

keepassxc-browser/content/keepassxc-browser.js
droidmonkey: Absolutely not. `some` means collect certain elements that match certain criteria. Returning true means "include this considered element".

droidmonkey: Interesting, so this is a tester not a collector. Sigh...

wolfsilver:
> > > Then there should be a checkbox under that option that allows filling the relevant credential automatically. It shouldn't affect to the "fill single credential" option.
> > 
> > 
> > Add new option to control the filling of relevant credential.
> > Use AI to translate into other languages. Needs to be recalibrated.
> 
> As I said in my previous message, that option is related to the combobox selection in another section. Its place is not here.
> 
> Use AI to translate..? What? Do not include any translations to PR's. Those are handled via Transifex.

Do you mean to move under "Sort credentials after fill by" ?
And translations reverted.

image

wolfsilver:
> > Do you mean to move under "Sort credentials after fill by" ? And translations reverted.
> 
> Yes, but the select element's help text should be left under the selection. The new option should be below as its own.

OK now?
image

varjolintu:
My question still is that where does this feature works if I enable the option? The internal login ID is only set _after_ filling the credentials, and this option assumes the most relevant credential is automatically filled. If it hasn't been filled before, the sorting hasn't been even done at the first place.

wolfsilver:
The sole purpose is to avoid re-selecting from the credential list if the previously used credential is chosen. When there are multiple credentials, users previously needed to select one. Now, if the last-used credential is being used, no selection is needed. However, if a different credential is required, users will still need to make a selection.

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -513,6 +513,19 @@ kpxc.prepareCredentials = async function() {
 
     kpxc.initLoginPopup();
     kpxc.initAutocomplete();
+
+    if (kpxc.settings.autoFillRelevantCredential) {
+        const pageUuid = await sendMessage('page_get_login_id');
+        if (pageUuid) {
+            kpxc.credentials.some(c => {
+                if (c.uuid === pageUuid) {
```
varjolintu, line 521: You could just replace this with a `kpxc.credentials.find()`, and if a credential is found you get the correct combination for the fill.
Is there some reason why the last combination (or -1 as index if there's no combinations) is returned? There could be multiple combinations available.

wolfsilver:
@varjolintu Should I submit translations in en/messages.json?

image

varjolintu:
> @varjolintu Should I submit translations in en/messages.json?

Yes. That's the only place where new strings/messages are needed.

keepassxc-browser/content/keepassxc-browser.js
wolfsilver: changed `some` to `kpxc.credentials.find`

varjolintu:
Is there some site I can reliably test this feature?

wolfsilver:
> Is there some site I can reliably test this feature?

google.com?
You can add more than one test data, then refresh the page, select "xxx", then refresh the page again and "xxx" will be automatically filled.

keepassxc-browser/_locales/en/messages.json
```diff
@@ -914,6 +918,10 @@
         "message": "Let KeePassXC-Browser automatically fill in credentials if it receives only a single entry.",
         "description": "Auto-Fill Single Entry option help text."
     },
+    "optionsAutoFillRelevantCredentialHelpText": {
+        "message": "Let KeePassXC-Browser automatically fill in credentials if it receives relevant credential.",
```
varjolintu, line 922: I think a better wording for this is: "Automatically fill in the relevant credential if a second fill is needed."

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -513,6 +513,17 @@ kpxc.prepareCredentials = async function() {
 
     kpxc.initLoginPopup();
     kpxc.initAutocomplete();
+
+    if (kpxc.settings.autoFillRelevantCredential) {
+        const pageUuid = await sendMessage('page_get_login_id');
+        if (pageUuid) {
+            const relevantCredential = kpxc.credentials.find(c => c.uuid === pageUuid);
+            if (relevantCredential) {
+                const index = kpxc.combinations.length - 1;
```
varjolintu, line 522: I'd probably change this to something like:
```
const relevantCredential = kpxc.credentials.find(c => c.uuid === pageUuid);
const combination = kpxc.combinations?.at(-1);
if (relevantCredential && combination) ...
```

wolfsilver: fixed in [49c25f7](https://github.com/keepassxreboot/keepassxc-browser/pull/2409/commits/49c25f7711e15cca63a4ced3a8127ff80fae4875)

keepassxc-browser/_locales/en/messages.json
wolfsilver: fixed in [49c25f7](https://github.com/keepassxreboot/keepassxc-browser/pull/2409/commits/49c25f7711e15cca63a4ced3a8127ff80fae4875)

varjolintu (approved):

keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -513,6 +513,17 @@ kpxc.prepareCredentials = async function() {
 
     kpxc.initLoginPopup();
     kpxc.initAutocomplete();
+
+    if (kpxc.settings.autoFillRelevantCredential) {
+        const pageUuid = await sendMessage('page_get_login_id');
+        if (pageUuid) {
+            const relevantCredential = kpxc.credentials.find(c => c.uuid === pageUuid);
+            const combination = kpxc.combinations?.at(-1);
```
droidmonkey, line 521: Does this return the last combination pair discovered on the page?

droidmonkey (approved):

keepassxc-browser/content/keepassxc-browser.js
varjolintu: I think this is here because if there are two combinations, the latter is the one with the password field. Still, it wouldn't hurt to actually check that the combination has that field.

PR #2421, opened 2024-12-24
Add all_urls permission to manifest files

varjolintu:
Google wants this permission to be added. The permission is not listed in https://developer.chrome.com/docs/extensions/reference/permissions-list but handled under `host_permissions`: https://developer.chrome.com/docs/extensions/develop/concepts/match-patterns#special. They didn't provide any detailed info where this should be set, so this is the best bet.

## Type of change
- ✅ Maintenance

### DIFF
--- a/dist/manifest_chromium.json
+++ b/dist/manifest_chromium.json
@@ -159,6 +159,7 @@
         "extension_pages": "script-src 'self'"
     },
     "host_permissions": [
+        "<all_urls>",
         "https://*/*",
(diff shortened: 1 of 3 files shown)


droidmonkey:
Since this is a chrome thing, should you put this in the Firefox manifest?

varjolintu:
> Since this is a chrome thing, do you really need to put this in the Firefox manifest?

Probably not.

PR #2422, opened 2024-12-24
Modifications to Redetect Fields

varjolintu:
Noticed that the `get_status` request after Redetect Fields was not an internal call, but it requested status from KeePassXC that is not even needed here. It is enough to update the `page` variables and refresh the popup with new data directly.

Other improvements:
- Replace direct calls to `page` with proper functions.
- Only set username field detected when the content script URL matches the page URL. Otherwise any iframes (from Google etc.) would set the username field detected variable back to `false`.

## Testing strategy
Manually. For example Patreon's login page can be used when Predefined Sites is disabled in the extension settings.

## Type of change

### DIFF
--- a/keepassxc-browser/popups/popup.js
+++ b/keepassxc-browser/popups/popup.js
@@ -125,7 +125,8 @@ const sendMessageToTab = async function(message) {
         }
 
         statusResponse(await browser.runtime.sendMessage({
-            action: 'get_status'
+            action: 'get_status',
+            args: [ true ]
         }));
     });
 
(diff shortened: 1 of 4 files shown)


keepassxc-browser/popups/popup.js
```diff
@@ -125,7 +125,8 @@ const sendMessageToTab = async function(message) {
         }
 
         statusResponse(await browser.runtime.sendMessage({
-            action: 'get_status'
+            action: 'get_status',
+            args: [ true ]
```
varjolintu, line 129: This boolean changes the request to an internal call without sending messages to KeePassXC.

droidmonkey:
What if the login dialog is in an iframe with a different url and I need to set the username only detection?

varjolintu:
> What if the login dialog is in an iframe with a different url and I need to set the username only detection?

It's valid concern that I also thought about. I need to find a site where I can test this.

PR #2426, opened 2025-01-02
Store and retrieve loginId from local storage for better session mana…

wolfsilver:
Use persistent caching to save the `loginId` for a better experience.
Currently, the `loginId` only exists in the current session and is cleared when the browser is closed.

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -252,12 +252,18 @@ page.getLoginId = async function(tab) {
     if (currentTab && !currentTab.loginId && currentTab.credentials.length === 1) {
         return currentTab.credentials[0].uuid;
     }
+    if (currentTab && currentTab.loginId) {
+        return currentTab.loginId;
+    }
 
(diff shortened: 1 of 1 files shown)


varjolintu:
In my opinion there's no need for this. The login ID is a temporary variable and only needed for storing the value between pages that separate filling the username and password. Meaning, the ID is stored after username fill to find the relevant password from the correct entry.

wolfsilver:
> In my opinion there's no need for this. The login ID is a temporary variable and only needed for storing the value between pages that separate filling the username and password. Meaning, the ID is stored after username fill to find the relevant password from the correct entry.

The login ID is also used to set the relevant password.
It will auto automatically fill in the relevant password after restarting browser.

https://github.com/keepassxreboot/keepassxc-browser/pull/2409#issuecomment-2543646959

varjolintu:
I don't think it's a possible scenario that you are in a password input page after a browser restart. In general, we don't want to save any credential related data to localstorage.

Issue #1582, opened 2022-03-21
Autofill with multiple logins


ep0ns:
There is an autofill setting, when only one entry is available which should be improved to work when multiple logins available

## Expected Behavior
Implement intelligent autofill working with multiple logins for the same page.

## Current Behavior
Autofill is only working when the option "Autofill when one entry is available" is set and only one entry is available.

## Possible Solution
Typically Keepass entries have a history to see when the  entry has been used the last time. This information could be used to autofill the login even when multiple logins are available for the same site.  In case the history is not available it should be done with A-Z sorting or optimally a "priority or preferred" flag which can be set for the entry which overwrites the history setting.

varjolintu:
This would need some custom data history feature to KeePassXC side, and I'm not quite sure how to do that. There should be a separate table for storing each entry with a flag that it has been previously selected. Even if this could be possible to do, entries with multiple URL's could flag more than one entry as "previously selected" when this feature wouldn't work at all, unless the data is stored with the entry and the URL it was being used.

Thinking the complexity of this, I highly doubt it will be implemented.

ep0ns:
Ye but it would make the autofill so much more convenient. A simple option would be make another option "Autofill with multiple entries" and autofill according to sort order of Name (not username) of the entry.

Example: 
Entry 1)
name "google.com - A",
user   "D"
Entry 2)
name "google.com - B"
user "A"

In this user "D" would be autofilled.

varjolintu:
If Auto-Fill is always used with the same credentials for site with multiple credentials, doesn't that make the other credentials unneeded? In this case the most simplest solution would be just disable those other credentials which are no longer used anyway.

ep0ns:
In my case im working on a page where i need to login multiple times per day. In 90% I use login X but in the other 10% I use login Y or Z. So it would be nice to have autofill set to login X per default without manually choosing it.

varjolintu:
Then how would you select login Y or Z if Auto-Fill is always set to X? You would have to disable the Auto-Fill totally every tenth time?

ep0ns:
You can select other entries if you click the user field or when you click the Keepass XC Icon which is showed next to the user field.

varjolintu:
The easiet solution for this would be that the extension remembers the entry UUID per site and keeps a list of it. Anything for KeePassXC side would be too much.
Maybe an extra option to Site Preferences.

Deagan:
[...]

PR #2504, opened 2025-04-02
Autofill preferred user on multiple matches

tmomp:
This PR adds the feature to autofill credentials of a preferred user if there are multiple matches for a specific site.
The selection menu for users still works in that case.

This helps me a lot in scenarios where I usually login with a default user but in rare cases have to choose a different user which I would otherwise have to search for in the keepassxc app manually.

The preferred user can be configured in the site preferences settings.

(!) The implementation is only done for form logins.
(!) Translations for the options label are still missing

### DIFF
--- a/keepassxc-browser/_locales/en/messages.json
+++ b/keepassxc-browser/_locales/en/messages.json
@@ -1098,6 +1098,10 @@
         "message": "Ignore",
         "description": "Site Preferences list column title."
     },
+    "optionsColumnPreferredUser": {
+        "message": "Preferred username",
+        "description": "Preferred username if there are multiple matches."
+    },
     "optionsColumnUsernameOnly": {
         "message": "Username-only Detection",
         "description": "Site Preferences list column title."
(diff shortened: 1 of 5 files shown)


varjolintu:
FYI: Credentials are usually sorted by "Group and title", so you could achieve the same with just changing your entry title.

tmomp:
But that wouldn't fill in the first entry automatically. Or is this already possible somehow?

varjolintu:
Nope, there's no auto-submit when there are multiple credentials.

droidmonkey:
I am torn on this being an extension setting vs a field pushed from keepassxc (after allowing credentials). We have the option to "only return best matching credentials" that is a global option but really should be a database option and even a preferential thing per site "collection". Perhaps another way to do this is to have a checkbox on the browser page of the entry editing to mark that entry as "preferred" which is passed along to the extension with everything else for presentation and filling purposes.

varjolintu:
> I am torn on this being an extension setting vs a field pushed from keepassxc (after allowing credentials). We have the option to "only return best matching credentials" that is a global option but really should be a database option and even a preferential thing per site "collection". Perhaps another way to do this is to have a checkbox on the browser page of the entry editing to mark that entry as "preferred" which is passed along to the extension with everything else for presentation and filling purposes.

I would also suggest something like this. The extension side is getting bloated with tons of different settings.

varjolintu:
Related: https://github.com/keepassxreboot/keepassxc-browser/issues/1582

PR #2510, opened 2025-04-12
Fixes for page object checks and context menu item creation

### LINKED ISSUE #2507: Browser extension not working with Edge
I am unable to get the Browser extension in Edge to connect. (It was working correctly last week. The location of my database has changed as part of a corporate IT Migration)

I have tried:-
remove extension - restart browser - disable integration - reboot - start keepassxc - enable browser integration - install extension
When i click on connect database, nothing happens.

I get error:-
[Error page.js:148] KeePassXC-Browser - Cannot send activated_tab message: Could not establish connection. Receiving end does not exist.
global.js:168 [Error keepass.js:313] KeePassXC-Browser - 5: Cannot connect to KeePassXC. Check that browser integration is enabled in KeePassXC settings.

varjolintu:
Sometimes `page.settings` is not loaded before other scripts are doing their job. This can happen on browser launch, after browser update etc. Additional checks are added.

There's some strange delay after browser start before the toolbar icons update. The whole reason for this behavior is kind of weird. When browser starts for the first time, the scripts are initialized in a different order. If the extension is reloaded using `edge://extensions` the order is again the correct one. These changes should ensure the order is kept intact.

Fixes #2507.

## Testing strategy
Manually by restarting the browser and inspecting the service worker, plus reloading the extension on-the-fly. Using Microsoft Edge.

## Type of change

### DIFF
--- a/keepassxc-browser/background/browserAction.js
+++ b/keepassxc-browser/background/browserAction.js
@@ -11,7 +11,7 @@ browserAction.show = async function(tab, popupData) {
         path: await browserAction.generateIconName(popupData.iconType)
     });
 
-    if (popupData.popup) {
+    if (popupData.popup && tab?.id) {
         browserActionWrapper.setPopup({
             tabId: tab.id,
             popup: `popups/${popupData.popup}.html`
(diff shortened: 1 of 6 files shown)


droidmonkey (approved):

droidmonkey:
I really don't like having all those tab checks sprinkled everywhere. Is there some initial place we can put the basic check to avoid initializing too early that covers the majority of the functions?

varjolintu:
> I really don't like having all those tab checks sprinkled everywhere. Is there some initial place we can put the basic check to avoid initializing too early that covers the majority of the functions?

I can try doing some improvements, yes.. It's not very pretty now.

varjolintu:
Found another thing: while the background scripts are loading (something) at the browser start, if you go to a some login page during that phase, the content scripts cannot connect to background scripts at all.

varjolintu:
Seems the `browser.manager.get()` can take seconds during browser start. That's the main reason for all the problems here.

varjolintu:
I think I finally got it. Changing the request to a callback with Chromium based browsers eliminates the delay. Can you try this again? Thanks.

PR #2521, opened 2025-04-18
Add support for autocomplete=username

### LINKED ISSUE #2436: Support autocomplete=username on single input fields
I'm building a login form that asks for the email address first. After this form is submitted (with a standard HTML post form), the user is sent to a new page where they can enter their password.

Is there some markup I need to do to get keepassxc to detect the standalone email address as the username field? I tried `autocomplete="username"`, but it seems like keepasscx only does something if there's also a password field on the page.

varjolintu:
Currently we only support single input fields if those are explicitly allowed via `Username-Only Detection`. Those fields should be also allowed when `autocomplete="username"` attribute is included, or if it contains `email`. Some sites use multiple values here, separated by spaces.

Fixes #2436.

## Testing strategy
Manually. Disable Predefined Sites in the extension settings and go to https://www.adobe.com/ and login.

## Type of change

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -46,7 +46,11 @@ kpxcFields.getAllCombinations = async function(inputs) {
         }
     }
 
-    if (kpxc.singleInputEnabledForPage && combinations.length === 0 && usernameField) {
+    // Allow single input if autocomplete="username" is present, or Username-Only Detection is enabled for the page
+    const autoComplete = usernameField?.getLowerCaseAttribute('autocomplete');
+    if (combinations.length === 0 && usernameField &&
+        (kpxc.singleInputEnabledForPage || autoComplete?.includes('username') || autoComplete?.includes('email'))
+    ) {
         const combination = {
             username: usernameField,
             password: null,


keepassxc-browser/content/fields.js
```diff
@@ -46,7 +46,11 @@ kpxcFields.getAllCombinations = async function(inputs) {
         }
     }
 
-    if (kpxc.singleInputEnabledForPage && combinations.length === 0 && usernameField) {
+    // Allow single input if autocomplete="username" is present, or Username-Only Detection is enabled for the page
+    const autoComplete = usernameField?.getLowerCaseAttribute('autocomplete');
+    if (combinations.length === 0 && usernameField &&
+        (kpxc.singleInputEnabledForPage || autoComplete?.includes('username') || autoComplete?.includes('email'))
```
droidmonkey, line 52: Do we need to lowercase the autoComplete value?

droidmonkey (approved):

keepassxc-browser/content/fields.js
varjolintu: Yes. Websites can use `Autocomplete`,  `AUTOCOMPLETE` etc. because they're not interested following the standards.

varjolintu: Actually that is what the function is doing, not lowercasing the attribute name itself.. :)

evert:
Thank you!

PR #2524, opened 2025-04-21
Add support for opening Firefox shortcut settings page

### LINKED ISSUE #389: Drop hot key customization options
You currently provide a custom way to adjust hot keys of your extension.

Starting with Firefox 66 [Firefox will provide a built-in way of doing so](https://bugzilla.mozilla.org/show_bug.cgi?id=1303384) ([German article](https://www.soeren-hentzschel.at/firefox/tastatur-kommandos-von-erweiterungen-koennen-ab-firefox-66-geaendert-werden/)) and thus you can IMHO drop your own customization options in `/options/shortcuts.html`.

AFAIK other browsers like Chrome/ium already support that customization as a built-in feature too, so there is no need to reimplement that in your extension anymore.

varjolintu:
Firefox 137 and newer finally support opening extension shortcut pages: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/commands/openShortcutSettings.

The internal shortcuts page must be removed when support for `openShortcutSettings()` hits the Firefox ESR.

Fixes #389.

## Testing strategy
Manually.

## Type of change

### DIFF
--- a/keepassxc-browser/options/options.js
+++ b/keepassxc-browser/options/options.js
@@ -200,9 +200,21 @@ options.initGeneralSettings = async function() {
     });
 
     $('#configureCommands').addEventListener('click', function() {
+        if (isFirefox()) {
+            if (typeof(browser.commands.openShortcutSettings) === 'function') {
+                browser.commands.openShortcutSettings();
+            } else {
+                // TODO: Remove internal shortcuts page after Firefox ESR has support for openShortcutSettings()
+                browser.tabs.create({
+                    url: browser.runtime.getURL('options/shortcuts.html')
+                });
+            }
+            return;
+        }
+
         const scheme = isEdge() ? 'edge' : 'chrome';
         browser.tabs.create({
-            url: isFirefox() ? browser.runtime.getURL('options/shortcuts.html') : `${scheme}://extensions/shortcuts`
+            url: `${scheme}://extensions/shortcuts`
         });
     });
 


keepassxc-browser/options/options.js
```diff
@@ -200,9 +200,22 @@ options.initGeneralSettings = async function() {
     });
 
     $('#configureCommands').addEventListener('click', function() {
+        if (isFirefox()) {
+            // Opening extension shortcuts is only supported with Firefox 137 and newer
+            if (getBrowserMajorVersion() >= 137) {
+                browser.commands.openShortcutSettings();
```
droidmonkey, line 206: Is there some way to test if `browser.commands` has the `openShortcutSettings` function? I don't necessarily like testing for version number, especially if Firefox variants don't implement it or have different version schemes.

droidmonkey (changes requested):

keepassxc-browser/options/options.js
varjolintu: True, let's do that instead.

droidmonkey (approved):

keepassxc-browser/options/options.js
varjolintu: It is needed. Older Firefox versions (for example Firefox ESR that's quite common) doesn't have this feature yet.

Issue #2588, opened 2025-06-24
Better handling of extension permissions and practices/communication involving them


worstje:
First of all: I'm not a web/browser developer, but I am a developer who is familiar with security-related topics in other areas. In this context I'm just a tech-savvy user at best, so please consider this a critique and/or desperate cry from me as I try to answer my own concerns and am left wanting in regards to the answers I find.

---

Today, my browser (Vivaldi) just updated to version 1.9.9 of the browser plugin, and in the process broke my workflow because it got disabled. Apparently it is because the permissions changed?

A quick glance confirms: it went from four bullet points worth of permissions to five lines worth of permissions when trying to re-enable it.

This is a considerable red flag to me: my plugin was working fine beforehand, so why did the permissions change? Did someone take over the repository? Is there a new feature? I have no clue: it does not list a changelog in the webstore when I visit there, and since I don't pay attention much, I have no clue whether the owner changed, whether I may be batching a few different versions worth of updates or anything else that might explain this behavior.

I see no other choice but to dig into this matter to make sure this is on the up and up since leaking dozens of passwords due to some sneaky backdoor would be by far the worse alternative.

🚩**Red flag 1: No mention of a change of permissions on the platform I downloaded the plugin from.**

Okay. The webstore page says to check the release notes in this github repository. I dutifully check the release notes for v1.9.9 and start checking every pull request for changes to `manifest.json` since this is where this data lives according to a quick search online. _(Note: since writing most of this issue I have come to realize there also exist versions in `dist/` specific to the webstores, but for simplicities sake I'll only refer to the single file since I see no relevant differences in regards to the concerns listed in this issue.)_

🚩**Red flag 2: The `CHANGELOG.md` file does not clearly communicate security-related changes that an end-user may be interested in.**

PR #2570 adds something to do with a privacy setting, but `CHANGELOG.md` documents it as _'Add Getting Started page when extension is installed for the first time [#2570]'_. Weird. Awkward commit naming, but I found the commit messing with `manifest.json`.

But why is the plugin responsible for setting itself as the default password manager? This seems like a strange thing to file away under 'privacy', but a quick check using an AI says it is probably okay. Is it? I lack the expertise to be sure, but in my brain, this is the sort of global setting a user should configure rather than have the extension configure itself. Could this be a change that prepares for some other trojan-horse change down the line? `README.md` is very succinct about why this plugin needs certain permissions.

[...]

PR #2595, opened 2025-07-01
Move privacy to optional permissions

varjolintu:
Moves the new `privacy` permission to `optional_permissions` list. This needs a new setting that is in sync with the permission, otherwise the page tries to use the permission every time on load and would trigger the prompt when querying the default password manager setting from browser. The permission prompt is shown only when trying to change the option for the first time, or denying it and trying again.

I didn't add this option to managed settings.

Related issue thread: #2588

## Testing strategy
Manually. Using Firefox with `web-ext` resets the permissions every time.

## Type of change
- ✅ Breaking change (causes existing functionality to change)

### DIFF
--- a/dist/manifest_chromium.json
+++ b/dist/manifest_chromium.json
@@ -162,13 +162,15 @@
         "nativeMessaging",
         "notifications",
         "offscreen",
-        "privacy",
         "storage",
         "tabs",
         "webNavigation",
         "webRequest",
         "webRequestAuthProvider",
         "webRequestBlocking"
+    ],
+    "optional_permissions": [
+        "privacy"
     ],
     "content_security_policy": {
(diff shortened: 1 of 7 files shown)


droidmonkey:
I think this is an Edge thing because Edge doesn't have a built-in password manager, it uses the Microsoft account for that.

droidmonkey (approved):

varjolintu:
> I think this is an Edge thing because Edge doesn't have a built-in password manager, it uses the Microsoft account for that.

For me the relevant setting is found at edge://wallet/settings#settings-passwords-section

varjolintu:
Now I suddenly could reproduce the problem you mentioned. No idea why it didn't happen earlier. Making another fix.

PR #2635, opened 2025-08-16
Update observer-helper.js to handle mutations with multiple elements

mihai-vlc:
On dynamic pages, a single MutationRecord can contain multiple elements in the addNodes / removedNodes properties.

https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord

## Testing strategy
Example form where username/password field detection was not working as expected (even after choosing the Custom Login Fields): https://account.demandware.com

## Type of change

### DIFF
--- a/keepassxc-browser/content/observer-helper.js
+++ b/keepassxc-browser/content/observer-helper.js
@@ -80,7 +80,8 @@ kpxcObserverHelper.initObserver = async function() {
                     mut.addedNodes.forEach(function (node) {
                         kpxcObserverHelper.handleObserverAdd(node);
                     });
-                } else if (mut.removedNodes.length > 0) {
+                }
+                if (mut.removedNodes.length > 0) {
                     mut.removedNodes.forEach(function (node) {
                         kpxcObserverHelper.handleObserverRemove(node);
                     });
--- a/keepassxc-browser/content/observer-helper.js
+++ b/keepassxc-browser/content/observer-helper.js
@@ -76,16 +76,12 @@ kpxcObserverHelper.initObserver = async function() {
             kpxcObserverHelper.cacheStyle(mut, styleMutations, mutations.length);
 
             if (mut.type === 'childList') {
-                if (mut.addedNodes.length > 0) {
-                    mut.addedNodes.forEach(function (node) {
-                        kpxcObserverHelper.handleObserverAdd(node);
-                    });
-                }
-                if (mut.removedNodes.length > 0) {
-                    mut.removedNodes.forEach(function (node) {
-                        kpxcObserverHelper.handleObserverRemove(node);
-                    });
-                }
+                mut.addedNodes.forEach(function (node) {
+                    kpxcObserverHelper.handleObserverAdd(node);
+                });
+                mut.removedNodes.forEach(function (node) {
+                    kpxcObserverHelper.handleObserverRemove(node);
+                });
             } else if (mut.type === 'attributes' && (mut.attributeName === 'class' || mut.attributeName === 'style')) {
                 // Only accept targets with forms
                 const forms = matchesWithNodeName(mut.target, 'FORM')
--- a/keepassxc-browser/content/observer-helper.js
+++ b/keepassxc-browser/content/observer-helper.js
@@ -77,9 +77,13 @@ kpxcObserverHelper.initObserver = async function() {
 
             if (mut.type === 'childList') {
                 if (mut.addedNodes.length > 0) {
-                    kpxcObserverHelper.handleObserverAdd(mut.addedNodes[0]);
+                    mut.addedNodes.forEach(function (node) {
+                        kpxcObserverHelper.handleObserverAdd(node);
+                    });
                 } else if (mut.removedNodes.length > 0) {
-                    kpxcObserverHelper.handleObserverRemove(mut.removedNodes[0]);
+                    mut.removedNodes.forEach(function (node) {
+                        kpxcObserverHelper.handleObserverRemove(node);
+                    });
                 }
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/observer-helper.js
```diff
@@ -77,9 +77,13 @@ kpxcObserverHelper.initObserver = async function() {
 
             if (mut.type === 'childList') {
                 if (mut.addedNodes.length > 0) {
-                    kpxcObserverHelper.handleObserverAdd(mut.addedNodes[0]);
+                    mut.addedNodes.forEach(function (node) {
+                        kpxcObserverHelper.handleObserverAdd(node);
+                    });
                 } else if (mut.removedNodes.length > 0) {
```
droidmonkey: Just noticed this, what if we have both added and removed nodes? They will not be removed.

droidmonkey (changes requested):

keepassxc-browser/content/observer-helper.js
mihai-vlc: adjusted it to check for both

keepassxc-browser/content/observer-helper.js
```diff
@@ -77,9 +77,14 @@ kpxcObserverHelper.initObserver = async function() {
...
-                } else if (mut.removedNodes.length > 0) {
-                    kpxcObserverHelper.handleObserverRemove(mut.removedNodes[0]);
+                    mut.addedNodes.forEach(function (node) {
+                        kpxcObserverHelper.handleObserverAdd(node);
+                    });
+                }
+                if (mut.removedNodes.length > 0) {
+                    mut.removedNodes.forEach(function (node) {
+                        kpxcObserverHelper.handleObserverRemove(node);
+                    });
```
droidmonkey: Since we are now using forEach, we do not need to check if the length is > 0. Drop the if clauses and then this is gtg.

droidmonkey (changes requested):

droidmonkey (approved):

droidmonkey:
@varjolintu should give your look over before merging

varjolintu:
I tested the example site and couldn't reproduce the original problem. While inspecting the MutationObserver, `addedNodes` and `removedNodes` arrays are both empty all the time. The input fields are also recognized normally. How did you actually tested this?

mihai-vlc:
@varjolintu 
it looks like they released an update on the login form
2 weeks ago the username/password were rendered client side
today they are rendered on the server, so the observer doesn't need to find any new elements after the initial page load

A simple example of multiple nodes being added/removed at the same time would be something like this:
```
var observerConfig = {
    subtree: true,
    attributes: true,
    childList: true,
    characterData: true,
    attributeFilter: [ 'style', 'class' ]
};

var observer = new MutationObserver(function(mutations, obs) {
  for (const mut of mutations) {
    if (mut.type === 'childList') {
      console.log("added", mut.addedNodes.length, mut.addedNodes);
      console.log("removed", mut.removedNodes.length, mut.removedNodes);
    }
  }
});

observer.observe(document.body, observerConfig)

setTimeout(function () {
  document.body.innerHTML = `
    <label>username</label>
    <input type="text" name="username" />
  `;
}, 1000);

setTimeout(function () {
  document.body.innerHTML = `
    <label>password</label>
    <input type="password" name="password" />
  `;
}, 3000);
```

droidmonkey:
Here is a fully working demo site:

<details>
<summary>Python Server:</summary>

```python
#!/usr/bin/env python3
import os
from http.server import SimpleHTTPRequestHandler, ThreadingHTTPServer
from pathlib import Path

PORT = 8000

# Serve files relative to this script's directory
BASE_DIR = Path(__file__).resolve().parent
os.chdir(BASE_DIR)

class TestFileHandler(SimpleHTTPRequestHandler):
    # Map "/" to "/test.html"
    def do_GET(self):
        if self.path in ("/", "/index.html"):
            self.path = "/test.html"
        return super().do_GET()

if __name__ == "__main__":
    # Optional: avoid "address already in use" after quick restarts
    ThreadingHTTPServer.allow_reuse_address = True
    with ThreadingHTTPServer(("127.0.0.1", PORT), TestFileHandler) as httpd:
        print(f"Serving test.html at http://localhost:{PORT}/")
        print(f"Also available at http://localhost:{PORT}/test.html")
        httpd.serve_forever()
```
</details>

<details>
<summary>HTML file (test.html)</summary>

```html
<html lang="en">

<head>
    <title>Test Page</title>
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <meta charset="utf-8">
    <script>
        document.addEventListener('DOMContentLoaded', function () {
            var observerConfig = {
                subtree: true,
                attributes: true,
                childList: true,
                characterData: true,
                attributeFilter: ['style', 'class']
            };

            var observer = new MutationObserver(function (mutations, obs) {
                for (const mut of mutations) {
                    if (mut.type === 'childList') {
                        console.log("added", mut.addedNodes.length, mut.addedNodes);
                        console.log("removed", mut.removedNodes.length, mut.removedNodes);
                    }
                }
            });

            observer.observe(document.body, observerConfig);

            setTimeout(function () {
                document.body.firstElementChild.innerHTML = `
    <label>username</label>
    <input type="text" name="username" />
  `;
            }, 1000);

            setTimeout(function () {
                document.body.firstElementChild.innerHTML = `
    <label>password</label>
    <input type="password" name="password" />
  `;
            }, 3000);
        });
    </script>
</head>

<body>
    <div></div>
</body>

</html>
```
</details>

Using this PR on this demo site results in much improved behavior over develop branch.

varjolintu:
Thank you for the fix :)

PR #2649, opened 2025-08-23
Add special error message for untrusted iframes

droidmonkey:
image

## Testing strategy
Tested on multiple types of forms

## Type of change

### DIFF
--- a/keepassxc-browser/content/fill.js
+++ b/keepassxc-browser/content/fill.js
@@ -217,7 +217,7 @@ kpxcFill.fillFromUsernameIcon = async function(combination) {
  */
 kpxcFill.fillInCredentials = async function(combination, predefinedUsername, uuid, passOnly = false) {
     if (kpxc.credentials.length === 0) {
-        kpxcUI.createNotification('error', tr('credentialsNoLoginsFound'));
+        showErrorNotification(`${tr('credentialsNoLoginsFound')} ${document.location.origin}`);
         return;
     }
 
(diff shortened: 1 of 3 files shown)


keepassxc-browser/content/keepassxc-browser.js
```diff
@@ -646,7 +646,7 @@ kpxc.rememberCredentialsFromContextMenu = async function() {
 // Credential Banner can force the retrieval for reloading new/modified credentials.
 kpxc.retrieveCredentials = async function(force = false) {
     if (!await isIframeAllowed()) {
-        return [];
+        return;
```
varjolintu, line 649: Not sure why this is changed. We prefer returning an empty list if there are no credentials.

keepassxc-browser/content/fill.js
```diff
@@ -358,6 +358,9 @@ const showErrorNotification = async function(errorMessage, notificationType = 'e
     const connectedDatabase = await sendMessage('get_connected_database');
     if (!connectedDatabase?.identifier) {
         kpxcUI.createNotification('error', tr('errorCurrentDatabaseNotConnected'));
+    } else if (!await isIframeAllowed()) {
+        // Special error if we are blocking due to iframe
+        kpxcUI.createNotification('error', tr('credentialsBlockedInIframe'));
```
varjolintu, line 363: This should probably be inside the `isIframeAllowed()` after the `logDebug()` for sandboxed iframe. Otherwise this notification is triggered every time a non-allowed iframe is trying to be filled?

keepassxc-browser/content/keepassxc-browser.js
droidmonkey: This function doesn't return anything

varjolintu: Hah, true.

keepassxc-browser/content/fill.js
droidmonkey: isIframeAllowed is called all over the place and is done prior to credential retrieval from KeePassXC. It would not be appropriate to show a banner at this stage of the process. The banner notification should only show when you try to fill credentials.

I am kind of abusing the function call by calling it AGAIN in the notification call here, but the effect is the same as if we set a global boolean or something less desirable.

varjolintu: Sounds reasonable.

varjolintu (approved):

Issue #2645, opened 2025-08-22
Potential Clickjacking


philsward:
Potential issue that may need eyeballs.

Pretty much ALL password manager browser extensions appear to suffer from a DOM clickjack.

https://thehackernews.com/2025/08/dom-based-extension-clickjacking.html?m=1

I'm just the messenger.

hannes-ucsc:
This is a more authoritative source (the researcher's blog): https://marektoth.com/blog/dom-based-extension-clickjacking/

droidmonkey:
We tested against the articles criteria, and we were only susceptible to one issue, which has been corrected with our 1.9.9.2 release.

varjolintu:
Some details. The clickjacking can happen with versions before 1.9.9.2. only if:
- User has remembered the credential allowance setting for that site. By default KeePassXC will prompt for the access every time the extension triggers a credential request. So with default settings we are not affected at all. If credentials are stored to KeePassXC using the extension, the site is automatically allowed.
- The criteria above is met. User interacts with the site, or has Auto-Fill enabled for a single credential. If Auto-Fill is disabled, just opening the page is not enough to trigger the credential request.
- Parent element opacity is not zero. The fix for 1.9.9.2 raises this treshold to the default where 0.7 is the minimum accepted opacity.

Few other things to note:
- Iframes from another domain are not accepted by default. User needs to allow this explicitly as well, per site.
- All DOM content provided by the extension is inside a closed Shadow DOM. The page scripts have no access to the credentials before filling.

varjolintu:
Seems we are still vulnerable to overlays:
https://marektoth.com/blog/dom-based-extension-clickjacking/#keepassxc-browser

Test URLs:
https://websecurity.dev/password-managers/dom-based-extension-clickjacking/overlay/
https://websecurity.dev/password-managers/dom-based-extension-clickjacking/overlay/visible/

thomasboehm:
https://addons.mozilla.org still has 1.9.9.1. How long does it usually take to get the latest version from there? I guess I'll lose the auto update if I install the latest version from Github.

varjolintu:
> https://addons.mozilla.org still has 1.9.9.1. How long does it usually take to get the latest version from there? I guess I'll lose the auto update if I install the latest version from Github.

Nowadays about a week. We'll try to hit another update to the stores soon.

varjolintu:
And a note for this thread too:
[...]

PR #2651, opened 2025-08-24
Protect against overlays

varjolintu:
Overlay popups are detected and checked against the input field positions. If the overlay partially covers (or is inside) the input field, it is ignored.

Also any other elements that can overlap the inputs are checked with the `document.elementFromPoint()` call. All input fields are checked from three different points for possible overlaps.

Some of the unit tests now fail becuse of the topmost check, and that is mostly because how our tests are constructed. Few of those are now temporarily disabled until I figure out a solution for it (probably a lot to rewrite). The fix itself has higher priority than the few tests.

Related issue: #2645

## Testing strategy
Using the following URLs:
https://websecurity.dev/password-managers/dom-based-extension-clickjacking/overlay/
https://websecurity.dev/password-managers/dom-based-extension-clickjacking/overlay/visible/
https://websecurity.dev/overlay/login.html
https://websecurity.dev/overlay/login-opacity.html

## Type of change

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -107,6 +107,7 @@
         "createStylesheet": "readonly",
         "DatabaseState": "readonly",
         "debugLogMessage": "readonly",
+        "elementsOverlap": "readonly",
         "EXTENSION_NAME": "readonly",
         "getCurrentTab": "readonly",
         "getLoginData": "readonly",
--- a/keepassxc-browser/common/global.js
+++ b/keepassxc-browser/common/global.js
@@ -182,10 +182,18 @@ const getCurrentTab = async function() {
     return tabs?.length > 0 ? tabs[0] : undefined;
 };
 
+// Check if two elements overlap
+const elementsOverlap = function(rect1, rect2) {
+    const isInside = (a, b) => (b.x >= a.x || b.right <= a.right) && (b.y >= a.y || b.bottom <= a.bottom);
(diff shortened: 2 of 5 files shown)


droidmonkey:
This doesn't seem to work, I can still interact with KeePassXC icon while an overlay is on top.

https://github.com/user-attachments/assets/543bfc36-0234-4d1d-a5e2-0d4c46908069


https://github.com/user-attachments/assets/98b83ed5-b170-4ff2-b955-0233cb2ef5ff

Note: I selected custom fields before this worked so maybe that is bypassing the overlay checker.

varjolintu:
@droidmonkey Using Custom Fields or filling any field manually via context menu bypasses those (manual filling to a selected field must work always even if the input is not recognized automatically).

droidmonkey:
Should merely defining custom fields cause overlay protection to be bypassed? Is there a specific reason for that? I just don't like that I can just click around until it works without ever seeing the keepassxc icon.

I agree that manual selection and intervention, like shown on the second video, is a case where we don't want to prevent action.

varjolintu:
We must ignore Custom Fields as well if there are any overlays. Thanks for pointing out this scenario.

varjolintu:
Selecting Custom Fields should be still allowed. I tested checking the topmost element there, but it breaks the field selectors. This is not a big thing because the whole point of Custom Fields is to be able to select almost every possible field, and it's a manual thing.

droidmonkey:
Works great! I cannot replicate my previous findings.

droidmonkey (approved):

PR #2670, opened 2025-09-01
Cache overlay discovery prior to checking top-level elements

### LINKED ISSUE #2669: extension slows down webpages with an extremely excessive number of forms

### Brief Summary

On documents with an excessive number of forms (10k+) the browser will report a slowdown due to the extension in the following function
https://github.com/keepassxreboot/keepassxc-browser/blob/1.9.9.3/keepassxc-browser/content/fields.js#L414-L439

Image

In a previous version this used to not be an issue and have no significant slowdown/page hang. But in recent versions it has now caused these pages to become unresponsive

### Steps to Reproduce

1. Construct a [webpage with an excessive number of forms](https://gist.github.com/Scrxtchy/8d5785506f738b09b691023c9baec32b) (10 thousand of these)
2. Load said webpage

```Text
KeePassXC - 2.7.10
KeePassXC-Browser - 1.9.9.3
Operating system: Win64
Browser: Mozilla Firefox 143.0
```

droidmonkey:
* Fixes #2669

This completely solves the problem, however I am concerned with two things:
1. Is this the right location to "cache" the overlay discovery? Does this init function get called ever again after page load?
2. What happens if an overlay appears AFTER the discovery, should we add a mutation observer, can we?

## Testing strategy
Tested on Edge using the performance tab in developer tools.

## Type of change

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -126,6 +126,7 @@
         "isEdge": "readonly",
         "isElementInside": "readonly",
         "isFirefox": "readonly",
+        "isIframeAllowed": "readonly",
         "keepass": "readonly",
         "keepassClient": "readonly",
         "kpActions": "readonly",
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -411,6 +411,18 @@ kpxcFields.isSearchField = function(target) {
     return false;
 };
 
+// :popover-open selector is supported only with Firefox >= 125 and Chrome >= 114
+kpxcFields.discoverOverlays = function() {
+     try { 
+        kpxcFields.overlays = document.querySelectorAll(':popover-open, [popover]');
+    } catch (e) {
+        // Ignore SyntaxError (e.g., unsupported selector)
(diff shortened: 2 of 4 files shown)


varjolintu:
Caching is a good idea for this. Need to test it that MutationObserver triggers the update if something changes. Overlays can be added to the page via script and any changes might go undetected.

varjolintu:
I added some changes to this PR. MutationObserver should now detect dynamically added overlays and remove any credentials if input fields are no longer visible.

Tested with locally saved https://websecurity.dev/password-managers/dom-based-extension-clickjacking/loginform/overlay/ where I replaced `document.body.appendChild(dialog);` with
```
setTimeout(() => {
    document.body.appendChild(dialog);
}, 2000);
```

And increased the timeout of for "POPOVER ON COOKIE BANNER" to 5000.

droidmonkey:
This works great, ready to ship. Tested on the massive submit button page as well, gtg.

varjolintu (approved):

PR #2676, opened 2025-09-04
Protect against form overlays

varjolintu:
Any elements with `popover` attributes must be included to the overlays list.

Related issue: #2645

## Testing strategy
Using the sites:
https://websecurity.dev/password-managers/dom-based-extension-clickjacking/loginform/overlay/
https://websecurity.dev/password-managers/dom-based-extension-clickjacking/loginform/overlay/hidden/

## Type of change

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -439,11 +439,19 @@ kpxcFields.isTopElement = function(elem, rect) {
     }
 
     // Check for popup overlays
-    const overlays = document.querySelectorAll(':popover-open, [popover]');
-    for (const overlay of overlays) {
-        const overlayRect = overlay?.getBoundingClientRect();
-        if (overlayRect && elementsOverlap(rect, overlayRect)) {
-            return false;
+    try { 
+        // :popover-open selector is supported only with Firefox >= 125 and Chrome >= 114
+        const overlays = document.querySelectorAll(':popover-open, [popover]');
+        for (const overlay of overlays) {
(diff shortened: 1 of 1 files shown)


droidmonkey:
Should we do the caching as well?

varjolintu:
> Should we do the caching as well?

That would be nice to have.

keepassxc-browser/content/fields.js
```diff
@@ -439,7 +439,7 @@ kpxcFields.isTopElement = function(elem, rect) {
     }
 
     // Check for popup overlays
-    const overlays = document.querySelectorAll(':popover-open');
+    const overlays = document.querySelectorAll(':popover-open, [popover]');
```
droidmonkey: Wrap this call in a try/catch block to ignore SyntaxError throws on older firefox.

droidmonkey (changes requested):

keepassxc-browser/content/fields.js
varjolintu: I was thinking about that, but then we lose the overlay protection for older Firefoxes. We do that (and mention somewhere that older versions are vulnerable with our extension), or we just bump the minimum version..
Firefox ESR 128 was released in July 9, 2024, so it meets the new minimum version requirement.

droidmonkey: So people using us just fine on 115 esr will lose the extension? We will get complaints. Not having overlay protection in those old browsers on old operating systems is kind of the least of their worries.

droidmonkey (approved):

PR #2721, opened 2025-10-06
Fix JavaScript error in console

ygoe:
This extension causes an annoying log message that tells me about an error that was caught in the extension itself. It could easily be avoided in the first place by adding appropriate checks rather than running into the exception each time and happily telling me about it.

Error message:
Cannot override navigator.credentials:  TypeError: undefined is not a non-null object in passkeys.js:211:17
    <anonymous> moz-extension://25853ed3-e2b0-47d6-ad17-2b172340068c/content/passkeys.js:204
    <anonymous> moz-extension://25853ed3-e2b0-47d6-ad17-2b172340068c/content/passkeys.js:213

## Type of change

### DIFF
--- a/keepassxc-browser/content/passkeys.js
+++ b/keepassxc-browser/content/passkeys.js
@@ -201,12 +201,14 @@
     // select a software authenticator. This could be removed in the future.
     try {
         Object.defineProperty(navigator, 'credentials', { value: passkeysCredentials });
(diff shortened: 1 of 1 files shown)


varjolintu:
What browser gives you this error?

droidmonkey (approved):

ygoe:
Firefox, latest stable, Windows 11. It's been like that for a year at least.

varjolintu (approved):

PR #2724, opened 2025-10-10
Fix isFirefox() check

### LINKED ISSUE #2672: KeePassXC-Browser stops working on Firefox after changing User Agent to Chrome

### Brief Summary

If the User Agent is set to Chrome  in Firefox(by setting general.useragent.override in about:config to "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36" and restarting) the extension gets stuck at "Checking status".

I think this may have something to do with the changes made in this commit:https://github.com/keepassxreboot/keepassxc-browser/pull/2510/files#diff-d916a0d73ed1fe3f8df914320b309db296f0f9f6d651e8be56922591aab07250R79
because the extension used to work with overridden User Agent earlier this year.

Image

Expected behavior: KeePassXC-Browser connects to the database if User Agent is overridden.

Actual behavior: KeePassXC-Browser  gets stuck at "Checking status". 

### Steps to Reproduce

1. Set "general.useragent.override" in about:config to "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36"
2. Restart Firefox
3. Try to connect to the database

```Text
KeePassXC - 2.7.10
KeePassXC-Browser - 1.9.9.3
Browser: Mozilla Firefox 142.0
```

varjolintu:
Fixes `isFirefox()` by setting the variable during extension start. Content scripts only request the already parsed value. Additional check for `browser.runtime.getBrowserInfo()` is added to the check, because this API call only exists in Firefox and currently it's the only way to distinquish Firefox if user has changed the browser user agent to something else.

Also moved few log messages to `console.debug()` instead, and changed `debugLogMessage()` to pass calls to that function. Especially the `page.js` spammed a lot of console messages because of the managed settings check.

* Fixes #2672

## Testing strategy
Manually.

## Type of change

### DIFF
--- a/keepassxc-browser/background/init.js
+++ b/keepassxc-browser/background/init.js
@@ -10,12 +10,6 @@ const contextMenuItems = [
     { title: tr('contextMenuRequestGlobalAutoType'), action: 'request_autotype' }
 ];
 
-const menuContexts = [ 'editable' ];
-    
-if (isFirefox()) {
-    menuContexts.push('password');
-}
-
 const initListeners = async function() {
     /**
      * Generate information structure for created tab and invoke all needed
@@ -161,13 +155,17 @@ const initListeners = async function() {
 };
 
 const initContextMenuItems = async function() { 
+    if (page.isFirefox) {
+        page.menuContexts.push('password');
+    }
+
     // Create context menu items
     await browser.contextMenus.removeAll();
     for (const item of contextMenuItems) {
         try {
             await browser.contextMenus.create({
                 title: item.title,
-                contexts: menuContexts,
+                contexts: page.menuContexts,
                 visible: item.visible,
                 id: item.id || item.action
             });
@@ -180,6 +178,7 @@ const initContextMenuItems = async function() {
 (async () => {
     try {
         await keepass.migrateKeyRing();
+        await page.initBrowser();
         await page.initSettings();
         await page.initSitePreferences();
         await page.initOpenedTabs();
--- a/keepassxc-browser/background/init.js
+++ b/keepassxc-browser/background/init.js
@@ -154,7 +154,8 @@ const initListeners = async function() {
     });
 };
 
-const initContextMenuItems = async function() { 
+const initContextMenuItems = async function() {
+    page.menuContexts = [ 'editable' ];
     if (page.isFirefox) {
         page.menuContexts.push('password');
     }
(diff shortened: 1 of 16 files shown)


keepassxc-browser/background/init.js
```diff
@@ -161,13 +155,17 @@ const initListeners = async function() {
 };
 
 const initContextMenuItems = async function() { 
+    if (page.isFirefox) {
+        page.menuContexts.push('password');
```
droidmonkey, line 160: As unlikely as it is to call this function multiple times, you should "reset" the menuContexts first in this function. Moving the initial entry of "editable" into here would be prudent.

keepassxc-browser/options/options.js
```diff
@@ -982,6 +983,8 @@ window.addEventListener('scroll', function() {
 
         const keyRing = await browser.runtime.sendMessage({ action: 'load_keyring' });
         options.keyRing = keyRing;
+        options.isFirefox = await browser.runtime.sendMessage({ action: 'is_firefox' });
```
droidmonkey, line 986: Shouldn't this be sussed out BEFORE calling the addEventListeners for click and scroll?

keepassxc-browser/popups/popup_functions.js
```diff
@@ -16,7 +16,8 @@ async function initSettings() {
     });
 
     const customLoginFieldsButton = document.body.querySelector('#settings #choose-custom-login-fields-button');
-    if (isFirefox()) {
+    const isFirefox = browser.runtime.sendMessage({ action: 'is_firefox' });
+    if (isFirefox) {
```
droidmonkey, line 20: same here

droidmonkey (changes requested):

keepassxc-browser/options/options.js
varjolintu: What click and scroll events do you mean here (and below)?

droidmonkey: sorry this was a PR review failure, it was reporting this was in the add listener call... didn't look above this code block.

droidmonkey (approved):

PR #2733, opened 2025-10-25
Check overlays before fill

varjolintu:
Probably because of the caching feature of overlays (introduced in #2670) the overlay DOMRect can sometimes be full of zeroes and its size cannot be reliably compared. Or the overlay size is updated after we have built the cache. We should check possible overlays also before filling any input fields if the overlay has passed our checks.

Moved the relevant `.setValue()` and `.setValueWithChange()` functions to `kpxcFill`.

## Testing strategy
Saw this happening on https://websecurity.dev/password-managers/dom-based-extension-clickjacking/loginform/overlay/. If the overlay is detected too early, sometimes it's possible to fill credentials using Autocomplete Menu or the popup.

## Type of change

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -466,6 +466,28 @@ kpxcFields.checkExistingFields = function() {
     }
 };
 
+// Check for popup overlays
+kpxcFields.isOverlayOnTop = function(rect) {
+    for (const overlay of kpxcFields.overlays ?? []) {
+        if (kpxcSites.overlayExceptionFound(overlay)) {
+            continue;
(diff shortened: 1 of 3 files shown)


droidmonkey:
To clarify the change, does this change the overlay detection to a "just in time" detection when a fill is attempted? If so, that is a great idea.

varjolintu:
> To clarify the change, does this change the overlay detection to a "just in time" detection when a fill is attempted? If so, that is a great idea.

No, but that could be even better addition. Updating the cache again before fill could work nicely.

droidmonkey (approved):

Issue #2286, opened 2024-07-17
[Serious] Breaks browser styleseets for XML documents


benbucksch:
## Expected Behavior
* The XML, as a tree, expanded (and branches collapsable).
* Stylesheets of the browser and the websites are not affected or destroyed

## Current Behavior
* The expected and correct display flashes for a split-second on the screen, then:
* The XML goes completely unstyled, and you see the tags and attributes as plaintext text, without linebreaks
* It appears that there is a stylesheet applied correctly, but then the KeePassXc browser extension comes in and removes it or breaks it.
* The XML is completely unreadable.

## Possible Solution
* Don't interfere with browser or site CSS style sheets.
* Scope your own stylesheet properly to avoid class name clashes

## Steps to Reproduce (for bugs)
1. Open https://v1.ispdb.net/office365.com in Firefox
(This URL returns XML with a Content-Type of "text/xml")

## Debug info
KeePassXC - 2.7.7
KeePassXC-Browser - 1.9.1.1
Operating system: Linux
Browser: Firefox 128.0 and earlier

varjolintu:
Thanks. Found the reason. The extension does not check the content type before using content scripts. I made a fix where all XML files are ignored with the file extension type or checking the ContentType.

benbucksch:
Thanks for the super fast response.

Can you use a whitelist instead of a blacklist, to avoid that other file types are broken?

droidmonkey:
Other file types have style sheets?

varjolintu:
> Other file types have style sheets?

I think this depends on the browser. Some file types are shown inside an internal page that can also allow some styling. At least this seem to happen with Firefox.

PR #2748, opened 2025-11-07
Fix passkeys exclusion rules for `xml`

a2kolbasov:
Not all URLs containing `.xml` point to XML files. For example:
  - https://bafybeifvkyzsg6skzb3tsk7u7sueex574fzzwfoad7sebudldbv6sazuki.ipfs.dweb.link/test.xml.html
  - https://bafybeifvkyzsg6skzb3tsk7u7sueex574fzzwfoad7sebudldbv6sazuki.ipfs.dweb.link/test.xml/html
  - https://bafybeifvkyzsg6skzb3tsk7u7sueex574fzzwfoad7sebudldbv6sazuki.ipfs.dweb.link/?test=.xml

Not all XML files containing `.xml`:
- https://v1.ispdb.net/office365.com

Type checking is implemented in

https://github.com/keepassxreboot/keepassxc-browser/blob/c2dd4cbb98b1dd8df347d2228d3f01cbe6095e74/keepassxc-browser/content/passkeys-inject.js#L90-L94

## Screenshots or videos

## Testing strategy

Through the self-written site above. (You can substitute another URL path)

- #2506
- #2287

## Type of change

### DIFF
--- a/dist/manifest_chromium.json
+++ b/dist/manifest_chromium.json
@@ -69,7 +69,6 @@
                 "<all_urls>"
             ],
             "exclude_matches": [
-                "*://*/*.xml*",
                 "file:///*.xml*"
             ],
             "js": [
--- a/dist/manifest_firefox.json
+++ b/dist/manifest_firefox.json
@@ -82,7 +82,6 @@
                 "<all_urls>"
             ],
             "exclude_matches": [
-                "*://*/*.xml*",
                 "file:///*.xml*"
             ],
             "js": [
--- a/keepassxc-browser/manifest.json
+++ b/keepassxc-browser/manifest.json
@@ -69,7 +69,6 @@
                 "<all_urls>"
             ],
             "exclude_matches": [
-                "*://*/*.xml*",
                 "file:///*.xml*"
             ],
             "js": [


droidmonkey:
~~Removing the block is NOT correct. It is on there for good reason. Fixing the regex would be correct to only block actual xml _files_.~~

a2kolbasov:
> Removing the block is NOT correct. It is on there for good reason. Fixing the regex would be correct to only block actual xml _files_.

Before this, I examined git log. The addition of the rule (#2287) was a response to #2286. It contains the following example XML file:

https://v1.ispdb.net/office365.com

It doesn't fit the pattern.

Then this pattern was then removed (#2506). But not for Passkeys.

The main code checks Content-Type:
https://github.com/keepassxreboot/keepassxc-browser/blob/c2dd4cbb98b1dd8df347d2228d3f01cbe6095e74/keepassxc-browser/content/keepassxc-browser.js#L889-L898

Passkeys code too:
https://github.com/keepassxreboot/keepassxc-browser/blob/c2dd4cbb98b1dd8df347d2228d3f01cbe6095e74/keepassxc-browser/content/passkeys-inject.js#L89-L94

---

As I understand [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns#examples), the pattern isn't a regular expression (it's very limited). It also doesn't track content type, only URLs.

droidmonkey:
Ah I see thank you

droidmonkey (approved):

varjolintu:
Hmm. Is there some site that actually offers passkeys using those kind of URLs that seem to point to a XML file?

a2kolbasov:
I haven't seen such sites. Therefore, I tested it using a self-written example (links in the PR header).

varjolintu:
> I haven't seen such sites. Therefore, I tested it using a self-written example (links in the PR header).

This is why I think this fix is not even needed.

droidmonkey:
I agree, we added this for a reason and removing it without any specific complaints is not wise.

a2kolbasov:
😕 In that case, will you roll back #2506? Or is there some reason why the search for fields and their auto-completion with traditional password should work on all sites, and WebAuthn only where URL path does not contain `.xml`?

And on which "website" did you check #2416 PR? Here manual analysis proved sufficient.

a2kolbasov:
> I agree, we added this for a reason and removing it without any specific complaints is not wise.

https://github.com/keepassxreboot/keepassxc-browser/pull/2748#issuecomment-3505394330

varjolintu:
> 😕 In that case, will you roll back #2506? Or is there some reason why the search for fields and their auto-completion with traditional password should work on all sites, and WebAuthn only where URL path does not contain `.xml`?

In my opinion we should ignore those URL paths as well. The idea is to restrict content scripts and injected scripts for working with such URLs. I don't see any actual reason why we should remove those protections. If a site is hosting WebAuthn stuff under those kind of strange/faulty URLs where the content doesn't match, it's their problem.

I'd rather block content on URL level than let content/injects scripts to fall in their place and only do some kind of verification there. If you are using your own sites with certain generated content as an example that needs to be fixed, it's pretty far away from an actual site where this kind of error even could happen.

> And on which "website" did you check #2416 PR? Here manual analysis proved sufficient.

Just matching the error codes with: https://github.com/keepassxreboot/keepassxc/blob/develop/src/browser/BrowserMessageBuilder.h

PR #2752, opened 2025-11-09
Added option for connection timeout used when browser extension is trying to connect to KeePassXC

### LINKED ISSUE #2751: Reconnect timeout hardcoded to 1500ms creates problems with QubesOS split pass functionality
### Have you searched for an existing feature request?

- [x] Yes, I tried searching

### Brief Summary

Hi guys,

I am not exactly sure if this is a new requirement or not, but I searched open issues and it seems none of them matches what is required here.

The background story: I am using QubesOS with split pass behavior, which basically connects to a separate qube (vault) from within the browser qube, where a custom keepass proxy is running. During connection a dialog appears whether this connection should be allowed or not. The whole process of showingn the message, waiting for the ok button to become active and confirming does take roundabout 1-2 seconds.

The problem: the keepassxc browser extension does hardcode a connection timeout of 1,5s here:

Image

This makes it very tricky to actually confirm the qube connection before the timeout kicks in. My requirement would be to:

- either increase timeout to something like 3 seconds
- make this timeout configurable (which would be the best option, but requires more effort)

Is this something that could be changed?

Thanks!

### Example

_No response_

### Context

_No response_

Talkabout:
This small feature adds a connection timeout setting which is used when keepassxc-browser extension is trying to connect to KeePassXC instance.

Fixes https://github.com/keepassxreboot/keepassxc-browser/issues/2751

## Screenshots or videos

image

## Testing strategy
I tested this feature by making sure that:
- default value (1.5) is considered, if no other value has been set
- changed value (10) is used when assigned in extension options

## Type of change

### DIFF
--- a/keepassxc-browser/_locales/de/messages.json
+++ b/keepassxc-browser/_locales/de/messages.json
@@ -1018,6 +1018,14 @@
         "message": "Anmeldedaten werden nach dem Timeout aus Hintergrund-Tabs entfernt. Für diese Seiten wird erneut um Erlaubnis gebeten.",
         "description": "Clear credentials timeout help text."
     },
+    "optionsConnectionTimeout": {
+        "message": "KeepassXC Zeitlimit für die Verbindung (Sekunden, 2-60).",
+        "description": "Connection timeout label text."
+    },
+    "optionsConnectionTimeoutHelpText": {
+        "message": "Wenn die KeePassXC Browser Erweiterung versucht, sich mit KeePass zu verbinden, wird dieses Zeitlimit verwendet.",
+        "description": "Connection timeout label text."
+    },
     "optionsVersionInfoText": {
         "message": "KeePassXC-Browser benötigt KeePassXC, um Anmeldedaten abzurufen.",
         "description": "Settings page version info text."
--- a/keepassxc-browser/_locales/de/messages.json
+++ b/keepassxc-browser/_locales/de/messages.json
@@ -1018,14 +1018,6 @@
         "message": "Anmeldedaten werden nach dem Timeout aus Hintergrund-Tabs entfernt. Für diese Seiten wird erneut um Erlaubnis gebeten.",
         "description": "Clear credentials timeout help text."
     },
-    "optionsConnectionTimeout": {
-        "message": "KeepassXC Zeitlimit für die Verbindung (Sekunden, 2-60).",
-        "description": "Connection timeout label text."
-    },
-    "optionsConnectionTimeoutHelpText": {
-        "message": "Wenn die KeePassXC Browser Erweiterung versucht, sich mit KeePass zu verbinden, wird dieses Zeitlimit verwendet.",
-        "description": "Connection timeout label text."
-    },
     "optionsVersionInfoText": {
         "message": "KeePassXC-Browser benötigt KeePassXC, um Anmeldedaten abzurufen.",
         "description": "Settings page version info text."
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -804,7 +804,12 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab = null, connectionTimeout = page.settings.connectionTimeout) {
+keepass.reconnect = async function(tab = null, connectionTimeout = CONNECTION_TIMEOUT) {
+    if (arguments.length < 2) {
+        connectionTimeout = page.settings.connectionTimeout;
+    }
+
+    connectionTimeout = Math.min(60000, Math.max(CONNECTION_TIMEOUT, connectionTimeout))
     keepassClient.connectToNative();
     keepass.generateNewKeyPair();
     const keyChangeResult = await keepass.changePublicKeys(tab, !!connectionTimeout, connectionTimeout).catch(() => false);
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -804,11 +804,7 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab = null, connectionTimeout = -1) {
-    if (connectionTimeout === -1) {
-        connectionTimeout = page.settings.connectionTimeout * 1000;
-    }
-
+keepass.reconnect = async function(tab = null, connectionTimeout = page.settings.connectionTimeout) {
     keepassClient.connectToNative();
     keepass.generateNewKeyPair();
     const keyChangeResult = await keepass.changePublicKeys(tab, !!connectionTimeout, connectionTimeout).catch(() => false);
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -805,7 +805,7 @@ keepass.disableAutomaticReconnect = function() {
 };
 
 keepass.reconnect = async function(tab = null, connectionTimeout = -1) {
-    if (connectionTimeout == -1) {
+    if (connectionTimeout === -1) {
         connectionTimeout = page.settings.connectionTimeout * 1000;
     }
 
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -804,7 +804,11 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab = null, connectionTimeout = 1500) {
+keepass.reconnect = async function(tab = null, connectionTimeout = -1) {
+    if (connectionTimeout == -1) {
+        connectionTimeout = page.settings.connectionTimeout * 1000;
+    }
+
     keepassClient.connectToNative();
     keepass.generateNewKeyPair();
     const keyChangeResult = await keepass.changePublicKeys(tab, !!connectionTimeout, connectionTimeout).catch(() => false);
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -14,6 +14,7 @@ const defaultSettings = {
     bannerPosition: BannerPosition.TOP,
     checkUpdateKeePassXC: CHECK_UPDATE_NEVER,
     clearCredentialsTimeout: 10,
+    connectionTimeout: 2,
     colorTheme: 'system',
     credentialSorting: SORT_BY_GROUP_AND_TITLE,
     debugLogging: false,
@@ -47,6 +48,7 @@ page.autoSubmitPerformed = false;
 page.attributeMenuItems = [];
 page.blockedTabs = [];
 page.clearCredentialsTimeout = null;
+page.connectionTimeout = null;
 page.currentRequest = {};
 page.currentTabId = -1;
 page.isFirefox = false;
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -14,7 +14,7 @@ const defaultSettings = {
     bannerPosition: BannerPosition.TOP,
     checkUpdateKeePassXC: CHECK_UPDATE_NEVER,
     clearCredentialsTimeout: 10,
-    connectionTimeout: 2,
+    connectionTimeout: 1500,
     colorTheme: 'system',
     credentialSorting: SORT_BY_GROUP_AND_TITLE,
     debugLogging: false,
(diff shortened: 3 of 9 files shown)


keepassxc-browser/_locales/de/messages.json
```diff
@@ -1018,6 +1018,14 @@
         "message": "Anmeldedaten werden nach dem Timeout aus Hintergrund-Tabs entfernt. Für diese Seiten wird erneut um Erlaubnis gebeten.",
         "description": "Clear credentials timeout help text."
     },
+    "optionsConnectionTimeout": {
```
varjolintu: Normally we only add translations to en/messages.json and let the Transifex handle the rest. Translations are synced from there, so any others might get overwritten.

Talkabout: Should I remove it then or let it being overwritten?

keepassxc-browser/background/keepass.js
```diff
@@ -804,7 +804,11 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab = null, connectionTimeout = 1500) {
+keepass.reconnect = async function(tab = null, connectionTimeout = -1) {
+    if (connectionTimeout == -1) {
```
varjolintu: We use strict comparisons, for example `===` and `!==`. Check your ESLint settings if you did't get any warning about these.

keepassxc-browser/_locales/de/messages.json
varjolintu: The safest bet is to remove them. I'll upload the latest translation file (en) to Transifex always before release. So just keep the `en/messages.json` changes.

keepassxc-browser/background/keepass.js
Talkabout: Corrected.

varjolintu: And here you could just use the default value in the function parameter instead of `-1`.

keepassxc-browser/_locales/de/messages.json
Talkabout: Done

keepassxc-browser/background/keepass.js
Talkabout: Not sure what you mean here? Are you suggesting to use "page.settings.connectionTimeout" in the method header? Was my first idea but it seams that at the point of time when the method is created the setting is not loaded yet resulting in unexpected behavior.

Talkabout: Another thing to mention here is the difference between the parameter type in the options and the value that is required in the method header: setting is in seconds, parameter is in milliseconds. So either way we would need to add a handling that allows to transform seconds to milliseconds. What is your preference? From my point of view being able to determine whether the default value has been provided or not is a cleaner approach, but I am open for suggestions.

keepassxc-browser/background/keepass.js
```diff
@@ -804,7 +804,11 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab = null, connectionTimeout = 1500) {
+keepass.reconnect = async function(tab = null, connectionTimeout = -1) {
+    if (connectionTimeout === -1) {
+        connectionTimeout = page.settings.connectionTimeout * 1000;
+    }
```
droidmonkey: This is not a good implementation. connectionTimeout should never be _less than 1500_ and never be more than... let's say 60,000 (1 minute). This ensures proper behavior regardless of user input shenanigans. This function should _expect to receive milliseconds_ and any conversion should happen before calling this function.

droidmonkey (changes requested):

droidmonkey:
I recommend a slide bar that varies between 1.5 seconds and 60 seconds vice a text input.

keepassxc-browser/background/keepass.js
Talkabout: Thanks for the feedback! Will check what ways there are to implement it more in line with your suggestions.

keepassxc-browser/background/keepass.js
varjolintu: Seconds in UI, milliseconds in the actual scripts :)

keepassxc-browser/background/keepass.js
droidmonkey: Rather simply: `connectionTimeout = Math.max(1500, Math.min(60000, connectionTimeout))`

keepassxc-browser/background/keepass.js
Talkabout: Implemented now with a slidebar component.

keepassxc-browser/background/keepass.js
Talkabout: Removed this logic completely and assigned the default value directly in parameter list.

keepassxc-browser/background/keepass.js
```diff
@@ -804,7 +804,7 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab = null, connectionTimeout = 1500) {
+keepass.reconnect = async function(tab = null, connectionTimeout = page.settings.connectionTimeout) {
```
droidmonkey: Do not use a variable for default value.

keepassxc-browser/options/options.js
```diff
@@ -173,7 +176,21 @@ options.initGeneralSettings = async function() {
...
+    $('#tab-general-settings input#connectionTimeout').addEventListener('input', function(e) {
+        $('#connectionTimeoutLabel').textContent = tr('optionsConnectionTimeout', e.target.value);
+    });
+
+    $('#tab-general-settings input#connectionTimeout').addEventListener('change', async function(e) {
+        if (e.target.valueAsNumber < 1.5 || e.target.valueAsNumber > 60) {
+            return;
+        }
+
+        options.settings['connectionTimeout'] = e.target.valueAsNumber * 1000;
```
droidmonkey: This should be condensed into: `options.settings['connectionTimeout'] = Math.min(60000, Math.max(1500, e.target.valueAsNumber * 1000))`. Alternatively, a warning message could be issued if the value falls outside the acceptable range.

keepassxc-browser/background/keepass.js
```diff
@@ -804,7 +804,7 @@ keepass.disableAutomaticReconnect = function() {
...
     keepassClient.connectToNative();
     keepass.generateNewKeyPair();
     const keyChangeResult = await keepass.changePublicKeys(tab, !!connectionTimeout, connectionTimeout).catch(() => false);
```
droidmonkey, line 815: Need to do another Min/Max check here just to be sure passed values make sense and won't cause undesirable behavior.

droidmonkey (changes requested):

keepassxc-browser/background/keepass.js
Talkabout: Which value to choose here then? The idea behind this is to consider the connection timeout setting as the default behavior. I can move that into the method body, but I need to know if a value has been passed or if the default value should be considered. I find the other alternative to pass the setting in every call of this method not optimal.

varjolintu: If a default value must be used in both background script and content script / options page, best place is to create a new const variable for it in `global.js`. That is then set to `page.js` as the default value, and you can refer to it anywhere in the code.

Talkabout: For me to understand: create new const variable i global.js and use it with every call to "reconnect" where no second param is provided? Or use it as the default value in the method definition of "reconnect"?

varjolintu: Everywhere you use the default value.

Talkabout: I am a little bit confused to be honest... I searched for an example implementation in the code but there seems to be none. No constants there are assigned to a settings value.

I checked inside keepass.js and saw that you are using "page.settings" object quite a lot, so wondering why it shouldn't be used directly elsewhere..

Next thing is the fact that you mentioned to NOT use a variable as default value. So even when I create a constant (or a variable in globals js) I won't be able to use it as the default value. This means that every call to the "reconnect"-method needs to pass the default value, which seems wrong to me.

Maybe I am not getting your idea but some hints would be very helpful!

Thanks!

varjolintu: For example `clearCredentialsTimeout` could be used from a global variable defined in `global.js`. And if you look at `credentialSorting` that has a default value from global variable `SORT_BY_GROUP_AND_TITLE`.

`page.settings` is the current settings configured by user, not the default values. Any variables inside that object should not be used as default values in functions. Only as current values.

Talkabout: Sorry for the late reply!

I have now added the constant as you requested and replaced occurrences accordingly.
The decision whether to use the configuration value or not is now taken based on the number of provided arguments to "reconnect" method.

Hope this solves the problems you mentioned.

keepassxc-browser/background/keepass.js
```diff
@@ -804,7 +804,12 @@ keepass.disableAutomaticReconnect = function() {
     keepass.reconnectLoop = null;
 };
 
-keepass.reconnect = async function(tab = null, connectionTimeout = page.settings.connectionTimeout) {
+keepass.reconnect = async function(tab = null, connectionTimeout = CONNECTION_TIMEOUT) {
+    if (arguments.length < 2) {
```
varjolintu, line 808: This code block is not needed.

Talkabout: From my point of view it is the most important part of the change as it makes sure to use the user configured timeout if no argument is provided. Am I missing something?

varjolintu: The default value `CONNECTION_TIMEOUT` in the function is already setting that to the value if no argument is provided.

Talkabout: CONNECTION_TIMEOUT is set to a hard value of 1500 in global.js. The user defined value in the settings is not the same value. We need to make sure that if a developer has not provided any value to the second argument of "reconnect" we are considering the settings value.

droidmonkey: By definition, by allowing a connectionTimeout parameter at all we can _ignore the setting value_. You can set it to whatever you want with a direct call. If we don't want that behavior then you need to remove the parameter altogether and always use the settings value (defaulting to CONNECTION_TIMEOUT if invalid or unset).

Talkabout: The idea behind the implementation was to keep the backward compatibility of the "reconnect" method calls to not require the second parameter and still use the user setting value. I am aware that passing the settings value with every call would be possible too, but that means that in the future a developer needs to remember to pass the settings value when a new call to "reconnect" is made. To avoid this dependency I decided to handle it in "reconnect" directly. Do you think passing the settings value with every "reconnect" call is a better solution?

droidmonkey: My preference is to drop the parameter altogether and use the settings value directly in this function, falling back to the default value if unset. This is an internal function call and there is no reason to retain any compatibility.

Talkabout: I don't think dropping this parameter is an option. Looking at background/init.js there we are using this exact parameter to pass a different timeout. It is the only place I found, but I assume this was changed for a reason :)

keepassxc-browser/background/page.js
```diff
@@ -47,6 +48,7 @@ page.autoSubmitPerformed = false;
 page.attributeMenuItems = [];
 page.blockedTabs = [];
 page.clearCredentialsTimeout = null;
+page.connectionTimeout = null;
```
varjolintu, line 51: This variable is not used, and not needed.

keepassxc-browser/background/keepass.js
varjolintu: The easiest solution for this is to:
- Keep the default parameter value `connectionTimeout = CONNECTION_TIMEOUT` in the function. It doesn't hurt to keep the default here.
- From `init.js` the parameter is overridden with `5000` (we should make this a const too).
- The function is called from `kpxcEvent.onReconnect()` where the `page.settings.connectionTimeout` must be used as a parameter.
- The function is called also from `keepass.enableAutomaticReconnect()`. This function is not used at the point, but it should also include `page.settings.connectionTimeout` as the parameter.

keepassxc-browser/options/options.js
```diff
@@ -124,6 +124,9 @@ options.initGeneralSettings = async function() {
     $('#tab-general-settings input#defaultGroup').value = options.settings['defaultGroup'];
     $('#tab-general-settings input#defaultPasskeyGroup').value = options.settings['defaultPasskeyGroup'];
     $('#tab-general-settings input#clearCredentialTimeout').value = options.settings['clearCredentialsTimeout'];
+    const connectionTimeout = (options.settings['connectionTimeout']/1000);
```
varjolintu, line 127: Don't do this. Just retrieve the value from settings and do any modifications to it when inserting it as element value.

PR #2789, opened 2025-12-04
Show notifications always on the top window

### LINKED ISSUE #2255: Notifications (errors) sometimes partly invisible (YouTrack)
In some cases, KeePassXC error messages after trying to fill in credentials are not fully visible. This happens at least on Pidgin's ITS, which is powered by JetBrains YouTrack.

## Expected Behavior
The notification's full contents ("Error! Non connecté à KeePassXC.") should be visible: 
image

## Current Behavior
The start of the notification's text is hidden, as can be seen in the following screenshot: 
KeePassXC notification

## Possible Solution
I could not figure out why the context outside of the dialog displays above the notification. The `div.kpxc-notification` has `z-index: 2147483646`, which is (hopefully) above everything else, so I would have thought this would suffice.

However, another bug is that the notification's width (520 pixels) exceeds that of the dialog supposed to contain it (397 px). This is a CSS bug caused by [`.kpxc-notification`'s `width: 520px`](https://github.com/keepassxreboot/keepassxc-browser/blob/917e2d8f1aeeabb207a8dc871ec383f6c942bb00/keepassxc-browser/css/notification.css#L18). Changing to `max-width` suffices to fix, but I can't say why even that would be needed. Removing the property altogether fixes for me, and even improves the result when this bug doesn't happen.

A more robust approach would be to avoid playing with the DOM and just display notifications outside of the tab's content.

1. Make sure that KeePassXC is not open
2. Open https://issues.imfreedom.org/projects
3. Activate the _Log in..._ button (top right)
4. Activate KeePassXC's login form button (the icon in the form's first field)

## Affected environments
KeePassXC 2.7.9
KeePassXC-Browser - 1.9.0.5
Operating system: Windows 11
Browser: Chrome, Firefox 127, Edge

## Debugging tips
KeePassXC automatically hides its notifications after 5 s, making it hard to use the inspector for debugging. To workaround that without changing the extension's code, one may put a breakpoint on [`kpxcUI.createNotification`'s last line](https://github.com/keepassxreboot/keepassxc-browser/blob/917e2d8f1aeeabb207a8dc871ec383f6c942bb00/keepassxc-browser/content/ui.js#L314) (closing brace) and run `clearTimeout(notificationTimeout);` in the console before continuing.

varjolintu:
If a notification is shown inside an iframe, it's highly possible that it will be partially cut off, or is displayed in strange position. If `kpxcUI.createNotification()` is called inside an iframe, pass the message to the top window and show it there instead.

* Fixes #2255

## Testing strategy
Any site with login form inside an iframe, e.g. https://mail.163.com/ or https://issues.imfreedom.org/projects. Just don't have any entries for it, and click the Username Icon.

## Type of change

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -939,9 +939,14 @@ browser.runtime.onMessage.addListener(async function(req, sender) {
             kpxcFill.fillAttributeToActiveElementWith(req.args);
         } else if (req.action === 'frame_message') {
             if (req.args?.[0] === 'frame_request_to_frames' && window.self !== window.top) {
+                // Handle message from top window in iframe
                 kpxcCustomLoginFieldsBanner.handleParentWindowMessage(req.args);
             } else if (req.args?.[0] === 'frame_request_to_parent' && window.self === window.top) {
+                // Handle message from iframe in top window
                 kpxcCustomLoginFieldsBanner.handleTopWindowMessage(req.args);
+            } else if (req.args?.[0] === 'notification_from_frame' && window.self === window.top) {
+                // Handle notification from iframe in top window
+                kpxcUI.createNotification(req?.args[1], req?.args[2]);
             }
         } else if (req.action === 'ignore_site') {
             kpxc.ignoreSite(req.args);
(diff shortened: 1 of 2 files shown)


droidmonkey (approved):

droidmonkey:
This will be a very welcome change.

PR #2791, opened 2025-12-06
Refactor icon handling

varjolintu:
A small refactor to icon handling. Mostly functions are just moved around so it makes more sense.

Includes the following changes:
- Rename `content/icons.js` to `content/icon-handler.js`.
- Create a new `content/icon.js` for the `Icon` base class (moved from `content/ui.js`).
- Move all icon-related functions from `content/ui.js` to `content/icon-handler.js`. UI file should include only general functions. Maybe `createNotification()` could also be a separate file in the next refactor round.

## Testing strategy
Use a site that has disappearing icons, for example https://security.nl site's login popup. Everything should work normally.

## Type of change


varjolintu:
Great catch from unit tests with the first commit :)

keepassxc-browser/content/icon-handler.js
```diff
@@ -0,0 +1,238 @@
...
+ * @Object kpxcIcons
+ * Icon handling.
+ */
+const kpxcIcons = {};
+kpxcIcons.icons = [];
+kpxcIcons.iconTypes = { USERNAME: 0, PASSWORD: 1, TOTP: 2 };
+
+// Adds an icon to input field
+kpxcIcons.addIcon = async function(field, iconType) {
+    if (!field || iconType < 0 || iconType > 2) {
```
droidmonkey: since you define iconTypes, why not just check if the value is in the set? A debug warning should be thrown here as well since there is clearly an issue in the code if iconType is not valid.

keepassxc-browser/content/icon-handler.js
```diff
@@ -0,0 +1,238 @@
...
+            // For example PayPal and Amazon is handled like this.
+            if (c.username && !c.password && c.passwordInputs.length === 1) {
+                kpxcIcons.addIcon(c.passwordInputs[0], kpxcIcons.iconTypes.USERNAME);
+            }
+
+            if (c.username && !c.username.readOnly) {
+                kpxcIcons.addIcon(c.username, kpxcIcons.iconTypes.USERNAME);
+            } else if (c.password && (!c.username || (c.username && c.username.readOnly))) {
+                // Single password field
+                kpxcIcons.addIcon(c.password, kpxcIcons.iconTypes.USERNAME);
```
droidmonkey: USERNAME type?

keepassxc-browser/content/icon-handler.js
```diff
@@ -0,0 +1,238 @@
...
+};
+
+// Adds all icons from a form struct
+kpxcIcons.addIconsFromForm = async function(form) {
+    const addUsernameIcons = async function(c) {
+        if (kpxc.settings.showLoginFormIcon && await kpxc.passwordFilledWithExceptions(c) === false) {
+            // Special case where everything else has been hidden, but a single password field is now displayed.
+            // For example PayPal and Amazon is handled like this.
+            if (c.username && !c.password && c.passwordInputs.length === 1) {
+                kpxcIcons.addIcon(c.passwordInputs[0], kpxcIcons.iconTypes.USERNAME);
```
droidmonkey: USERNAME type?

droidmonkey (commented):
Quick initial read, seems to have some inconsistency on when you use USERNAME vs PASSWORD icon types. Maybe that is a naming issue only?

keepassxc-browser/content/icon-handler.js
varjolintu: Yes, if there's only a single password field on the page, we still use the default icon (Username) for the fill. Password icon is reserved only for password fields where we want to trigger the password generator.

keepassxc-browser/content/icon-handler.js
varjolintu: Same here as above.

keepassxc-browser/content/icon-handler.js
varjolintu: Good point. Let's improve this a bit. Also, numbers are used here instead of enums..

keepassxc-browser/content/icon-handler.js
droidmonkey: Recommend rename USERNAME to DEFAULT.

droidmonkey (approved):
This works well, although PayPal is still broken. The password field doesn't get an icon after setting the username and clicking next.

varjolintu:
> This works well, although PayPal is still broken. The password field doesn't get an icon after setting the username and clicking next.

There's a separate issue about that: https://github.com/keepassxreboot/keepassxc-browser/issues/2798

PR #2767, opened 2025-11-22
"Reopen database" shortcut

a2kolbasov:
Now the keyboard shortcut uses the `#reopen-database-button` functionality from `popup.js ` ⇒ no credential extraction is required.

Fixes #2766

## Testing strategy

Manually

## Type of change

- ✅ Bug fix (non-breaking change that fixes an issue)
- ✅ New feature (change that adds functionality)

PR #2809, opened 2025-12-29
Force credentials extraction upon manual request

a2kolbasov:

*(continuing of https://github.com/keepassxreboot/keepassxc-browser/pull/2767#issuecomment-3642472272)*

This function is blocked by the `autoRetrieveCredentials` setting. If the user turned it off, then manually requested credentials, and then changed something in the database, he does not have the opportunity to request them again without updating the page.

An option like
```js
if ((kpxc.settings.autoRetrieveCredentials || force) && kpxc.url && kpxc.submitUrl) {
```
is not suitable. It is necessary to separate the desire not to use the cache from the user's explicit request.

```js
kpxc.retrieveCredentials(true)
```
also called here:

https://github.com/keepassxreboot/keepassxc-browser/blob/8c3df94db3b32d6f1e8a8060a5321ad1c4372d98/keepassxc-browser/content/banner.js#L370-L384

If a user saves new credentials, this does not mean that he wants to automatically retrieve them from the database again afterwards (and see a request from the KeePassXC-desktop).

## Screenshots or videos

## Testing strategy

Manually

## Type of change


varjolintu (changes requested):
> If a user saves new credentials, this does not mean that he wants to automatically retrieve them from the database again afterwards (and see a request from the KeePassXC-desktop).

Actually this is something we want to do. Any added or updated credentials should be retrieved from the database.

a2kolbasov:
> Actually this is something we want to do. Any added or updated credentials should be retrieved from the database.

Maybe in this case, we should change the business logic: when updating data, first clear the cache, and then request passwords from the database as if the user had refreshed the page? In this case, users with `autoRetrieveCredentials = false` will simply remain without cached passwords and (if they need them) will explicitly request them again. And those who only want to update their password will not see the "Confirm Access Dialog" from KeePassXC-desktop.

PR #2810, opened 2025-12-30
Passkeys fallback: wait the `window.onfocus` event

### LINKED ISSUE #2493: "passkeys fallback" doesn't appear to work as advertised
> When enabled, a failed or **cancelled** request to KeePassXC will trigger the browser's own internal passkeys request.

This is a little convoluted. Please bear with me:

Using a YubiKey for MFA with CloudFlare with "Enable passkeys fallback" enabled in the extension:
- If no passkey exists in the keepassxc database fallback seems OK; the browser takes over and lets me use the YubiKey.
- But with a passkey present in the database, dismissing the passkey credentials dialog generates an error in the extension, but doesn't fallback to the browser:

Image

KeePassXC - 2.7.10
KeePassXC-Browser - 1.9.7
Browser: Mozilla Firefox 137.0

a2kolbasov:

Some browsers reject requests to `navigator.credentials.create/get` if the page is out of focus (meaning the built-in functionality -- the `originalCredentials` variable). When the user selects a passkey in KeePassXC-desktop, there is no focus on the page. When the request is canceled, a race begins between whether the focus returns to the page and whether the `originalCredentials` call occurs. It depends on the operating system's window manager.

If the user starts switching windows during the request, and when canceling the request, focus is returned to a non-browser, the browser will reject the request regardless of the OS.

Resolves #2493

## Screenshots or videos

## Testing strategy

As in <https://github.com/keepassxreboot/keepassxc-browser/issues/2493#issuecomment-3493120344>.

It is also possible at the time of the Passkey request:
1. Do not cancel it
2. Return to the browser and focus on the Web Console (click on the code entry field), docking option is not "Separate Window")
3. Return to KeePassXC and cancel the request

The focus will return in DevTools and the browser will reject the request.

If you need to call the built-in library in Firefox to work with USB keys, regardless of the OS: go to `about:config`, set `security.webauth.webauthn_enable_softtoken` to `true`.

- <https://www.w3.org/TR/webauthn-2/#sctn-abortoperation:~:text=The%20visibility,aborted>
- <https://github.com/w3c/webauthn/issues/316>

## Type of change


varjolintu:
This is clearly some bug in Firefox itself.

droidmonkey:
This is an ugly workaround to what is ultimately browser behavior. The behavior should be reported to mozilla for correction.

a2kolbasov:
Chrome behaves differently, but still in a similar way.

Disable keepassxc-browser, go to <https://github.com/login>, execute
```js
setTimeout(() => document.querySelector('.js-webauthn-confirm-button').click(), 5_000)
```
, switch to another tab and wait for the timer to expire. When you return, you will see the same error in the console.

------

Browsers block requests from inactive pages, but they define its activity differently.

The call from the password manager is no different (for browsers) from the code on the page that calls `navigator.credentials.get`.

And the spoof of `navigator.credentials` itself is **generally ugly too**. But this is a hack due to the fact that browsers do not provide an API similar to the API for HTTP Auth (or the like).

------

And here are examples from the WebAuthn specification:
- [§ 5. Web Authentication API](https://www.w3.org/TR/webauthn-3/#sctn-api:~:text=scripts%20can%20%28with%20the%20user%E2%80%99s%20consent%29%20request%20the%20browser)
- [§ 5.1.3. Create a New Credential, № 2.2 - 2.3](https://www.w3.org/TR/webauthn-3/#sctn-createCredential:~:text=If%20the%20relevant,object%2E)
- [User Consent](https://www.w3.org/TR/webauthn-3/#user-consent)

a2kolbasov:
The same thing happens in Safari. The video shows:
1. If the focus is on DevTools
2. If the focus is on `window.top` instead of `window.self`.

https://github.com/user-attachments/assets/87e3ef50-f423-4bf2-9bc9-652e35bb289b

_tested on a [self-written site](https://bafybeigty6mi6sumnuwjopidomiidbd5srmd4q4otcyqq5hagd6pxmqguu.ipfs.dweb.link/) via [BrowserStack](https://www.browserstack.com/), MacOS Tahoe, Safari 26.2_

------

I conducted additional testing. `document.hasFocus()` returns `true` in all iframes, regardless of which one is being checked (`document.hasFocus() === window.top.document.hasFocus()` always). However, the `focus` event does not pass through an iframe. When switching between applications, the focus returns to the last active iframe (where the WebAuthn call was made), and the event should be handled on that iframe.

I also replaced a callback with `await Promse<void>`.

thePanz:
Any updates on this direction @a2kolbasov or @droidmonkey ?
Looks like Aleksandr confirmed that the problem is not related to Firefox only

varjolintu:
As said before, this is clearly a bug in the browser itself. It should be reported to Mozilla and Apple.

Zocker1999NET:
> This is an ugly workaround to what is ultimately browser behavior. The behavior should be reported to mozilla for correction.

> As said before, this is clearly a bug in the browser itself. It should be reported to Mozilla and Apple.

I get why you find this workaround ugly. But what would be a "proper" fix by the browser?
Because IMO browsers will not lift the focus requirement for passkey requests, as that otherwise provides all websites you have currently opened with the ability to throw OS-specific passkey popups in your focus whenever they like.

The only thing I can imagine through is that browsers could provide an official API for such third-party passkey provider. That API could then e.g. "remember" that the initial request started while the browser tab was in focus. But implementing such an API is work and would then only be needed for Linux, as Windows already provides an OS-level API for the same feature, but which is more capable. (I have no idea whether Mac OSX provides such an API for apps. And to add further, IMO it would be better to develop a Linux Passkey DBus service to handle this use case similarly to what Windows does.)

a2kolbasov:
> As said before, this is clearly a bug in the browser itself. It should be reported to Mozilla and Apple.

- https://www.w3.org/TR/webauthn-2/#sctn-abortoperation:~:text=When%20the%20Window%20object%20associated%20with%20the%20%5BDocument%20loses%20focus%2C&text=operations%20SHOULD%20be%20aborted
- https://github.com/w3c/webauthn/issues/316

a2kolbasov:
> The only thing I can imagine through is that browsers could provide an official API for such third-party passkey provider.

- https://github.com/w3c/webextensions/issues/361

> But implementing such an API is work and would then only be needed for Linux, as Windows already provides an OS-level API for the same feature, but which is more capable.

This should be done in the browser, not the OS. Password managers like Bitwarden and Proton Pass don't require a separate desktop app -- they work solely as browser extensions. Extensions don't have access to built-in OS APIs.

a2kolbasov:
> This is an ugly workaround to what is ultimately browser behavior.

> I get why you find this workaround ugly.

When KPXC replaces standard functions with its own and manipulates data to provide users with an alternative, that's fine. When I add waiting for the focus-return to restore the functionality of business logic I didn't implement, that's "ugly". Interesting 😅.

... Yeah, I've got miles to go before I can write code as this:

https://github.com/keepassxreboot/keepassxc-browser/blob/f66c0cb148bc5c044240cecef2c250b759562cab/keepassxc-browser/common/sites.js#L78-L115

varjolintu:
> > This is an ugly workaround to what is ultimately browser behavior.
> 
> > I get why you find this workaround ugly.
> 
> When KPXC replaces standard functions with its own and manipulates data to provide users with an alternative, that's fine. When I add waiting for the focus-return to restore the functionality of business logic I didn't implement, that's "ugly". Interesting 😅.
> 
> ... Yeah, I've got miles to go before I can write code as this:
> 
> https://github.com/keepassxreboot/keepassxc-browser/blob/f66c0cb148bc5c044240cecef2c250b759562cab/keepassxc-browser/common/sites.js#L78-L115

Do you think this kind of "code-shaming" is going to get your PR reviewed and merged? FYI: the meaning and usage of Predefined Sites is a totally different thing.

The "ugly" here is not a personal attack against your code, but against the method itself that is possibly required to get passkeys to work properly.

varjolintu:
Please update the PR template to the latest version, including this information (do this to all of your PRs):
https://github.com/keepassxreboot/keepassxc-browser/pull/3001/changes

Zocker1999NET:
> - https://github.com/w3c/webextensions/issues/361

Okay that's interesting to see, I didn't know about a plan to standardize that, but it makes sense.

> This should be done in the browser, not the OS.

That's an opionion depennding on whether you prefer differences per webbrowser vs. configuration overhead per browser. Also without supporting an OS-specific API, it's not possible to use such Passkeys from KeePass in desktop apps (e.g. Windows Remote Desktop can forward passkeys to the remote side, which could become handy).

> Password managers like Bitwarden and Proton Pass don't require a separate desktop app -- they work solely as browser extensions. Extensions don't have access to built-in OS APIs.

At least Bitwarden provides both possibilities, the Desktop app integrates with Windows while the browser extension integrates with the browser.

While reading about Bitwarden, I discovered another possible solution to this problem, quote from [their help page](https://bitwarden.com/help/storing-passkeys/):
page:

> If you previously selected Use your device or hardware key, Bitwarden will not offer to enter or save a passkey. If you clicked this by accident or change your mind, remove the website from blocked domains.

When the browser extension would remember, at least for a short period, that the user denied (possibly with an explicit button), then KeePassXC browser could just passively forward that request, making it possible to use hardware keys on the 2nd try. This at least feels less uglier.

droidmonkey (approved):

varjolintu:
> > * [Add an API to integrate with the Credential Management Web API w3c/webextensions#361](https://github.com/w3c/webextensions/issues/361)
> 
> Okay that's interesting to see, I didn't know about a plan to standardize that, but it makes sense.

Yes. I can clearly see why this is required.

varjolintu:
Tested it without and with the fix:

I tried Cloudflare's login page with passkeys without this fix, and canceling the passkeys request prompted me the native dialog normally.
Screenshot 2026-06-16 at 18 53 14

I didn't have to touch the `security.webauth.webauthn_enable_softtoken` setting in Firefox. However, this did not return the native dialog if the focus was gone.

With the fix, when the passkeys popup was displayed in KeePassXC, I returned to the browser that had Developer Tools open. Then I focused to the Console tabs input field, and then canceled the KeePassXC dialog. That showed the native dialog when window had the focus again (that might have a long delay if the focus was in the console for a long time). Then it worked as expected.

varjolintu (changes requested):
Update the PR template as requested in https://github.com/keepassxreboot/keepassxc-browser/pull/2810#issuecomment-4714559134, and solve the branch conflicts.

a2kolbasov:
> Please update the PR template to the latest version, including this information (do this to all of your PRs): https://github.com/keepassxreboot/keepassxc-browser/pull/3001/changes

> Update the PR template as requested in [#2810 (comment)](https://github.com/keepassxreboot/keepassxc-browser/pull/2810#issuecomment-4714559134), and solve the branch conflicts.

Done

a2kolbasov:
> Do you think this kind of "code-shaming" is going to get your PR reviewed and merged?

I think it's important to be consistent. There are two options for implementing passkeys support:

1. Wait about 10 years for browsers (perhaps) to add an API for integrating extensions with the Credential Management API and the built‑in password manager.
2. Implement a workaround so that the browser doesn't interfere.

You've chosen the second approach.
https://github.com/keepassxreboot/keepassxc-browser/blob/5ae0da9b45c16ce27092f5830b30c89cbe6e0174/keepassxc-browser/content/passkeys.js#L234-L238

Is it good or bad? I think it's good. However, this method has consequences. From the browser's perspective, `passkeys.js` is part of the page. The user clicks a button, and a script on the page starts doing something. Then the user switches to another app. And then **the website** (`passkeys.js`) asks the browser to provide the keys.

Even if rejecting such a request weren't part of the specification but rather a decision made by the developers of a specific browser, I'd still be in agreement with the browser developers. It's a browser extension, not the other way around!

******

> but against the method itself that is possibly required to get passkeys to work properly.

I proposed a patch not for the browser but for the method of implementing passkeys support used in KPXC. Yes, it's a workaround (a hack), but within the current architecture - because supporting this feature is more important. The main thing is that it shouldn't introduce any security risks.

Your response is "report to Mozilla". This would mean returning to the question of waiting for APIs. Even if this bug isn't limited to Linux and Firefox alone, it should still just not work.

That's when I decided to cite code from another file:

******

> FYI: the meaning and usage of Predefined Sites is a totally different thing.

You accepted a code change for the sake of a single website. And then again, and again. You didn't say "report to Google, Epic Games, ProtonMail... Let them create proper input fields". You've accepted fixes many times. But the website design can change at any moment, and the selector will stop working.

https://github.com/keepassxreboot/keepassxc-browser/blob/f66c0cb148bc5c044240cecef2c250b759562cab/keepassxc-browser/common/sites.js#L104

*(This can't be expected to work forever)*

Let's assume this bug only affected Firefox. And why, even in this case, is it not possible to accept the patch? The answer is: because.

******

> this kind of "code-shaming"

A hack by any other name is still a hack. And if it's allowed to do this, then... it's allowed to do this! There are parts of the specification that are problematic to implement in a browser extension due to limitations. There are restrictions imposed by individual browser vendors (especially monopolistic ones). For now, it either works this way or it doesn't work at all - though a third option may become available in the future.

varjolintu:
> You accepted a code change for the sake of a single website. And then again, and again. You didn't say "report to Google, Epic Games, ProtonMail... Let them create proper input fields". You've accepted fixes many times. But the website design can change at any moment, and the selector will stop working.

These are compatibility fixes that would be ages to be fixed by a browser, or a site. Usually sites do not fix them even if you report they are not using standards. But this is a totally different thing when compared to this "hack" with the passkeys. I fail to undestand what is your point bringing this up again and again. Using the compatibility fixes as an example of ugly code is just rude. I kindly suggest you to behave.
 
> A hack by any other name is still a hack. And if it's allowed to do this, then... it's allowed to do this! There are parts of the specification that are problematic to implement in a browser extension due to limitations. There are restrictions imposed by individual browser vendors (especially monopolistic ones). For now, it either works this way or it doesn't work at all - though a third option may become available in the future.

I repeat:
The "ugly" here is not a personal attack against your code, but against the method itself that is possibly required to get passkeys to work properly.

varjolintu (approved):

Issue #2808, opened 2025-12-25
Can't add a Passkey on facebook


yashikada:
### Have you searched for an existing issue?

- [x] Yes, I tried searching and reviewed the pinned issues

### Brief Summary

I can't add a passkey to facebook account.

### Expected Versus Actual Behavior

The actual behavior is this:
I tried on Firefox Linux/Windows the latest version and the button "Create Passkey" is disabled.
I tried on Brave Linux/Windows the latest version and I got this error: `Error!Origin and RP ID do not match`

The expected behavior: add passkey to facebook account.

### Steps to Reproduce

1. Login facebook
2. Goto settings and searching password in search settings
3. Click Password and Security
4. Click Passkey and follow the procedure

### KeePassXC-Browser Debug Information

```Text
I tested on 4 browser these are the informations:

Brave:

KeePassXC - 2.7.10
KeePassXC-Browser - 1.9.11
Operating system: Linux x86_64
Browser: Chrome/Chromium 143.0.0.0

KeePassXC - 2.7.11
KeePassXC-Browser - 1.9.11
Operating system: Win64
Browser: Chrome/Chromium 143.0.0.0

Firefox:
KeePassXC - 2.7.10
KeePassXC-Browser - 1.9.11
Operating system: Linux x86_64
Browser: Mozilla Firefox 146.0

KeePassXC - 2.7.11
KeePassXC-Browser - 1.9.11
Operating system: Win64
Browser: Mozilla Firefox 146.0
```

varjolintu:
It's a known issue. For me Facebook still restricts passkeys to mobile only, so I cannot even test this. The reports I've seen before indicates that Facebook is making passkey stuff from inside an iframe where the URL origin differs from the RP ID. This is Facebook doing non-standard things.

varjolintu:
Checked again, and the feature is now enabled for me. The creation fails because no cross-origin registration is allowed. I couldn't inspect the actual request, but it's only logical that the probably use facebook.com as the RP ID but the passkey creation page is under https://accountscenter.facebook.com.

I tried this with Firefox and the Create button was enabled for me for a couple of tries, but after that it remained in disabled state.

yashikada:
> I tried this with Firefox and the Create button was enabled for me for a couple of tries, but after that it remained in disabled state.

This happened to me on Windows too, but on Linux it remains disabled. I think it's because Windows Security tries to create the passkey as a failback.

> It's a known issue.

[...]

PR #2828, opened 2026-01-17
Add support for Related Origin Requests with passkeys

varjolintu:
Adds support for Related Origin Requests for passkeys. The specification for this is available here: https://www.w3.org/TR/webauthn-3/#sctn-related-origins

In short, the RP ID domain is requested for a list of related origins that can be used when registering a new passkey. For example Facebook and Microsoft is using this feature, and #2808 depends on this. KeePassXC side implementation is also needed for this to actually work. The Related Origins list is always downloaded from ` https://<RP_ID>/.well-known/webauthn`.

The change also introduces a static timeout for the `AbortSignal` so the web requests will timeout after five seconds. Because of this the minimum browser versions must be upgraded.

Related KeePassXC side PR: https://github.com/keepassxreboot/keepassxc/pull/12971

## Testing strategy
For now the only way to check the related origins array is when a new passkeys is created in Facebook or Microsoft account page. A breakpoint must be set to the `keepass.passkeysRegister()` for checking the value. Actual validation for the labels will be implemented to KeePassXC side.

At least the following lists are found:
https://login.microsoftonline.com/.well-known/webauthn
https://accounts.meta.com/.well-known/webauthn
https://www.amazon.de/.well-known/webauthn

## Type of change

### DIFF
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -630,7 +630,7 @@ keepass.passkeysRegister = async function(tab, args = []) {
 
         const messageData = {
             action: kpAction,
-            publicKey: JSON.parse(JSON.stringify(publicKey)),
+            publicKey: passkeyPublicKey,
             origin: origin,
             relatedOrigins: relatedOrigins,
             groupName: page?.settings?.defaultPasskeyGroup,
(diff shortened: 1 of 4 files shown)


varjolintu:
Possibly the maximum amount of labels needs to be increased. For example Amazon has 56 labels in the list:
https://www.amazon.de/.well-known/webauthn

varjolintu:
This is also needed for authentication.

keepassxc-browser/background/keepass.js
```diff
@@ -2,20 +2,23 @@
 
 const keepass = {};
 keepass.associated = { 'value': false, 'hash': null };
```
varjolintu, line 4: Just made some alphabetical ordering here for the properties. Some Prettier is also used for functions that went over the max length.

droidmonkey:
Since this bumps minimum browser version, this might be a change for 1.10.0 instead of a minor release.

droidmonkey (approved):

PR #2916, opened 2026-03-20
Fixes some passkeys errors

### LINKED ISSUE #2915: There is no check on `rpId` to ensure it's a hostname

### Brief Summary

The `rpId` value isn't verified anywhere before the request is sent. As a result, the request can be sent to a different port and path. Naturally, the passkey won't be created after this. However, the request itself will be executed, which is a possible method for "analytics" (user tracking).

https://github.com/keepassxreboot/keepassxc-browser/blob/eab53c6a90f578ef169e7abfce91a6abc9594b1d/keepassxc-browser/background/keepass.js#L923

_No response_

### Steps to Reproduce

1. Go to <https://ipfs.io/ipfs/bafybeib2yc2elnlw4xmhxiwsjm2iu2l5jtz3nf3hxvoiuj6ueukeu6d7iy/> (or use [archive](https://github.com/user-attachments/files/26130311/kpxc-ror.zip) for localhost server)
2. Change `ntfy` server and/or channel if needed
3. Open "Read messages on" link in another tab/browser/device
4. Send a message. The script will call `navigator.credentials.create` with `rpid` like `server:port/path?search&` to call a webhook.

```Text

```

a2kolbasov:

1. `length` - the length in elements, `byteLength` - the length in bytes. Also `length` does not exist in `ArrayBuffer`. For `ArrayBuffer` validation is `undefined < 16`. But `BigUint64Array` requires 128 bytes to pass validation.

    https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray
    https://www.w3.org/TR/webauthn-2/#sctn-cryptographic-challenges

2. `rp.id` in PublicKeyCredentialCreationOptions, but `rpId` in PublicKeyCredentialRequestOptions

    https://www.w3.org/TR/webauthn-2/#idl-index

3. strict ROR validation
    - Prevent tracking and non-domain requests
    - Checking response status code

    https://www.w3.org/TR/webauthn-3/#sctn-validating-relation-origin
4. Fix missing function [`preventSilentAccess`](https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/preventSilentAccess)

------

Fixes #2915

## Screenshots or videos

## Testing strategy

1. Manually via debugger
2. Register and authenticate on <https://webauthn.io/>

## Type of change


varjolintu:
Validation of Related Origins is made in the KeePassXC side.

keepassxc-browser/content/passkeys-utils.js
```diff
@@ -23,7 +23,7 @@ const checkErrors = function(pkOptions, sameOriginWithAncestors) {
         throw new DOMException('Cross-origin register or authentication is not allowed.', DOMException.NotAllowedError);
     }
 
-    if (pkOptions.challenge.length < 16) {
+    if (!Number.isSafeInteger(pkOptions.challenge?.byteLength) || pkOptions.challenge.byteLength < 16) {
```
varjolintu, line 26: `pkOptions.challenge.byteLength` would be enough here.

keepassxc-browser/background/keepass.js
```diff
@@ -920,11 +920,30 @@ keepass.getPasskeysRelatedOrigins = async function(rpId) {
...
     try {
+        let hostname;
+        try {
+            hostname = new URL(`https://${rpId}`).hostname;
+        } catch { }
+
+        if (hostname !== rpId) {
+            logError(`getRelatedOrigins error: "${rpId}" is wrong rpId`);
+            return [];
+        }
```
droidmonkey, line 931: This assumes same subdomain between current hostname and the rpid. We could run into problems with `www` or some other subdomain that might be an alias to the original registration. Agree?

Overall I think it might be good to add a per-site option to ignore this check for edge cases like that.

droidmonkey (changes requested):

a2kolbasov:
> Validation of Related Origins is made in the KeePassXC side.

I'm not checking ROR. I'm checking the data for a GET request the same way you check the `JSON<string[]>` compliance of the response. (https://github.com/keepassxreboot/keepassxc-browser/issues/2915#issuecomment-4099182386)

keepassxc-browser/content/passkeys-utils.js
a2kolbasov: I want to make sure `challenge` exists and is the correct type. That it's not a base64string or something like that. So that there's no more `undefined < 16` check.

keepassxc-browser/background/keepass.js
a2kolbasov: > This assumes same subdomain between current hostname and the rpid. We could run into problems with `www` or some other subdomain that might be an alias to the original registration. Agree?

No, I'm not checking the current hostname (from `location.hostname`), I'm checking the `rpId` provided by the site for the "domain name" definition. `URL` is a parser for any input value, not just the current site's URL.

```javascript
new URL("https://" + "site.example").hostname === "site.example"
new URL("https://" + "www.site.example").hostname === "www.site.example"
new URL("https://" + "site.example:8443").hostname === "site.example" // !
new URL("https://" + "site.example/path").hostname === "site.example" // !
```

> Overall I think it might be good to add a per-site option to ignore this check for edge cases like that.

This is likely a specification violation. The site itself specifies its desired `rpId`. And we also have `location.domain`/`location.origin`.

varjolintu: As I said in the issue thread, RP ID is set by the web site itself. If that is in invalid form, it doesn't pass KeePassXC's validation, and the try/catch here will trigger if the fetch fails.

We've tried to kept the validation side in the extension very minimal, and KeePassXC is where the full validation happens.

EDIT: See https://github.com/keepassxreboot/keepassxc/blob/develop/src/browser/PasskeyUtils.cpp#L107

a2kolbasov: You're talking about checking the request result, but I'm talking about the request itself.

This allows the site to "probing" the URL not on behalf of the current site (page context), but on behalf of the add-on. The background script is the same for private windows, regular windows, and all Firefox containers. If I enabled extension debugging in the network tab and made several requests, the browser reported that they were cached. I also added the `cache: 'no-store'` option, but the current behavior still confuses me. The specification also implies that the rpId is the domain name, not something else.

keepassxc-browser/background/keepass.js
```diff
@@ -662,7 +662,7 @@ keepass.passkeysGet = async function(tab, args = []) {
         const nonce = keepassClient.getNonce();
         const [ publicKey, origin ] = args;
         const passkeyPublicKey = JSON.parse(JSON.stringify(publicKey));
-        const relatedOrigins = await keepass.getPasskeysRelatedOrigins(passkeyPublicKey?.rp?.id);
+        const relatedOrigins = await keepass.getPasskeysRelatedOrigins(passkeyPublicKey?.rpId);
```
varjolintu, line 665: This is incorrect. The original `rp.id` is the correct one. See:
Image

a2kolbasov: 1. It's not clear from the screenshot where exactly you set the breakpoint. No function name, no line number(s), no `kpAction`...
2. The screenshot shows a `user` property, which contains an object with a `displayName` property. This is <https://www.w3.org/TR/webauthn-2/#dictionary-makecredentialoptions>, not <https://www.w3.org/TR/webauthn-2/#dictionary-assertion-options>.
3. Here's what **you** added earlier: <https://github.com/keepassxreboot/keepassxc-browser/blame/1.10.0.1/keepassxc-browser/content/passkeys-utils.js#L112>
4. Also I set a brakepoint on <https://github.com/keepassxreboot/keepassxc-browser/blob/1.10.0.1/keepassxc-browser/content/passkeys.js#L203> and got this:

    webauthn.io AUTHENTICATION OPTIONS
    "webauthn.io" does not violate the specification

keepassxc-browser/background/keepass.js
```diff
@@ -920,11 +920,30 @@ keepass.getPasskeysRelatedOrigins = async function(rpId) {
     }
 
     try {
+        let hostname;
+        try {
+            hostname = new URL(`https://${rpId}`).hostname;
+        } catch { }
```
varjolintu, line 926: Should return already in here?

a2kolbasov: So there should be code duplication?
```javascript
try {
  if (new URL(`https://${rpId}`).hostname !== rpId) {
    logError(`getRelatedOrigins error: "${rpId}" is wrong rpId`);
    return [];
  }
} catch {
  logError(`getRelatedOrigins error: "${rpId}" is wrong rpId`);
  return [];
}
```

keepassxc-browser/content/passkeys-utils.js
a2kolbasov: Does your silence mean that I should refuse this additional check anyway?

keepassxc-browser/background/keepass.js
varjolintu: Of course we want to handle the exception.

keepassxc-browser/content/passkeys-utils.js
varjolintu: My silence means I haven't answered you yet. `if (pkOptions.challenge.byteLength < 16)` is enough here.

a2kolbasov: Done

keepassxc-browser/background/keepass.js
a2kolbasov: The error itself does not provide any useful information. The `rpId` may be incorrect both in case of an error and without one.

If an error occurs, the assignment operation will not be executed, and the `hostname` will be equal to `undefined`.

PR #2926, opened 2026-03-24
Russian localization fix

11DFF:
Simple fix in Russian localization ("or" translated to "или" on line 1298).
Current translation (before fix):
Um7UoHkIro


varjolintu:
Translations are handled in Transifex.

https://github.com/keepassxreboot/keepassxc-browser/blob/develop/.github/CONTRIBUTING.md#translations

PR #2934, opened 2026-03-26
Load non-discarded tabs only

tanriol:
One may have up to a few thousand (at least) discarded tabs in a persistent browser session. Trying to send messages to all of them on the clear credentials timer can introduce noticeable jank on every tab switch.

I'm not sure whether such a targeted fix is the right approach or some more significant refactoring should be done. For example, no documentation I know says anything about tab IDs being sequentially assigned, so using an array for `page.tabs` looks a bit questionable... maybe a `Map` would be a better choice here.

## Testing strategy
Minimally tested in my main session (3+ thousand tabs). Filling fields from context menu works, and Firefox Profiler confirms the extension process freezes have been reduced significantly.

## Type of change

### DIFF
--- a/keepassxc-browser/background/init.js
+++ b/keepassxc-browser/background/init.js
@@ -70,6 +70,10 @@ const initListeners = async function() {
      * @param {object} tab
      */
     browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+        // Could be not tracked yet if discarded from browser startup
+        if (tab?.id && !page.tabs[tab.id]) {
+            page.createTabEntry(tab.id);
+        }
         // If the tab URL has changed (e.g. logged in) clear credentials
         if (changeInfo.url) {
             page.clearLogins(tabId);
@@ -77,9 +81,6 @@ const initListeners = async function() {
 
         if (changeInfo.status === 'complete' && tab?.id) {
             browserAction.showDefault(tab);
-            if (!page.tabs[tab.id]) {
-                page.createTabEntry(tab.id);
-            }
         }
     });
 
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -115,7 +115,7 @@ page.initSettings = async function() {
 
 page.initOpenedTabs = async function() {
     try {
-        const tabs = await browser.tabs.query({});
+        const tabs = await browser.tabs.query({ discarded: false });
         for (const i of tabs) {
             page.createTabEntry(i.id);
         }


keepassxc-browser/background/init.js
```diff
@@ -70,16 +70,17 @@ const initListeners = async function() {
      * @param {object} tab
      */
     browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+        // Could be not tracked yet if discarded from browser startup
+        if (!page.tabs[tab.id]) {
```
varjolintu: `tab.id` should be checked just in case:
```
if (tab?.id && !page.tabs[tab.id]) { ...
```

varjolintu:
> maybe a Map would be a better choice here.

Good idea. I'll add this to my refactor TODO list :)

keepassxc-browser/background/init.js
tanriol: Fixed, thank you!

varjolintu:
If you can easily test this, could you try one thing? I was looking at the code and probably it could be enough to use the following code at the beginning of `page.switchTab()`:

```
page.switchTab = async function(tab) {
    if (tab?.discarded) {
        return;
    }
```

Then the clear credential timeout should trigger only to non-discarded tabs.

tanriol:
> Then the clear credential timeout should trigger only to non-discarded tabs.

I don't think so. A check in `switchTab` would check the tab that's being created or activated, and that seems the wrong thing to do here - if the tab is being created or activated during a session, it's likely to either be already loaded or get loaded soon.

The problem here is not that the timer happens too often, but that it does too much heavy work - the `clearCredentialsTimeout` handler not only loops over all tab records in the array (that seems fine at this scale), but calls `clearCredentials` for every tab index with `complete: true`, which in turn calls `sendMessage` and `contextMenus.update` for every tab including discarded, causing two IPC calls from extension process to main process per tab.

varjolintu:
Got it, thanks! It's just that the `page.clearCredentials()` shouldn't be called if `page.switchTab()` is returned as early as possible (in my suggestion) if the tab is discarded.

Anyway, this discussion has a few good ideas for a refactor.

tanriol:
> It's just that the `page.clearCredentials()` shouldn't be called if `page.switchTab()` is returned as early as possible (in my suggestion) if the tab is discarded.

`clearCredentialsTimeout` is called (by design) almost every time I switch from one tab to another even if neither is discarded, which happens quite often during regular browsing.

In my current understanding, ideally the extension shouldn't be sending these messages to all background tabs. Instead it should remember for each tab when it last went into background and send the `clear_credentials` message to individual tabs as their specific timeouts expire, but that'd be a bigger refactoring.

For a side note, calling `setFillAttributeContextMenuItemVisible` from `clearLogins` for background tabs sounds wrong, but that again requires a more careful refactoring. If you want I can look into these refactors.

varjolintu:
I have a pretty good idea for the refactor, thanks :)

varjolintu (approved):

PR #2944, opened 2026-04-10
Refactor tab handling

varjolintu:
Insipired by recent changes in: https://github.com/keepassxreboot/keepassxc-browser/pull/2934. Refactors how tabs are handled in the background scripts, and makes the credential clearing more simple and efficient.

## Changes in files

A new file `background/tabs.js` is created. Tab related event listeners from `init.js`, and tab related functions and variables from `page.js` are moved to this new file.

## Changes in the implementation

Instead of relying a basic JavaScript array, tabs are now stored inside a `Map()`. When retrieving a single tab object, the old direct access to object `page.tabs[tab.id]` is now replaced with `tabs.getTabFromId()`. When updating values to a tab object, or deleting it, functions in `tabs` are used instead of accessing the list directly.

The new `tabs` file also has a second Map `tabs.timeoutList` that stores a clear credential timeout to it when a tab is switched. The list stores switched tab's ID and the clear credential timeout ID to it. If the tab is activated again, the timeout is cleared. Otherwise the tab's credentials are cleared. Even if a previous timeout is already triggered and user switches to another tab, a new timeout is added to the list with that tab ID.

Previously the timeout cleared credentials from all other tabs except the active one, causing a lot of calls, especially if there were a huge amount of tabs present.

## Summary

In short, the old implementation worked like this:
- All tab related functionality was splitted between `init.js` and `page.js` (those files have other purposes).
- Credentials were always cleared from all tabs, even if those were already inactive.

The new implementation:
- All tab related functionality is now moved to the new `tabs.js` file.
- Credentials are always cleared from the tabs that user leaves to the background when changing to a new active tab.

## Testing strategy
Only way to verify how this works is doing it manually using breakpoints. Automatic tests needs to be added for this later, if possible. The tests will need some refactoring too, so I'll investigate it with those changes.

## Type of change

### DIFF
--- a/keepassxc-browser/background/browserAction.js
+++ b/keepassxc-browser/background/browserAction.js
@@ -18,10 +18,11 @@ browserAction.show = async function(tab, popupData) {
         });
 
         let badgeText = '';
+        const currentTab = tabs.getTabFromId(tab.id);
         if (popupData.popup === 'popup_login') {
-            badgeText = page.tabs[tab.id]?.loginList?.length;
+            badgeText = currentTab?.loginList.length;
         } else if (popupData.popup === 'popup_httpauth') {
-            badgeText = page.tabs[tab.id]?.loginList?.logins?.length;
+            badgeText = currentTab?.loginList?.logins?.length;
         }
 
         browserAction.setBadgeText(tab?.id, badgeText);
@@ -52,10 +53,11 @@ browserAction.showDefault = async function(tab) {
         return;
     }
 
-    if (page?.tabs[tab.id]?.loginList.length > 0) {
+    const currentTab = tabs.getTabFromId(tab?.id);
+    if (currentTab?.loginList.length > 0) {
         popupData.iconType = 'normal';
         popupData.popup = 'popup_login';
-        browserAction.setBadgeText(tab?.id, page.tabs[tab.id]?.loginList.length);
+        browserAction.setBadgeText(tab?.id, currentTab?.loginList.length);
     }
 
     await browserAction.show(tab, popupData);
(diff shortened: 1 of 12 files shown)


keepassxc-browser/background/event.js
```diff
@@ -174,20 +175,16 @@ kpxcEvent.onHTTPAuthPopup = async function(tab, data) {
         popup: 'popup_httpauth'
     };
 
-    page.tabs[tab.id].loginList = data;
+    tabs.updateTabValues(tab?.id, { loginList: data });
```
droidmonkey: this data element has a bunch of sub-elements that are not well defined anywhere in the tabs interface.

keepassxc-browser/background/browserAction.js
```diff
@@ -18,10 +18,11 @@ browserAction.show = async function(tab, popupData) {
         });
 
         let badgeText = '';
+        const currentTab = tabs.getTabFromId(tab.id);
         if (popupData.popup === 'popup_login') {
-            badgeText = page.tabs[tab.id]?.loginList?.length;
+            badgeText = currentTab?.loginList.length;
         } else if (popupData.popup === 'popup_httpauth') {
-            badgeText = page.tabs[tab.id]?.loginList?.logins?.length;
+            badgeText = currentTab?.loginList?.logins?.length;
```
droidmonkey: `loginList.logins` comes from the http data package, its rather specific to httpauth and we might want to either encapsulate those details in a function or spread them into already defined data elements.

droidmonkey (approved):
This functionally works, but we have an opportunity to cleanup the httpauth data structure a bit

keepassxc-browser/background/browserAction.js
varjolintu: I made a change where the logins is now a separate `basicAuthLogins` object inside the tab object. It was confusing that `loginList` was an array, but HTTP Basic Auth reused it as an object. Good find.

droidmonkey (approved):

PR #2948, opened 2026-04-14
Fix input field detection when label contains overlaying elements

tiboroche:
Allow input field detection when the topmost element is a child of the label (e.g., span overlay in Material Design labels). This fixes detection on pages using <label><input/><span>Label</span></label> pattern where the span is positioned absolutely and blocks elementFromPoint.

This change allows to properly detect login and password fields on login pages based on Material Design, like the one my company uses. 

HTML snippet with the fields not properly detected: 
```
<form class="loginFormInputs" method="post" action="https://[REDACTED]" autocomplete="off" onsubmit="login.disabled = true; return true;">

            <label class="pure-material-textfield-outlined">
              <input type="text" id="username" name="username" placeholder="" autocorrect="off" required autofocus>
              <span>Login</span>
            </label>

            <label class="pure-material-textfield-outlined passwordField">
              <input type="password" id="password" name="password" placeholder="" required>
              <span>Password</span>
            </label>
</form>
```

## Screenshots or videos
image

## Testing strategy

Tested on the login page of my company and it detects the fields properly. I also tried a few other websites and detection was still working. 

## Type of change

### DIFF
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -507,7 +507,8 @@ kpxcFields.isTopElement = function(elem, rect) {
     const getTopmostElement = (element, x, elementRect) => {
         const topElement = rootNode.elementFromPoint(x, elementRect.top + (elementRect.height / 2));
         return element?.labels &&
-            (element.labels[0] === topElement || element.labels[0].contains(topElement)) &&
+            (element.labels[0] === topElement ||
+                (element.labels[0] instanceof Node && element.labels[0].contains(topElement))) &&
             elementsOverlap(elementRect, topElement.getBoundingClientRect())
             ? element
             : topElement;
--- a/keepassxc-browser/content/fields.js
+++ b/keepassxc-browser/content/fields.js
@@ -506,9 +506,8 @@ kpxcFields.isTopElement = function(elem, rect) {
     // Also allows if the topmost element is a child of the label (e.g., span inside label).
     const getTopmostElement = (element, x, elementRect) => {
         const topElement = rootNode.elementFromPoint(x, elementRect.top + (elementRect.height / 2));
-        return element?.labels &&
-            (element.labels[0] === topElement ||
-                (element.labels[0] instanceof Node && element.labels[0].contains(topElement))) &&
+        return element?.labels instanceof NodeList &&
+            element.labels[0]?.contains(topElement) &&
             elementsOverlap(elementRect, topElement.getBoundingClientRect())
             ? element
             : topElement;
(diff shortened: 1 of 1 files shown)


droidmonkey (approved):

keepassxc-browser/content/fields.js
```diff
@@ -503,10 +503,11 @@ kpxcFields.isTopElement = function(elem, rect) {
 
     // Returns the topmost element from point x, height/2
     // If the input has a label as the top element and it's inside the input, allow it.
+    // Also allows if the topmost element is a child of the label (e.g., span inside label).
     const getTopmostElement = (element, x, elementRect) => {
         const topElement = rootNode.elementFromPoint(x, elementRect.top + (elementRect.height / 2));
         return element?.labels &&
-            element.labels[0] === topElement &&
+            (element.labels[0] === topElement || element.labels[0].contains(topElement)) &&
```
varjolintu: You are missing checks here and the code throws an exception. It should be checked is the `element.labels` is actually an array and it exists. This is why the tests fail. Run them locally with `npm run tests` to verify those will pass.

varjolintu (changes requested):

keepassxc-browser/content/fields.js
tiboroche: Thank you for your feedback, I added a check in my latest commit. 

elements.labels[0] was already accessed before my commit so I feel it is safe to access it, and elements?.labels in the first part of the test ensures its existence. 

Let me know if it is ok for you.

keepassxc-browser/content/fields.js
```diff
@@ -507,7 +507,8 @@ kpxcFields.isTopElement = function(elem, rect) {
     const getTopmostElement = (element, x, elementRect) => {
         const topElement = rootNode.elementFromPoint(x, elementRect.top + (elementRect.height / 2));
         return element?.labels &&
-            (element.labels[0] === topElement || element.labels[0].contains(topElement)) &&
+            (element.labels[0] === topElement ||
+                (element.labels[0] instanceof Node && element.labels[0].contains(topElement))) &&
```
varjolintu: It should be enough to do:
`(Array.isArray(element?.labels) && element.labels[0]?.contains(topElement))` etc. The `isArray()` also checks for undefined.

tiboroche: It is not possible to use isArray in this case as labels is a NodeList and not an array. I have simplified the test in my latest commit, we can only use contains as a node contains itself, and with the ? notation we have enough checks.

varjolintu: Looks fine :) I'll test it.

tiboroche: @varjolintu have you had time to have look ? it is OK for you ?

varjolintu: I'll have time when I have time.

varjolintu (approved):
The HTML snipped you provided did not work for testing (without the fix fields were still identified normally), but in theory this should be working correctly. Thanks!

PR #2954, opened 2026-04-21
Fix filling credentials with readonly username

### LINKED ISSUE #2953: Password Manager avoidance (looks like) on MyChart (used by medical organizations)

### Brief Summary

So there's an interface for patients called MyChart (here's an example one https://my.uclahealth.org)

It looks like recently they introduced a mechanism to thwart auto filling credentials. Basically now in order to log in you have to enter the username, click "Continue" then enter password and only then you can click "Log in". This seems variation of logging in over two pages, except everything happens on a single one.

I tried to use the "Username-only" detection, or also marking the login/password fields, but couldn't get it to work.

Being able to log in.

### Steps to Reproduce

1. Add credentials to KeePassXC (I don't think one needs to have real one to test it)
2. Try to log in with them

```Text
Did not see anything valuable, even after enabling debugging.
```

varjolintu:
In some cases where a password fill is expected, username input field changed to readonly, and the Username Icon is not detected for the password input. Filling from the icon or using popup from the extension toolbar doesn't work. 

This fix adds a workaround for the previously unidentified password input to the combination, and performs the fill anyway with both scenarios.  Actual detection problem of the password field detection needs a lot more work, and is done separately because many other sites have the same issue.

Refactored some of the code, moving additional checks made during the fill to a separate functions.

* Fixes #2953

## Testing strategy
Manually using https://my.uclahealth.org/. No actual account needed to test the site.

## Type of change

### DIFF
--- a/keepassxc-browser/content/fill.js
+++ b/keepassxc-browser/content/fill.js
@@ -245,16 +245,9 @@ kpxcFill.fillInCredentials = async function(combination, predefinedUsername, uui
         skipAutoSubmit = selectedCredentials.skipAutoSubmit === 'true';
     }
 
-    // Update the password field if form has been updated with a new identical element (Moodle).
-    // If found, replace the new input field to the combination.
-    if (combination?.form && combination?.password && !combination.form.contains(combination.password)) {
-        const formInputs = kpxcObserverHelper.getInputs(combination.form);
-        const newPasswordField = formInputs?.find((formInput) =>
-            formInput?.getLowerCaseAttribute('type') === 'password');
-        if (newPasswordField && areNamedNodeMapsEqual(combination.password.attributes, newPasswordField.attributes)) {
-            combination.password = newPasswordField;
-        }
-    }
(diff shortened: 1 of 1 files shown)


keepassxc-browser/content/fill.js
```diff
@@ -473,3 +487,12 @@ const areNamedNodeMapsEqual = function(currentNodeMap, newNodeMap) {
 
     return true;
 };
+
+const getPasswordFieldFromForm = function(combination) {
```
droidmonkey, line 491: Should we include some simple checks to make sure the password field is usable (visible, not read only, etc)

droidmonkey (approved):

keepassxc-browser/content/fill.js
varjolintu: The `kpxcObserverHelper.getInput()` below already does these checks, except for `readOnly`. Lets add the latter there just in case.

PR #2960, opened 2026-04-29
Fix icon placement for two step login scenario where username is hidden

JanSvoboda:
In some two-step login scenarios, where the username is filled during the first step and the password is provided in the next, the icon placement might not work properly if the username input is kept hidden.

This fix checks whether username input is visible, if not then icon is displayed on password input.

## Screenshots or videos
screenshot_keepassxc_20260429

## Testing strategy
Manually tested on some corporate publicly inaccessible login screen. Currently I do not know of any other site that would have the same behaviour.

## Type of change

### DIFF
--- a/keepassxc-browser/content/icon-handler.js
+++ b/keepassxc-browser/content/icon-handler.js
@@ -53,10 +53,11 @@ kpxcIcons.addIconsFromForm = async function(form) {
                     kpxcIcons.addIcon(c.passwordInputs[0], kpxcIcons.iconTypes.DEFAULT);
                 }
             }
-
-            if (c.username && !c.username.readOnly && kpxcFields.isVisible(c.username)) {
+            const usernameFieldVisible = c.username && kpxcFields.isVisible(c.username);
+            
+            if (c.username && !c.username.readOnly && usernameFieldVisible) {
                 kpxcIcons.addIcon(c.username, kpxcIcons.iconTypes.DEFAULT);
-            } else if (c.password && (!c.username || c.username.readOnly || !kpxcFields.isVisible(c.username))) {
+            } else if (c.password && (!c.username || (c.username && (c.username.readOnly || !usernameFieldVisible)))) {
                 // Single password field, or username field is hidden (e.g. two-step login)
                 kpxcIcons.addIcon(c.password, kpxcIcons.iconTypes.DEFAULT);
             }
--- a/keepassxc-browser/content/icon-handler.js
+++ b/keepassxc-browser/content/icon-handler.js
@@ -54,10 +54,10 @@ kpxcIcons.addIconsFromForm = async function(form) {
                 }
             }
 
-            if (c.username && !c.username.readOnly) {
+            if (c.username && !c.username.readOnly && kpxcFields.isVisible(c.username)) {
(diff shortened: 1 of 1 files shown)


varjolintu:
This repository does not accept AI contributors, or AI-assisted commits. See https://github.com/keepassxreboot/keepassxc-browser/blob/develop/.github/CONTRIBUTING.md#using-ai.

JanSvoboda:
Hello,
is it because the copilot did pull request review? Can I disable that somehow? Not really sure here, as I haven't done any pull request in ages. I noticed that PR name was AI generated - honestly didn't notice that when I was doing the PR.

varjolintu:
For me it showed that Copilot created this pull request. Now it doesn't show that? Strange. The Copilot review stuff is temporary enabled to see if it helps.

JanSvoboda:
ok, thank you, Copilot didn't do that PR, I did, I was afraid that I screwed something up

JanSvoboda:
might I have a question? copilot suggested logical change, if I were to implement it as suggested, how would it be evaluated? as AI assisted or not?

varjolintu:
> might I have a question? copilot suggested logical change, if I were to implement it as suggested, how would it be evaluated? as AI assisted or not?

It would be AI assisted :) I'd rather just take some hints from its suggestions, but not implement anything directly. I'm also thinking disabling Copilot here. Sometimes its suggestions are pretty random.

keepassxc-browser/content/icon-handler.js
```diff
@@ -54,10 +54,10 @@ kpxcIcons.addIconsFromForm = async function(form) {
                 }
             }
 
-            if (c.username && !c.username.readOnly) {
+            if (c.username && !c.username.readOnly && kpxcFields.isVisible(c.username)) {
```
varjolintu: Use:
`const usernameFieldVisible = c.username && kpxcFields.isVisible(c.username);`

keepassxc-browser/content/icon-handler.js
```diff
@@ -54,10 +54,10 @@ kpxcIcons.addIconsFromForm = async function(form) {
...
                 kpxcIcons.addIcon(c.username, kpxcIcons.iconTypes.DEFAULT);
-            } else if (c.password && (!c.username || (c.username && c.username.readOnly))) {
-                // Single password field
+            } else if (c.password && (!c.username || c.username.readOnly || !kpxcFields.isVisible(c.username))) {
```
varjolintu: Shouldn't this be:
`} else if (c.password && (!c.username || (c.username && (c.username.readOnly || !usernameFieldVisible)))) {`

So the single password field applies if:
- Password field is defined
- Usename field is not found OR
- Username field is found, but is readOnly or not visible

varjolintu:
I would need some actual site where the error happens and I can confirm the fix works.

JanSvoboda:
I reflected your suggestions. I can provide you with some website, however I would rather avoid posting it here, if possible, is there some way how can I send it to you directly?

varjolintu:
> I reflected your suggestions. I can provide you with some website, however I would rather avoid posting it here, if possible, is there some way how can I send it to you directly?

Private message in Matrix, or via email.

varjolintu (approved):

PR #2969, opened 2026-05-07
Fix managed settings not being applied on Chromium

### LINKED ISSUE #2974: Managed settings not applied in Chrome

### Brief Summary

I am trying to deploy the KeePassXC browser with predefined settings (enabling passkey) in Chrome. Under chrome://policy/, the setting is recognized, but it remains disabled in the browser extension.

Image
Image

I try to debug the behavior. If i change in "keepassxc-browser\background\page.js"...

**From:**
```
page.initSettings = async function() {
    ...
    else if (!page.isSafari && typeof chrome.storage.managed === 'object') {
        chrome.storage.managed.get('settings').then((managedSettings) => {
            if (managedSettings?.settings) {
                debugLogMessage('Managed settings found.');
                item.settings = managedSettings.settings;
            }
        }).catch((err) => {
            debugLogMessage('page.initSettings: ' + err);
        });
    }
    ...
```

**To:**
```
page.initSettings = async function() {
    ...
    else if (!page.isSafari && typeof chrome.storage.managed === 'object') {
        try {
            const managedSettings = await chrome.storage.managed.get('settings');
            if (managedSettings?.settings) {
                debugLogMessage('Managed settings found.');
                item.settings = managedSettings.settings;
            }
        } catch (err) {
            debugLogMessage('page.initSettings: ' + err);
        }
    }
    ...
```
...it works.

_No response_

### Steps to Reproduce

- Run CMD as administrator.
- Execute the following command to add the policy
`reg.exe ADD "HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Google\Chrome\3rdparty\extensions\oboonakemofpalcgghocfoadofidjkkk\policy" /v "settings" /t REG_SZ /d "{\"passkeys\":true}"`
- Start Chrome and check if the Passkeys option is still disabled

```Text
KeePassXC - 2.7.12
KeePassXC-Browser - 1.10.1
Operating system: Win64
Browser: Chrome/Chromium 148.0.0.0
```

RGBCube:
The Chromium branch of `page.initSettings()` had `chrome.storage.managed.get('settings').then(...)` that set `page.settings = item.settings`, but the surrounding code never awaited the promise, so `page.settings = item.settings` could randomly not use the managed storage.

Switch the Chromium branch to await + try/catch (matching the Firefox branch) and then merge the two branches into a single block.

* Fixes #2974

### DIFF
--- a/keepassxc-browser/background/page.js
+++ b/keepassxc-browser/background/page.js
@@ -75,9 +75,10 @@ page.initSettings = async function() {
         const item = await browser.storage.local.get({ 'settings': {} });
 
         // Load managed settings if found
-        if (page.isFirefox && typeof(browser.storage.managed) === 'object') {
+        const managedStorage = page.isFirefox ? browser.storage.managed : chrome.storage.managed;
+        if (!page.isSafari && typeof managedStorage === 'object') {
             try {
-                const managedSettings = await browser.storage.managed.get('settings');
+                const managedSettings = await managedStorage.get('settings');
                 if (managedSettings?.settings) {
                     debugLogMessage('Managed settings found.');
                     item.settings = managedSettings.settings;
@@ -85,15 +86,6 @@ page.initSettings = async function() {
             } catch (err) {
                 debugLogMessage('page.initSettings: ' + err);
             }
-        } else if (!page.isSafari && typeof chrome.storage.managed === 'object') {
-            chrome.storage.managed.get('settings').then((managedSettings) => {
-                if (managedSettings?.settings) {
-                    debugLogMessage('Managed settings found.');
-                    item.settings = managedSettings.settings;
-                }
-            }).catch((err) => {
-                debugLogMessage('page.initSettings: ' + err);
-            });
         }
 
         page.settings = item.settings;


varjolintu:
Are you sure this works? I remember I had explicitly add a different handling for Chrome/Chromium because the default method didn't work.

RGBCube:
Yeah, the policy isn't getting applied at all for me even though `await chrome.storage.managed.get(null)` returns the values I've configured.

Also tested these:

```js
  console.log('start');
  chrome.storage.managed.get('settings').then(m => console.log('then ran with:', m));
  console.log('after .then call (this prints first - that is the bug)');

> start
> after .then call (this prints first - that is the bug)
undefined // this is the return value of the expr
> then ran with: {settings: {…}} // printed after
```

And:

```js
  (async () => {
      const before = (await chrome.storage.local.get({ settings: {} })).settings;
      const managed = await chrome.storage.managed.get('settings');
      console.log('managed.get returned:', managed);

      if (managed?.settings) {
          await chrome.storage.local.set({ settings: managed.settings });
          console.log('local.settings overwritten with managed values');
      } else {
          console.warn('no managed settings');
      }

      const after = (await chrome.storage.local.get('settings')).settings;
      console.log({ before, after });
  })();
```

Which successfully configured the plugin. Feel free to test my patch out, I tested all of these in [Helium](https://helium.computer/) which is an Ungoogled Chromium-like browser.

RGBCube:
@varjolintu could you test this out? it works for me, would love to get this released

varjolintu:
> @varjolintu could you test this out? it works for me, would love to get this released

I will do that when I have time, just like with every other PR.

varjolintu:
To test easily policy loading with Chromium in Fedora 44 Linux, create a file `/etc/chromium/policies/managed/policies.json` with the following content (replace the extension ID):

```
{
  "3rdparty": {
    "extensions": {
      "pdffhmdngciaglkoonimfcmckehcpafo": {
        "settings": {
          "passkeys": true,
        }
      }
    }
  }
}
```

keepassxc-browser/background/page.js
```diff
@@ -75,25 +75,17 @@ page.initSettings = async function() {
         const item = await browser.storage.local.get({ 'settings': {} });
 
         // Load managed settings if found
-        if (page.isFirefox && typeof(browser.storage.managed) === 'object') {
+        const managedStorage = page.isFirefox ? browser.storage.managed : page.isSafari ? null : chrome.storage.managed;
+        if (typeof managedStorage === 'object') {
```
varjolintu: It would be simpler to have this as: `if (!page.isSafari && typeof managedStorage === 'object')`.

varjolintu (changes requested):
Just a small change, otherwise it looks good. Also reproduced the error that managed settings were correctly read, but not set.

RGBCube:
Fixed

keepassxc-browser/background/page.js
RGBCube: Good catch, that is a lot better

varjolintu (approved):

varjolintu:
Thanks! :)

PR #2973, opened 2026-05-12
Fix use of old TOTP

### LINKED ISSUE #2972: TOTP keeps being filled with the same value

### Brief Summary

TOTP keeps inserting the same value, even though it expired and the page has been reloaded

It would be expected the value would expire after 30seconds and be refetched from keepassxc to get the updated value, instead of reusing the old value causing login attempts to fail.

### Steps to Reproduce

1. Login to website with wrong password
2. on the MFA TOTP page click the fill using keepass totp
3. get the login failed page and click try again, try again with correct password, but you have taking 30+ seconds to do this, and the totp has rotated, so when you get to the TOTP screen the *same* old value is inserted, instead of a fresh value.

```Text

```

jknockel:
Prior to this requested change `getTotp()` checked the `newTotpSupported` `featuresList` field, but this field is always undefined and always evaluates as falsy.  The actual name of the `featuresList` field that is defined is `newTotp`, so now with this change we check that field instead. This prevents old/stale TOTP values from being unnecessarily used.  Fixes #2972.

## Testing strategy
I manually tested on https://myaccount.uscis.gov/sign-in (*not* using login.gov sign-in method).

## Type of change

### DIFF
--- a/keepassxc-browser/background/keepass.js
+++ b/keepassxc-browser/background/keepass.js
@@ -560,7 +560,7 @@ keepass.createNewGroup = async function(tab, args = []) {
 
 keepass.getTotp = async function(tab, args = []) {
     const [ uuid, oldTotp ] = args;
-    if (!keepass.featuresList.newTotpSupported) {
+    if (!keepass.featuresList.newTotp) {
         return oldTotp;
     }
 


droidmonkey (approved):

varjolintu (approved):

varjolintu:
Thanks!

PR #2992, opened 2026-06-01
Update to 1.10.3

varjolintu:
Updates CHANGELOG, manifest files and offline translations for 1.10.3. Adds the new connection method option to managed storage template.

## Type of change
- ✅ Release

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,11 @@
+1.10.3 (2026-06-01)
+=========================
+- Add support for WebSocket connection method (disabled until KeePassXC support) [#2576]
+- Fix creating Credential Banner [#2984]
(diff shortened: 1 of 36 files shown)


droidmonkey (approved):

droidmonkey:
Just wondering, is there any way to store the version number centrally and dynamically inject into the relevant manifest files and such?

varjolintu:
> Just wondering, is there any way to store the version number centrally and dynamically inject into the relevant manifest files and such?

Only if we create a script that updates it. The build script should already do that, but the version numbers should be bumped anyway. It makes using `npm debug:firefox` etc. easier, because the same version number must updated to the repos as well. Otherwise just updating the main manifest.json will leave those files behind. I know, this is no way an optimal solution.

Issue #2594, opened 2025-06-30
NixOS/LibreWolf - Credential Entry Selection immediately disappears after DB-Unlock


becknik:
### Have you searched for an existing issue?

- [x] Yes, I tried searching and reviewed the pinned issues

### Brief Summary

When clicking on the grey icon next to a login field detected by keepassxc-browser and entering my password in the follow-up pop-up, the pop-up for the selection of the right credential entry appears and then immediately disappears after around 500ms. Then, an error appears with the message: "Current database is not connected".

Here's a video (I didn't click on any element, it just disappeared)

https://github.com/user-attachments/assets/6aeeb182-4df4-4098-85d3-b993cd1cb635

This is the log for the attempted credential selection on my computer:

```log
[Error ] KeePassXC-Browser - page.initSettings: Error: Managed storage manifest not found global.js:168:13
[Error ] KeePassXC-Browser - Error: Cannot find menu item with id fill_attribute global.js:168:13
KeePassXC-Browser: Connecting to native messaging host org.keepassxc.keepassxc_browser client.js:383:13
KeePassXC-Browser: Server public key: W8wMjhSpQeeTRri66Nb8V+atVdtMTHjqqsw7WUp7PBo= keepass.js:426:17
[Error ] KeePassXC-Browser - 1: Database not opened global.js:168:13
[Error ] KeePassXC-Browser - 1: Database not opened global.js:168:13
[Error ] KeePassXC-Browser - No content script available for this tab. global.js:168:13
[Error ] KeePassXC-Browser - 1: Database not opened global.js:168:13
[Error ] KeePassXC-Browser - No content script available for this tab. global.js:168:13
[Error ] KeePassXC-Browser - page.initSettings: Error: Managed storage manifest not found 2 global.js:168:13
[Error ] KeePassXC-Browser - page.initSettings: Error: Managed storage manifest not found 2 global.js:168:13
[Debug observer-helper.js:245] KeePassXC-Browser - Input fields found: global.js:160:13
[Debug ] KeePassXC-Browser - Login field combinations identified: global.js:160:13
[Error ] KeePassXC-Browser - page.initSettings: Error: Managed storage manifest not found global.js:168:13
[Error ] KeePassXC-Browser - page.initSettings: Error: Managed storage manifest not found global.js:168:13
[Error ] KeePassXC-Browser - page.initSettings: Error: Managed storage manifest not found global.js:168:13
[...]

Issue #1817, opened 2023-01-04
After unlocking database from the extension, need to reload page for extension to be able to auto-fill


lodavid:
## Expected Behavior

After unlocking the database in the extension, saved credentials should be available for use on the currently displayed web page in the browser

## Current Behavior

Saved credentials are not visible on the currently displayed web page after unlocking until the page is reloaded

## Steps to Reproduce (for bugs)

1. Lock database in KeepassXC, or lock using the lock database button in the extension
2. Navigate to a web page with saved credentials
3. Unlock the database using "reopen database" button in the extension
4. Attempt to use the extension to auto-fill the login fields, but there are no saved credentials visible, and clicking on the key icon next to the username field doesn't do anything
5. Refresh the page
6. Now the saved credentials are visible in the extension and clicking on the key icon fills in the saved credential

## Debug info

KeePassXC - 2.7.4
KeePassXC-Browser - 1.8.4
Operating system: Win 10
Browser: Firefox and Chromium (Brave)

varjolintu:
Cannot reproduce this. Is there a certain site where this happens, or with every page?

lodavid:
This happens on every page for me.

Some additional info that may be relevant:

- Have multiple databases open in KeepassXC (1 main, 2 secondary)
- 2 of the databases are auto-opened when the main DB is unlocked using AutoOpen
- The most reliable repro is to lock all of the databases in KeepassXC (either manually in the app) or by locking the screen
- When doing the above, upon the first unlock of the main database, I see the key icon on the username flash from grey (with a lock) to green back to grey (with a lock icon)

varjolintu:
So probably the issue is that when you lock databases only one is opened when you click the lock icon, and your credentials are in the second one that remains locked?

lodavid:
All of my databases are opened after unlocking the primary DB, thanks to AutoOpen. Also, this happens for sites in the main DB (the one that is unlocked by the extension).

varjolintu:
Any errors in the browser's JavaScript console?

lodavid:
No errors in the JS console, or any output in the console, for that matter. Yes, the option to search from multiple databases is enabled in KeepassXC. I've also connected the extension to all of the databases.

varjolintu:
No errors? It should be expected that at least "No credentials found" or a similar error would be displayed in the console. Please check the background scripts too via `about:debugging` when inspecting the extension.

lodavid:
[...]

Issue #1616, opened 2022-04-27
Locked symbol still displayed next to username fields in browser after unlocking database


jlkjl32d:
##Current Behaviour
When opening a page that has credential fields and the database is locked, the "keepass database locked" symbol is showed next to the username field. After unlocking the database, the locked symbol is still displayed . Only after a page reload the unlocked symbol is displayed.

Screenshot of locked symbol:
![image](https://user-images.githubusercontent.com/101341798/165460807-b1126f1a-e208-4a82-864e-2dcf28f84e54.png)

## Expected Behavior
After unlocking the database, the correct symbol is showed next to the username on the current page. No page reload is necessary.

## Steps to Reproduce (for bugs)
1. Lock database
2. Open any page with credential fields
3. See the locked symbol next to the username field
4. Click on the field to unlock the database
5. See that still the locked symbol is displayed next to the username field, but credential filling is available

## Debug info

KeePassXC - 2.7.1
KeePassXC-Browser - 1.7.12
Operating system: Win
Browser: Chrome

varjolintu:
Sometimes the "broadcast" messages like `database-unlocked` are not retrieved properly (don't know why). This is why the status doesn't change on icons.

OptimusPrimeme:
I encountered a similar situation, and even if I reload the page, it still show the locked icon and the database is still locked.

varjolintu:
This might have solved it https://github.com/keepassxreboot/keepassxc-browser/pull/1621. Test the current `develop` branch if you can, thanks.

varjolintu:
Still have the issue? It should be fixed.

varjolintu:
@aumouvantsillage Anything in the JavaScript consoles?

aumouvantsillage:
@varjolintu Nothing in the JavaScript console.

Actually, I have misread the original bug report. In the steps to reproduce, it says "the locked symbol is displayed [...] but credential filling is available".

In my case:
* If I unlock the database by clicking on the button "Reopen the database" of the browser extension, everything works correctly.
* If I unlock the database from the KeePassXC window, the lock symbol disappears from the extension button in the toolbar, but there is still a lock symbol in the login form. Autofilling the credentials does not work.

varjolintu:
[...]

PR #3002, opened 2026-06-09
Re-query database state on icon click after reconnect

dbankmann:
when keepassxc sends a database-unlocked notification, the content script's `iconClicked` function checked the stale cached `kpxc.databaseState` and icon CSS class to decide whether to fill credentials. if the database was locked at page load but unlocked later, both checks failed silently — clicking the icon did nothing.

now `iconClicked` calls `updateDatabaseState()` to re-query the background script for the actual database status. if the database is already unlocked, it immediately inits credential fields and fills. if still locked, it sets `manualFill` and triggers the unlock dialog as before.

the test commit extends the sinon-based mocking infrastructure from #3005. without this fix, 3 of the 4 tests fail — the original code never queries the background for current state, so it stays stuck at DISCONNECTED.

related: #1817 #1616 #2594

## Screenshots or videos
to reproduce manually:
1. open a page with a login form
2. KeePassXC database is locked or KeePassXC is not running
3. open/unlock the database in KeePassXC
4. click the KeePassXC icon in the username field
5. **before fix**: nothing happens. **after fix**: credentials fill immediately.

## Testing strategy
behavioral tests using the sinon-based mocking infrastructure from #3005.

## Type of change


keepassxc-browser/content/username-field.js
```diff
@@ -123,22 +123,27 @@ const iconClicked = async function(field, icon) {
...
-        await sendMessage('get_database_hash', [ false, true ]); // Set triggerUnlock to true
-        field.focus();
-    }
+    await kpxc.updateDatabaseState();
+
+    if (kpxc.databaseState === DatabaseState.UNLOCKED) {
+        kpxcIcons.switchIcons();
+
+        if (kpxc.credentials.length === 0) {
+            await kpxc.initCredentialFields();
```
varjolintu, line 137: Why is this required?

dbankmann: if database got unlocked but `detectDatabaseChange` did not run for some reason for this tab, the credentials still will be empty. This fixes the user facing problem at the very last level. Ultimately depends on which contract each internal function should have, whether this is _necessary_

varjolintu:

PR #3003, opened 2026-06-09
Broadcast database unlock notification to all tabs

dbankmann:
when keepassxc sends a database-unlocked notification, `updateDatabaseHashToContent` calls `getCurrentTab()` which queries `{active: true, currentWindow: true}`. in a multi-window setup this sends the notification to the active tab in the wrong window, and the tab with the login form never receives `detectDatabaseChange`.

additionally, switching back to the correct window does not trigger recovery because `windows.onFocusChanged` has no listener — only `tabs.onActivated` is handled, which doesn't fire on window switches.

to reproduce:
1. lock the database in KeePassXC
2. open a login page in browser window A
3. open a second browser window B with any page, keep it focused
4. unlock the database in KeePassXC
5. switch back to window A (without switching tabs)
6. icon still shows locked/disconnected, clicking does nothing

now broadcasts to all tabs via `browser.tabs.query({})`. database lock/unlock is a rare user-initiated event so this has negligible cost compared to the old 1-second polling loop that was used before 2020.

related: #1817 #1616 #2594

## Screenshots or videos
see reproduction steps above.

## Testing strategy
behavioral tests using the sinon-based mocking infrastructure from #3005. the test commit is placed before the fix commit. without the fix, 3 of 5 tests fail:
- "should query all tabs with {} instead of {active: true, currentWindow: true}"
- "should attempt to send to all 3 tabs even if tab 2 fails (per-tab .catch)"
- "should only send to tabs with a valid id"

## Type of change


varjolintu:
The whole idea of the function is just to update the state of the current active tab, not all tabs.

dbankmann:
> The whole idea of the function is just to update the state of the current active tab, not all tabs.

hm, okay, I get that. How about adding a database got unlocked detection to  the `triggerActivatedTab` instead? It seems like some kind of mechanism like that is truly missing.

dbankmann:
Hm, I was finally able to track the issue to an exact repro, see description above. The real problem is multiple brwoser windows. And `triggerActivatedTab` just does not fire on `window.onFocusChanged`. My suggestions as solution for this issue would thus be firing on changing focus as well.

varjolintu:

PR #3004, opened 2026-06-09
Update databaseState regardless of tab visibility

dbankmann:
`detectDatabaseChange` had the entire state update gated behind `document.visibilityState !== 'hidden'`. this was a leftover from the pre-2020 polling era (commit 7270acf) where it prevented background tabs from generating unnecessary traffic every 1-2 seconds.

after the refactor to push-based notifications this guard became harmful: if the notification arrived while the tab was hidden (e.g. keepassxc had focus during unlock), `databaseState` got stuck at LOCKED forever. the tab never recovered without a page reload.

now `databaseState` is always updated to the correct value. only the DOM operations (field init, auto-fill) are still gated behind the visibility check since those are pointless in a hidden tab.

related: #1817 #1616 #2594

## Screenshots or videos
will follow 

## Testing strategy
behavioral tests using the sinon-based mocking infrastructure from #3005. 

## Type of change


varjolintu:
An example would be nice. Could you add a video that shows the issue?

varjolintu:

Issue #2622, opened 2025-08-04
Filling TOTP on a page with different URL


i3v:
### Have you searched for an existing issue?

- [x] Yes, I tried searching and reviewed the pinned issues

### Brief Summary

An intranet web site I'm using requires the following login procedure:
 1. enter Login and Password on a specific web page, say `https://a.com`,
 2. next, enter TOTP on another web page, say `https://b.com`.

That is, the two URLs are very different, even the 2nd level domain names are different. 
This is weird, but I cannot do anything about it.

The thing is, this breaks TOTP autofill.

### Expected Versus Actual Behavior

Option *A* is to create a single KeePass Entry, containing Login, Password and the TOTP seed, with `url=https://a.com`.
Option *B* is to have two separate KeePass Entries (Login and Password in Entry1 with `url=https://a.com`, TOTP in Entry2 with `url=https://b.com`).

Either way, the [`const credentials = kpxc.credentials.find(c => c.uuid === uuid);`](https://github.com/keepassxreboot/keepassxc-browser/blob/bfbb317445d85955fdc951734f4d102177c44b79/keepassxc-browser/content/keepassxc-browser.js#L844) filter returns nothing. 

Even after I refresh the `https://b.com` webpage with `F5`, the `uuid` still refers to the KeePass Entry used on `https://a.com`.
Thus, I only get `Warning! No TOTP found` warning.

The only workaround I found is to duplicate the `https://b.com` tab --- on the duplicated tab KeePassXC-Browser forgets about `https://a.com`  and properly fills the TOTP from the `https://b.com` KeePass Entry. This way, option *B* works.

I wonder is it's possible to either:
 1. remove this filter 
 2. replace this filter with just "sorting" (matching-first)
 3. make it optional (either globally or entry-specific). 

### Steps to Reproduce

See "Summary" section.

### KeePassXC-Browser Debug Information

From the "Copy debug info to clipboard" in the KeePassXC-Browser settings:
```Text
KeePassXC - 2.6.6
KeePassXC-Browser - 1.9.9.1
Operating system: Win64
Browser: Chrome/Chromium 127.0.0.0
```
Actual backend:
```Text
KeePassOTP 1.10.2
KeePassNatMsg 2.0.17
KeePass 2.59 x64
```

i3v:
The [This](https://github.com/keepassxreboot/keepassxc-browser/issues/569#issuecomment-529215016) issue seem to be somewhat similar in that it refers to the same code.

[...]

Issue #1850, opened 2023-02-15
Let users select account from selection list under TOTP input field instead of browser extension button after 2FA expiration


ebbez:
Currently, when a user has 2 accounts linked to a URL and has to re-authenticate their 2FA again after it expired, the user has to select their account from the browser button. This only happens when the user does not have to fill in their username/password, because usually the account gets filtered from the username/password form and then the TOTP is selected from the account used in the username/password form. 

This takes more steps than it should (imo) and the extension also does not make it very clear why the TOTP button is not working in case no account gets selected in the extension button window (except for a small icon change), therefore I suggest moving the account selection for TOTP from the browser extension button to the input field where the TOTP has to be entered into.

## Steps to Reproduce (for bugs)
1. Have 2 accounts linked to microsoftonline.com / a website, one with a 2FA that expires after 30 mins.
2. Log in using username/password, then TOTP will be automatically selected from the account with that combination.
3. Wait till 2FA expires.
4. Press TOTP button (nothing appears)
5. Click on the KeePassXC browser extension button
6. Select account

## Possible solution
As said, it'd be more convenient and user-friendly if a selection list appeared on the TOTP button like a selection list appears just like it does when a user tries to fill in username/password and has 2 accounts.

## Debug info
Browser: Firefox 109.0.1
Extension version: 1.8.4

droidmonkey:
We do show a selection list and an icon on the totp, the website you are accessing is likely messing that up for you. So we'll need to know the site to diagnose what's wrong. 

Also, if you didn't go through a username password selection process, and you just are doing totp, how do we know which entry to choose?

varjolintu:
I guess the whole request is that there shouldn't be need to select an entry, but show TOTP for all entries in the Autocomplete Menu and Popup instead.

Slater91:
> Also, if you didn't go through a username password selection process, and you just are doing totp, how do we know which entry to choose?

Easy: you allow the user to select which entry they want like you do with passwords. Also, sometimes there is only one entry for the website that has TOTP enabled, but the extension still complains that there isn't a TOTP set up for the website.

ebbez:
[...]

PR #3008, opened 2026-06-10
Offer the TOTP entry picker when no login was pre-selected for the page

martenreeh:
## Summary

`kpxc.updateTOTPList()` returns early with **"No TOTP found."** whenever
`page_get_login_id` is `undefined` — i.e. whenever no entry was filled through
KeePassXC on the current tab. This is common with SSO / IdP flows that navigate
straight to a standalone 2FA code page: there is no preceding KeePassXC login on
that tab, so `getLoginId()` returns `undefined`, and the user never reaches the
entry picker that `fillFromTOTP()` already supports for multiple matches.

It is especially visible when several entries share one host (e.g. multiple
Microsoft 365 tenants on `login.microsoftonline.com`): clicking the in-field TOTP
icon just reports "No TOTP found." with no way to choose.

## Change

`updateTOTPList()` is restructured so that:

- it only bails when there are genuinely no credentials for the page;
- when a login **was** pre-selected, behaviour is unchanged (narrow to entries
  sharing that username; a single match still auto-fills);
- when **no** login was pre-selected, it returns every page-matching entry that
  has a TOTP, so `fillFromTOTP()` shows the existing picker for more than one
  match instead of failing.

Because it is user-initiated (clicking the in-field TOTP icon), it never silently
fills the wrong account — it only offers a choice.

## Related

Addresses the situation described in #1850 and #2622.

## Testing

- One matching TOTP entry, no prior login → fills directly (unchanged).
- Multiple matching TOTP entries, no prior login → picker now appears
  (previously "No TOTP found.").
- A prior KeePassXC login on the tab → unchanged (narrowed to the logged-in
  account).

### DIFF
--- a/keepassxc-browser/content/keepassxc-browser.js
+++ b/keepassxc-browser/content/keepassxc-browser.js
@@ -789,33 +789,39 @@ kpxc.updateDatabaseState = async function() {
 
 // Updates the TOTP Autocomplete Menu
 kpxc.updateTOTPList = async function() {
-    let uuid = await sendMessage('page_get_login_id');
-    if (uuid === undefined || kpxc.credentials.length === 0) {
-        // Credential haven't been selected
-        logDebug('Error: No credentials selected for TOTP.');
+    // Nothing was retrieved for this page at all -> genuinely nothing to offer.
+    if (kpxc.credentials.length === 0) {
+        logDebug('Error: No credentials available for TOTP.');
         return;
     }
 
-    // Use the first credential available if not set
-    if (uuid === '') {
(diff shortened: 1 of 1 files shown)


varjolintu:
I'm just about to update the pull request template. Could you also fill the missing section, thanks: https://github.com/keepassxreboot/keepassxc-browser/pull/3001/changes

varjolintu (changes requested):
Still waiting for the pull request template update.

varjolintu:
No information provided. Closing.

PR #3054, opened 2026-07-22
[passkeys] Fixes unhandled callback, double response errors & wrong return type

### LINKED ISSUE #2501: the "Error! Credential is excluded." error is swallowed and not propagated back to the browser/client code
trying to register a passkey twice for the same user/domain shows the red keepassxc-browser alert
```
Error! Credential is excluded.
```

however, it seems the error/exception is swallowed and not propagated back to the browser/client code?

i think if this error occurs, the client should be able to handle the error (e.g. show a message, remove a loading indicator, re-enable a disabled "register" button, etc.)?

using firefox 136.0.4 with keepassxc-browser 1.9.7.

from a quick look into the code this might be related:
https://github.com/keepassxreboot/keepassxc-browser/blob/1.9.7/keepassxc-browser/content/passkeys.js#L109-L112

thanks!

### LINKED ISSUE #3045: Promise not rejected, NotAllowedError cannot be caught

### Brief Summary

If passkeys fallback is disabled, the browser extension doesn't reject the Promise generated by navigator.credentials.get() and navigator.credentials.create() in some cases. It shows a toast notification of the error, but leaves the Promise in a pending state, and doesn't throw a NotAllowedError that can actually be caught.

Tested on Firefox and Microsoft Edge.

Expected: The Promise should be rejected, and the corresponding NotAllowedError should be catch-able.

Actual: The Promise is left pending. A toast notification saying "Error!" followed by some message will appear in the browser window, but no JavaScript Error is thrown. After a while, "passkeys-inject.js" will throw "Uncaught NotAllowedError: lifetimeTimer has expired", but that NotAllowedError cannot actually be caught. It appears to be confined to the browser extension's scope.

### Steps to Reproduce

1. Enable passkeys, but disable passkeys fallback. 
2. Register a passkey for any website.
3. Hide the passkey from the browser extension / Delete the passkey.
4. Try to login using the passkey.

```Text
KeePassXC - 2.7.12
KeePassXC-Browser - 1.10.3
Operating system: Win64
Browser: Microsoft Edge 150.0.0.0
```

a2kolbasov:

Fixed errors in processing error messages when passkeys are missing or a request is rejected.

1. The `setTimeout` callback was returning an error that was silently discarded. As a result, the page context never received the error, and the `Promise` remained pending forever.

2. When `kpxcPasskeysUtils.passkeysFallback === true`, `kpxcPasskeysUtils.sendPasskeysResponse` was being called (and sending a response) twice because of a missing `return ;`.

3. According to the WebAuthn specification, cancelled or rejected operations always return errors, not `null`. Some websites don’t expect this and throw a script error on the page, which also leads to an endless wait for UI updates. (<https://www.w3.org/TR/webauthn-2/#sctn-getAssertion>)

4. Fix lost callback on DB connection failure

------

- Fixes #3045
- Fixes #2501

## Screenshots or videos

## Testing strategy

On <https://webauthn.io/>

1. Try to authenticate without passkeys. Wait for the timeout (1 min).
2. Register.
3. Try to authenticate, but reject the passkey request. Wait for the timeout.

- [`Promise.withResolvers` browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers#browser_compatibility), [Firefox `strict_min_version`](https://github.com/keepassxreboot/keepassxc-browser/blob/d1ab3b4eb56bfb0a86779ca4775b3c364ffd0156/dist/manifest_firefox.json#L189)

## Type of change

### DIFF
--- a/keepassxc-browser/content/passkeys-inject.js
+++ b/keepassxc-browser/content/passkeys-inject.js
@@ -58,9 +58,13 @@ const enablePasskeys = async function() {
     const sendResponse = async function(command, publicKey) {
         const lifetimeTimer = startTimer(publicKey?.timeout);
 
-        const ret = await chrome.runtime.sendMessage({ action: command, args: [ publicKey, window.location.origin ] });
-        if (ret) {
-            passkeysLogDebug('Passkey response', ret.response);
+        let ret = await chrome.runtime.sendMessage({ action: command, args: [ publicKey, window.location.origin ] });
+        passkeysLogDebug('Passkey response', ret);
+
+        // Any error not related to passkeys (no connection to KPXC, database not opened, unknown error, etc.)
+        if (ret === null) {
+            ret = { response: { errorCode: PASSKEYS_REQUEST_CANCELED } };
+        }
 
             let errorMessage;
             if (ret.response?.errorCode) {
@@ -78,7 +82,6 @@ const enablePasskeys = async function() {
             kpxcPasskeysUtils.sendPasskeysResponse(ret.response, ret.response?.errorCode, errorMessage);
             lifetimeTimer.promise.catch(() => {}); // prevent error in console
             lifetimeTimer.abort();
-        }
     };
 
     const isSameOriginWithAncestors = function () {
--- a/keepassxc-browser/content/passkeys-inject.js
+++ b/keepassxc-browser/content/passkeys-inject.js
@@ -58,7 +58,7 @@ const enablePasskeys = async function() {
     const sendResponse = async function(command, publicKey) {
         const lifetimeTimer = startTimer(publicKey?.timeout);
 
-        let ret = await chrome.runtime.sendMessage({ action: command, args: [ publicKey, window.location.origin ] });
+        let ret = await chrome.runtime.sendMessage({ action: command, args: [publicKey, window.location.origin] });
         passkeysLogDebug('Passkey response', ret);
 
         // Any error not related to passkeys (no connection to KPXC, database not opened, unknown error, etc.)
@@ -66,22 +66,22 @@ const enablePasskeys = async function() {
             ret = { response: { errorCode: PASSKEYS_REQUEST_CANCELED } };
         }
 
-            let errorMessage;
-            if (ret.response?.errorCode) {
-                errorMessage = await chrome.runtime.sendMessage({
-                    action: 'get_error_message',
-                    args: ret.response.errorCode,
-                });
-                kpxcUI.createNotification('error', errorMessage);
+        let errorMessage;
+        if (ret.response?.errorCode) {
+            errorMessage = await chrome.runtime.sendMessage({
+                action: 'get_error_message',
+                args: ret.response.errorCode,
+            });
+            kpxcUI.createNotification('error', errorMessage);
 
-                if (!kpxcPasskeysUtils.passkeysFallback && letTimerRunOut(ret.response.errorCode)) {
-                    await lifetimeTimer.promise;
-                }
+            if (!kpxcPasskeysUtils.passkeysFallback && letTimerRunOut(ret.response.errorCode)) {
+                await lifetimeTimer.promise;
             }
+        }
 
-            kpxcPasskeysUtils.sendPasskeysResponse(ret.response, ret.response?.errorCode, errorMessage);
-            lifetimeTimer.promise.catch(() => {}); // prevent error in console
-            lifetimeTimer.abort();
+        kpxcPasskeysUtils.sendPasskeysResponse(ret.response, ret.response?.errorCode, errorMessage);
+        lifetimeTimer.promise.catch(() => { }); // prevent error in console
+        lifetimeTimer.abort();
     };
 
     const isSameOriginWithAncestors = function () {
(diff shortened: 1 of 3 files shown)


varjolintu (changes requested):
Quickly tested this and noticed that when canceling the authentication request, the extension now immediately returns `NotAllowedError`. This goes against: https://www.w3.org/TR/webauthn-2/#sctn-assertion-privacy

Also explained at https://www.w3.org/TR/webauthn-2/#sctn-getAssertion step 18:
> Return a DOMException whose name is "NotAllowedError". In order to prevent information leak that could identify the user without consent, this step MUST NOT be executed before lifetimeTimer has expired.

a2kolbasov:
> Quickly tested this and noticed that when canceling the authentication request, the extension now immediately returns `NotAllowedError`.

At first I thought it was very strange, because there is `setTimeout(resolve, timeout)`. But now I realized that the `PASSKEYS_REQUEST_CANCELED` condition is missing in
https://github.com/keepassxreboot/keepassxc-browser/blob/379dfaf08e3f63dadf17a126468818c59314c3ca/keepassxc-browser/content/passkeys-inject.js#L36-L42

> This goes against: https://www.w3.org/TR/webauthn-2/#sctn-assertion-privacy
> 
> Also explained at https://www.w3.org/TR/webauthn-2/#sctn-getAssertion step 18

On the other hand, in WebAuthn-3 privacy is implemented differently

- https://github.com/w3c/webauthn/pull/2134
- <https://www.w3.org/TR/webauthn-3/#sctn-getAssertion:~:text=Throw%20a%20%22NotAllowedError%22%20DOMException%2E,-5%2E1%2E4%2E2>
- https://www.w3.org/TR/webauthn-3/#sctn-assertion-privacy

varjolintu:
> On the other hand, in WebAuthn-3 privacy is implemented differently
> 
>     * [Update obsolete privacy concerns about throwing errors early w3c/webauthn#2134](https://github.com/w3c/webauthn/pull/2134)
> 
>     * https://www.w3.org/TR/webauthn-3/#sctn-getAssertion:~:text=Throw%20a%20%22NotAllowedError%22%20DOMException%2E,-5%2E1%2E4%2E2
> 
>     * https://www.w3.org/TR/webauthn-3/#sctn-assertion-privacy

The latest released version is still 2. It's not recommended to implement changes from 3, because those might still change: https://www.w3.org/standards/types/#CR. I agree it will simplify a lot of things just to return an error immediately instead of waiting for the lifetimer.

a2kolbasov:
Okay, got it. I've made changes in a044542.

a2kolbasov:
When I was recording the video demonstration, I found another error. If the database is locked, the content script does not handle this situation, and the page script will never receive a response with an error.

keepassxc-browser/content/passkeys-inject.js
```diff
@@ -21,50 +22,66 @@ const enablePasskeys = async function() {
...
-
-    const stopTimer = function(lifetimeTimer) {
-        if (lifetimeTimer) {
-            clearTimeout(lifetimeTimer);
-        }
+    /**
+     * @param {number=} timeout
+     */
+    const startTimer = function (timeout) {
+        let resolve, reject;
```
varjolintu, line 29: Why not use `const` instead of `let` inside the function?

keepassxc-browser/content/passkeys-inject.js
```diff
@@ -21,50 +22,66 @@ const enablePasskeys = async function() {
...
-                }
-            }
+        let ret = await chrome.runtime.sendMessage({ action: command, args: [publicKey, window.location.origin] });
+        passkeysLogDebug('Passkey response', ret);
 
-            passkeysLogDebug('Passkey response', ret.response);
-            kpxcPasskeysUtils.sendPasskeysResponse(ret.response, ret.response?.errorCode, errorMessage);
-            stopTimer(lifetimeTimer);
+        // Any error not related to passkeys (no connection to KPXC, database not opened, unknown error, etc.)
+        if (ret === null) {
```
varjolintu: Here's a chance for a small refactor:
`const errorCode = ret === null ? PASSKEYS_REQUEST_CANCELED : ret.response.errorCode;`

And then just use the `errorCode` for checks and send message etc. This way we can keep the `ret` as `const` and the code stays a bit cleaner.

varjolintu: Otherwise this seems to work great now!

keepassxc-browser/content/passkeys-inject.js
a2kolbasov: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers#description

`reject` uses [inside `abort`](https://github.com/keepassxreboot/keepassxc-browser/pull/3054/changes#diff-28fd434be65546ca50a96e1cc46fc4a19f545f27d3dbd849063ec654b034ad15R42)

a2kolbasov: ~~`let promise` & `let timerId` can be replaced with `const ...`~~

varjolintu (approved):
Now this looks good. Thanks!

varjolintu:
A few null checks were missing that caused the script to throw errors: https://github.com/keepassxreboot/keepassxc-browser/pull/3083

PR #3056, opened 2026-07-23
Initialize Custom Login Fields when automatic detection finds no fields

thanksmate1985:
## Summary

Custom Login Fields intentionally provide an override when automatic field detection cannot identify a login form. Previously, if automatic detection returned no fields, initialization returned before the custom-field combination could initialise icons or request credentials. This keeps valid custom combinations active and reinitialises them immediately after a selection is saved.

## Validation

- `npm run lint`

## Attribution

Implemented and validated by Codex, an AI coding agent, under the contributor's direction.

### DIFF
--- a/keepassxc-browser/content/custom-fields-banner.js
+++ b/keepassxc-browser/content/custom-fields-banner.js
@@ -334,6 +334,7 @@ kpxcCustomLoginFieldsBanner.confirm = async function(e) {
     if (!kpxc.settings[DEFINED_CUSTOM_FIELDS]) {
         kpxc.settings[DEFINED_CUSTOM_FIELDS] = {};
     }
+    let fieldsUpdated = false;
 
     // If the new selection is already used in some other field, clear it
     const clearIdenticalField = function(path, location) {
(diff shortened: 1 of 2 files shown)


varjolintu:
We don't allow AI contributions in this repository, thank you.
https://github.com/keepassxreboot/keepassxc-browser/blob/develop/.github/CONTRIBUTING.md#using-ai

thanksmate1985:
This is a real bug and should be fixed, if a user goes through the effort of adding a "redetect fields" that should not be ignored due to a filter designed to not include "search" fields.

FYI I'm a developer myself, and IMHO AI Coding agents have evolved to a point where for a simple project like this it can be fairly well trusted - that contribution note should be revisited in light of this. That said - If you don't want to accept this PR then feel free to code it manually yourself, as a developer on this project I don't doubt you could do it in a better way. Should an update rebreak my fix I'll simply move to another extension/solution for my password needs.

varjolintu:
You can create an issue about the bug.

> FYI I'm a developer myself, and IMHO AI Coding agents have evolved to a point where for a simple project like this it can be fairly well trusted - that contribution note should be revisited in light of this. 

The contributing rules of a repository has should be respected, whatever opinion you have personally.

thanksmate1985:
I appreciate your opinion and stance, honestly I do, I appreciate you taking the time to read and all the time you've spent on this project for free. However, you should respect your own time more, you're working on a project and if you want the best for it, don't just reject code bug fixes based on whatever this was based on - yes "rules are rules" but you're disrespecting your own efforts to make this project the best it can be by taking this line. Just my 2c, goodluck with it.

